PR #39, opened 2016-12-19
Fix inline_message_id not using inlineMessageId

runegan:
There was an error when I used editMessageText on a context object from a callback_query update and the update came from an inline message.


dotcypress:
Hey, you can use `edit*` method for both cases, see example:

```js
// Messages id
editMessageReplyMarkup (chatId, messageId, null, markup)

// Messages id
editMessageReplyMarkup (chatId, null, inlineMessaegId, markup)

```

dotcypress:
👍  Thank you!

dotcypress:
Landed in `3.6.2`

PR #102, opened 2017-05-01
Fix ignoring webhookReply option

TheBeastOfCaerbannog:
When webhookReply is set by app.telegram.webhookReply and not from options passed to contructor this option is ignored. Actually, all telegram options was reverted to defaults, but only this one has get/set methods.
Steps to reproduce:
```
const Telegraf = require('telegraf')

const app = new Telegraf()

const baseMessage = {
  chat: {
    id: 1
  }
}

app.telegram.webhookReply = false

app.on('message', (ctx) => {
  ctx.reply(':)')
    .then((res) => console.log(res))
})
const res = {
  setHeader: () => undefined,
  end: () => undefined
}
app.handleUpdate({message: baseMessage}, res)

```
ctx.reply should throw because no token were provided. But instead it replies to webhook.

### DIFF
--- a/lib/telegraf.js
+++ b/lib/telegraf.js
@@ -12,6 +12,7 @@ class Telegraf extends Composer {
       retryAfter: 1,
       handlerTimeout: 0
     }, options)
+    this.telegram = new Telegram(token, this.options.telegram)
     this.token = token
     this.handleError = (err) => {
       console.error()
@@ -32,7 +33,7 @@ class Telegraf extends Composer {
 
   set token (token) {
     this.options.token = token
-    this.telegram = new Telegram(this.options.token, this.options.telegram)
+    this.telegram = new Telegram(this.options.token, this.telegram.options)
   }/* eslint brace-style: 0 */
 
   catch (handler) {
@@ -96,7 +97,7 @@ class Telegraf extends Composer {
   handleUpdate (update, webhookResponse) {
     debug('⚡ update', update.update_id)
     const telegram = webhookResponse
-      ? new Telegram(this.token, this.options.telegram, webhookResponse)
+      ? new Telegram(this.token, this.telegram.options, webhookResponse)
       : this.telegram
     const ctx = new Context(update, telegram, this.options)
     Object.assign(ctx, this.context)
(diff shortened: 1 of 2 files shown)


dotcypress:
Hey.
Thank you for PR.

Now, all setting stores in `this.options` and duplicate this information in `this.telegram` isn't good idea.

Try to provide `webhookReply` as in following snippet:

```js
const Telegraf = require('telegraf')

const app = new Telegraf(null, {
  telegram: { webhookReply: false } 
})

const baseMessage = {
  chat: {
    id: 1
  }
}

app.on('message', (ctx) => {
  ctx.reply(':)')
    .then((res) => console.log(res))
})
const res = {
  setHeader: () => undefined,
  end: () => undefined
}
app.handleUpdate({message: baseMessage}, res)
```

TheBeastOfCaerbannog:
@dotcypress Thank you for quick response

Well, yes, it works when webhookReply passed into Telegraf constructor. But get/set methods for app.telegram.webhookReply [does exist](https://github.com/telegraf/telegraf/blob/develop/lib/network/client.js#L57). And doesn't do anything at the moment because at [this line](https://github.com/telegraf/telegraf/pull/102/files#diff-b1b494e6f6a8bb0e8a9899cc6c337b38L99) they are overwritten by options from constructor. 🤔

dotcypress:
Yes, now you can only use constructor, maybe makes sense to create getter/setter, but honestly i can't see any use cases when we need dynamically disable/enable webhook reply feature.

TheBeastOfCaerbannog:
I don't see any use cases either :)

And getter/setter already exists. 
Here's why I propose this change:
webhookReply was the only thing I needed, so I decided not to pass options in constructor, but set it with `
app.telegram.webhookReply = false`. And it didn't work. So getter/setter should be fixed or removed so it not would be misleading, right?

dotcypress:
Getter/setter exists only in `Telegram` class not in `Telegraf`, and i'm not sure that we need to duplicate functionality. Anyway, I have couple ideas how to fix this case, stay tuned.

TheBeastOfCaerbannog:
Ok, thank you :)

Should this PR be closed?

dotcypress:
No problem, I will close it later.

dotcypress:
@TheBeastOfCaerbannog  Fix landed in `v3.6.8`.

Now you can use `app.webhookReply = false` to disable `Reply via webhook` feature.

Thank you for you time 👍

TheBeastOfCaerbannog:
@dotcypress Thank you :)

PR #145, opened 2017-07-14
bug fix

Lisand7ru:
There are no payload.response, payload.migrate_to_chat_id and payload.retry_after
I fixed this bug in TelegramError exception and added a more abstract prop "parameters"

### DIFF
--- a/lib/telegraf.js
+++ b/lib/telegraf.js
@@ -119,7 +119,7 @@ class Telegraf extends Composer {
     }
     this.telegram.getUpdates(timeout, limit, offset, allowedUpdates)
       .catch((err) => {
-        const wait = this.parameters.retry_after
+        const wait = err.parameters.retry_after || this.options.retryAfter
         console.error(`Failed to get updates. Waiting: ${wait}s`, err)
         return new Promise((resolve) => setTimeout(resolve, wait * 1000, []))
       })


lib/network/error.js
```diff
@@ -2,11 +2,10 @@ class TelegramError extends Error {
...
+    this.response = payload
     this.description = payload.description
-    this.retryAfter = payload.retry_after
-    this.migrateToChatId = payload.migrate_to_chat_id
+    this.parameters = payload.parameters || {}
   }
 }
 
-module.exports = TelegramError
+module.exports = TelegramError
```
dotcypress: Please add new line here

lib/network/error.js
```diff
@@ -2,11 +2,10 @@ class TelegramError extends Error {
   constructor (payload = {}) {
     super(`${payload.error_code}: ${payload.description}`)
     this.code = payload.error_code
-    this.response = payload.response
+    this.response = payload
     this.description = payload.description
-    this.retryAfter = payload.retry_after
```
dotcypress: Could you also fix `retryAfter` usage here: https://github.com/telegraf/telegraf/blob/develop/lib/telegraf.js#L122 ?

Lisand7ru: done

lib/telegraf.js
```diff
@@ -119,7 +119,7 @@ class Telegraf extends Composer {
     }
     this.telegram.getUpdates(timeout, limit, offset, allowedUpdates)
       .catch((err) => {
-        const wait = err.retryAfter || this.options.retryAfter
+        const wait = this.parameters.retry_after
```
dotcypress: Please add fallback to `this.options.retryAfter`

PR #147, opened 2017-07-16
bug fix

Lisand7ru:
Sorry for new pull request

### DIFF
--- a/lib/telegraf.js
+++ b/lib/telegraf.js
@@ -119,7 +119,7 @@ class Telegraf extends Composer {
     }
     this.telegram.getUpdates(timeout, limit, offset, allowedUpdates)
       .catch((err) => {
-        const wait = err.retryAfter || this.options.retryAfter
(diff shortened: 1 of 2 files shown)


lib/telegraf.js
```diff
@@ -119,7 +119,7 @@ class Telegraf extends Composer {
     }
     this.telegram.getUpdates(timeout, limit, offset, allowedUpdates)
       .catch((err) => {
-        const wait = err.retryAfter || this.options.retryAfter
+        const wait = err.parameters.retry_after || this.options.retryAfter
```
dotcypress: I think we need to add an extra check:
`(err.parameters && err.parameters.retry_after) || this.options.retryAfter`

Lisand7ru: TelegramError.parameters always contains an object
`this.parameters = payload.parameters || {}`

dotcypress: Yes, but we can't claim that all Errors will be instance of `TelegramError`

dotcypress:
Thank you 👍

PR #148, opened 2017-07-17
fix for Composer.command

Lisand7ru:
Hello. It's me again.

I just have found out that Composer.command accept any command that is containing in a message. That's not a right way.

Look at official bots - @botfather, @telegraph. They consider a text message as a command only if it starts with a command.


dotcypress:
Hey
Makes sense.

dotcypress:
@Lisand7ru please fix or delete failing test: https://github.com/telegraf/telegraf/blob/develop/test/composer.js#L469

Lisand7ru:
Done

dotcypress:
Published @ `3.8.2`

PR #164, opened 2017-08-21
get message of callback_query from ctx.message

EdJoPaTo:
As a user I would expect `ctx.message` to be filled when `this.callbackQuery.message` contains a message.
It should be the same as for example `ctx.chat` which is filled with `ctx.callbackQuery.message.chat` when available.

### DIFF
--- a/lib/core/context.js
+++ b/lib/core/context.js
@@ -64,7 +64,8 @@ class TelegrafContext {
   }
 
   get message () {
-    return this.update.message
+    return this.update.message ||
+      (this.callbackQuery && this.callbackQuery.message)
   }
 
   get editedMessage () {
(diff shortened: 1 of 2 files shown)


dotcypress:
Hey @EdJoPaTo 
Thank you for your PR, but afraid I can't merge your PR.

I prefer to avoid inconsistency in library interface, let's me explain.

Telegraf context contains only a few synthetic properties:
* `state` - for sharing state between middlewares
* `chat` - just sugar for easy routing and internal use
* `from` - same as above

And dozen of properties mapped to current Telegram update payload:
* `message`
* `editedMessage`
* `inlineQuery`
* `shippingQuery`
* `preCheckoutQuery`
* `chosenInlineResult`
* `channelPost`
* `editedChannelPost`
* `callbackQuery`

So, adding an implicit reference between top-level update type(e.g. `message`) and synthetic property `message`, will produce inconsistency.

But if you want this sort of magic, you can easily copy-paste following code:
```js
bot.on('callback_query', (ctx, next) => {
  ctx.update.message = ctx.callbackQuery.message
  return next()
})
```

PR #190, opened 2017-10-13
Fix `from` getter of context class

olimsaidov:


dotcypress:
Thanks for your help 👍

PR #217, opened 2017-11-08
bugfix(stage): It was not possible to add a listener after registering the middleware

XFree:
It was not possible to add a listener after registering the middleware

### DIFF
--- a/stage.js
+++ b/stage.js
@@ -28,7 +28,7 @@ class Stage extends Composer {
         ctx.scene = new SceneContext(ctx, this.scenes, this.options)
         return next()
       },
-      super.middleware(),
+      lazy((ctx) => super.middleware()),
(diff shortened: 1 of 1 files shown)


dotcypress:
I'm afraid but I can't merge this PR, cause it will degrade performance for all `Stage` users.

But you can easily implement lazy routing in bot code:

```js
const { lazy } = require('telegraf')

stage.on('message', lazy(getDynamicHandler))
```

XFree:
I do not think that performance will get worse, but it will allow you to use, on, command method etc. after executing the method middeware().
In my code, I just inherited and overloaded the method, but not true way.
stage.on('message', lazy(getDynamicHandler)) not work after call stage.middleware().  This complicates the logic.

PR #240, opened 2017-11-28
new "on" parameter to TelegramError class

smaznet:


dotcypress:
Thank you!

core/network/error.js
```diff
@@ -1,10 +1,11 @@
 class TelegramError extends Error {
-  constructor (payload = {}) {
+  constructor (payload = {}, on) {
```
sergeysova: Maybe use default param?

dotcypress: Make sense

PR #248, opened 2017-12-04
Fixed links and a bit more

Piterden:


docs/README.md
```diff
@@ -160,7 +161,7 @@ Context is created per request and contains following props:
 
 ```js
 bot.use((ctx) => {
-  ctx.telegram             // Telegram client instance
+  ctx.tg                   // Telegram client instance
```
dotcypress: We have [public accessor](https://github.com/telegraf/telegraf/blob/develop/core/context.js#L62) for Telegram instance.

Piterden:
Hm... Okay. But why not to do `this.telegram = telegram` there
https://github.com/telegraf/telegraf/blob/develop/core/context.js#L43
???

dotcypress:
Because this is argument with name 'telegram'

Piterden:
Sorry, I am not follow.
So, an argument named `telegram` passed to the constructor, and what?

PR #263, opened 2017-12-11
answerWith to replyWith

Besatnias:
Fixed an oversight in the docs that was screwing with my bot-making disciples.


dotcypress:
Sorry for that 🙃

And thank you for PR

PR #271, opened 2017-12-16
Implemented onReply method

Daspharion:


dotcypress:
Great job!

Unfortunately, I can't merge this PR, cause storing bot state(reply listeners) in heap doesn't scale. For example, we can't run the cluster of bots behind the load balancer, and any bot restart will cause data loss. 

Anyway, you can extract middleware from PR.

Issue #305, opened 2018-01-18
TypeError: Cannot read property 'source' of null


gdouezangrard:
Last version 3.17.4 brought the following error, even with the code sample from the docs (http://telegraf.js.org/#/?id=example). Downgrading to version 3.17.2 fixed the issue.

```
node v9.3.0
telegraf v3.17.4
```

```js
const Telegraf = require('telegraf')

const bot = new Telegraf(process.env.BOT_TOKEN)

bot.start((ctx) => {
  console.log('started:', ctx.from.id)
  return ctx.reply('Welcome!')
})

bot.command('help', (ctx) => ctx.reply('Try send a sticker!'))
bot.hears('hi', (ctx) => ctx.reply('Hey there!'))
bot.hears(/buy/i, (ctx) => ctx.reply('Buy-buy!'))
bot.on('sticker', (ctx) => ctx.reply('👍'))

bot.startPolling()
```

```
/telegram/node_modules/telegraf/core/network/client.js:58
      : typeof payload[key] === 'object' && (payload[key].source || payload[key].url)
                                                          ^

TypeError: Cannot read property 'source' of null
    at Object.keys.some (/telegram/node_modules/telegraf/core/network/client.js:58:59)
    at Array.some (<anonymous>)
    at includesMedia (/telegram/node_modules/telegraf/core/network/client.js:55:31)
    at Telegram.buildFetchConfig (/telegram/node_modules/telegraf/core/network/client.js:135:12)
    at Telegram.callApi (/telegram/node_modules/telegraf/core/network/client.js:112:17)
    at Telegram.getUpdates (/telegram/node_modules/telegraf/telegram.js:28:17)
    at Telegraf.fetchUpdates (/telegram/node_modules/telegraf/telegraf.js:130:19)
    at Telegraf.startPolling (/telegram/node_modules/telegraf/telegraf.js:72:12)
    at Object.<anonymous> (/telegram/main.js:10:5)
    at Module._compile (module.js:660:30)
```

Piterden:
Could you highlight your code, please?

gdouezangrard:
Done. This is basically http://telegraf.js.org/#/?id=example

dotcypress:
Thanks for report.

Fixed in `v3.17.5`

PR #306, opened 2018-01-18
in case when payload[key] == null error exists

logovazzik:


gdouezangrard:
linking #305

core/network/client.js
```diff
@@ -55,7 +55,7 @@ function includesMedia (payload) {
   return Object.keys(payload).some(
     (key) => Array.isArray(payload[key])
       ? payload[key].some(({ media }) => typeof media === 'object' && (media.source || media.url))
-      : typeof payload[key] === 'object' && (payload[key].source || payload[key].url)
+      : typeof payload[key] === 'object' && payload[key] && (payload[key].source || payload[key].url)
```
Piterden: `payload[key] && typeof payload[key] === 'object' && (payload[key].source || payload[key].url)`

logovazzik: changed

Piterden: Even
```js
!!payload[key] && typeof payload[key] === 'object' && (payload[key].source || payload[key].url)
```

Piterden: No. Damm!

PR #339, opened 2018-02-10
Add optional messageId param to TelegrafContext::deleteMessage

GingerPlusPlus:
This change makes `ctx.deleteMessage` consistent with other context shortcuts, like `getChatMember`, `kickChatMember`, `promoteChatMember`, `reply`, which are just "scoped" to the current chat.

It just feels so right, to a point in which I tried to use it in my code and [inserted a bug](https://github.com/thedevs-network/the-guard-bot/issues/55) as a result, and I'm sure I'll do it again, hence the PR.

### DIFF
--- a/core/context.js
+++ b/core/context.js
@@ -461,7 +461,11 @@ class TelegrafContext {
     return this.reply(html, Object.assign({ 'parse_mode': 'HTML' }, extra))
   }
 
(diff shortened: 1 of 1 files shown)


dotcypress:
Thank you!

PR #340, opened 2018-02-11
Set message parameters

dark-dao:
Convenient setting of parameters for the message when using the keyboard

### DIFF
--- a/markup.js
+++ b/markup.js
@@ -15,8 +15,7 @@ class Markup {
   }
 
   extra (options) {
-    const reply_markup = Object.assign({}, this)
-    return Object.assign({reply_markup}, options)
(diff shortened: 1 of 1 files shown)


dotcypress:
@dark-dao could you fix lint errors and describe motivation to make this change?

dark-dao:
I needed to format the message text as Html.
In the current implementation of the library, I could not find such an opportunity
```telegraf.hears(`text`, (ctx) => {
    return ctx.reply('<b>Bold text</b>', Markup
      .keyboard([
        ['Item1'],
        [`Item2`]
      ])
      .oneTime()
      .resize()
      .extra({parse_mode: 'HTML'})
    );
});```

PR #384, opened 2018-03-30
Allow to pass the extra object to the editMessageReplyMarkup

Piterden:


AbdelrahmanMohamedDesoky:
@dotcypress 
Why did you close this, this is actually very good modification and I stumbled upon the editMessageReplyMarkup for Markup that I almost gave up.

Not to mention that there are a lot of people other than me who also got confused by the weird .extra() object.

Since you need .extra() for sending inline_keyboard using sendMessage() but for editMessageReplyMarkup() if you were to use Markup to generate the inline keyboard you can't use .extra() otherwise it removes the whole inline_keyboard from the message and doesn't display or throw any errors so it confuses people.

PR #402, opened 2018-04-25
Feature/typings

DinoLeung:
updates from sergeysova/telegram-typings


fennecbutt:
You need to add `index.d.ts` to the root of the project so that tsc/vscode can automatically find it.

dotcypress:
@DinoLeung Thank you!

PR #415, opened 2018-05-11
Fix TypeScript compilation

TimvdLippe:
I wanted to work on a TypeScript project with Telegraf, but discovered it was not compiling. Apparently, the `ReadableStream` interface is defined in the `NodeJS` namespace, which thus requires the prefixing.

To prevent any typings mistakes, I have added a `typecheck` command that runs `tsc` and enforce it in the precommit and on Travis. This caught an additional error in the `tsconfig.json` where the `files` were not correctly specified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,5 @@ node_js:
   - 8
 script:
   - npm run lint
+  - npm run typecheck
   - npm test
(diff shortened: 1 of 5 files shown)


dotcypress:
@TimvdLippe Thank you a lot 💪

Updated version will be published soon

PR #493, opened 2018-08-30
Adding Markup support to TS typings and some other things.

Fazendaaa:
# Description

Just improving TypeScript typings:
* Adding Markup class support
* Correcting answerCallbackQuery to answerCbQuery and adding catch support

# How Has This Been Tested?

No testing implemented

# Checklist:

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -245,6 +245,7 @@ export interface ContextMessageUpdate extends Context {
   answerInlineQuery(results: tt.InlineQueryResult[], extra?: tt.ExtraAnswerInlineQuery): Promise<boolean>
 
   /**
+   * @deprecated answerCallbackQuery() is deprecated, use answerCbQuery() instead
    * Use this method to send answers to callback queries.
    * @param text Notification text
    * @param url Game url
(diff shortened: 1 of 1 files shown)


Piterden:
This file is generated on a fly, you should edit telegram-typings package...

Fazendaaa:
@Piterden , do you know how can I make typings for Stage support? I'm having a little bit of trouble with that right now.

Piterden:
@Fazendaaa don't understand what you mean. Could you explain more detailed?

Fazendaaa:
@Piterden right now, I've my TS project and trying to do the following [Stage](https://telegraf.js.org/#/?id=stage) example and failing because of the used types are not defined -- which is ok.

My doubt is "how can I make the used types since is a submodule dependency?". I just don't know the procedure, but I'm willing to make it once I learn it.

Do you know how can I make it? Is there anywhere a example?

lenovouser:
@Piterden are you sure this is generated on the fly? `Markup` for example comes from [`telegraf/markup.js`](https://github.com/telegraf/telegraf/blob/develop/markup.js) which is custom to this module. So his change seems right to me?

lenovouser:
@Piterden `telegram-typings` are the typings for the responses that Telegram gives us over the network. These typings here are something different / for the node module - so what @Fazendaaa did is correct and can be merged as soon as he rebases in my opinion.

Piterden:
@lenovouser got it. Sorry, I was confused and hurried up a bit... Anyway, TT needs to be rewritten.

PR #494, opened 2018-08-31
Adds the option to reconnect after a 409 error

WolfyLPDC:
# Description

Adds the option to prevent Telegraf from stop working after a 409 error. This error happens when the getUpdates call to API has a conflict with another request.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

The breaking change exists because of the order of the `startPolling` arguments. I decided to keep the callback at the end to respect the style format but it can be changed easily by the the maintainers if needed.

# How Has This Been Tested?

This is actually a minor change that does not involve a global impact, so no big tests are needed. However, I tested it by creating a basic bot and forcing the 409 error.

Code:
```
const Telegraf = require('./telegraf.js')

const bot = new Telegraf("BOT_TOKEN"})

bot.on("message", ctx => {
  console.log(ctx.message);
})

bot.startPolling(undefined, undefined, undefined, true)
```

URL to force conflict (You have to replace BOT TOKEN by the real token):
https://api.telegram.org/botBOT_TOKEN/getUpdates?offset=0&limit=100&timeout=30

**Test Configuration**:
* Node.js Version: v8.9.1

# Checklist:

### DIFF
--- a/telegraf.js
+++ b/telegraf.js
@@ -61,7 +61,7 @@ class Telegraf extends Composer {
     return generateCallback(path, (update, res) => this.handleUpdate(update, res), debug)
   }
 
-  startPolling (timeout = 30, limit = 100, allowedUpdates, stopCallback = noop) {
+  startPolling (timeout = 30, limit = 100, allowedUpdates, reconnectAfterConflict = false, stopCallback = noop) {
     this.polling.timeout = timeout
     this.polling.limit = limit
     this.polling.allowedUpdates = allowedUpdates
@@ -70,7 +70,7 @@ class Telegraf extends Composer {
     this.polling.stopCallback = stopCallback
     if (!this.polling.started) {
       this.polling.started = true
-      this.fetchUpdates()
+      this.fetchUpdates(reconnectAfterConflict)
     }
     return this
   }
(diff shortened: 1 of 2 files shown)


Piterden:
`bot.startPolling(undefined, undefined, undefined, true)`

Really?!!! Are you crazy?

No one framework user except you uses a second fake request to API. There isn't any reason to do it. I advise you to make a separate fork for this functionality...

WolfyLPDC:
It is a fuctionality that can be perfectly compatible with the rest of the code. I decided to move the callback but you can add a method to set autoReconnect or another startPolling method that does not stop after a conflict.

This functionality is the same that the Telegram frameworks from another languages have. Take a look to telepot (Python). 

I deeply support that having an option to auto reconnect is critical when you are going to run the bot in a server 24/7.

Piterden:
Why not webhooks? @WolfyLPDC

WolfyLPDC:
@Piterden Because I do not know how to work with webhooks properly. However, I want to switch to them in my application in the next months.

Piterden:
Take a look there https://github.com/Piterden/syntax-highlighter-bot/blob/1.0/src/server.mjs

PR #498, opened 2018-09-03
typescript declaration moving to class-to-interface decomposiotion (ref: https://github.com/joshtynjala/typescript-notes/blob/master/Class-Decomposition.md )

tvolodimir:
fixing typos in typescript declaration and moving from class to interface decomposition.
see: https://github.com/joshtynjala/typescript-notes/blob/master/Class-Decomposition.md
That will help to easily merge declaration between class.
see: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -890,7 +890,6 @@ export interface TelegrafConstructor {
    * new Telegraf(token, options)
    */
   new <C extends ContextMessageUpdate>(token: string, options?: TelegrafOptions): Telegraf<C>;
-  new <C extends ContextMessageUpdate>(): Telegraf<C>;
 }
 
 export default Telegraf
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -342,19 +342,13 @@ export interface Middleware<C extends ContextMessageUpdate> {
 
 export type HearsTriggers = string[] | string | RegExp | RegExp[] | Function
 
-export class Telegram {
+export const Telegram: TelegramConstructor;
+
+export interface Telegram {
   /**
    * Use this property to control reply via webhook feature.
    */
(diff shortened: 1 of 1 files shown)


Piterden:
https://github.com/telegraf/telegraf/pull/493#issuecomment-417489379

tvolodimir:
@Piterden Are you sure? What scripts generate these files?

Piterden:
Ya, sorry, I was confused a bit, there is another file. But maybe you would fix the telegram-typings package and to your static  code would became dynamically generated?

tvolodimir:
@Piterden Did not catch your idea. I made changes exactly for telegraf package. it's not relative 'telegram-typings'. You may know that current state of ts declaration of this library isn't in good shape. And sadly to say that telegraf library give more pain for typescript projects. So my suggestion are to decomposite classes to interfaces - so any projects owner will be able to extend declaration with his intrenal ts declarations

Piterden:
@tvolodimir I agreed, but if we would make it generated on a fly, then it will be a hundred times more awesome stuff, isn't it?

As an example run this code in your browser console on the tg bots api page.
https://gist.github.com/Piterden/baa9ceaab30e295dabda9c7d6485a04b

https://en.wikipedia.org/wiki/Metaprogramming

tvolodimir:
@Piterden Cool script, In this case someone written manually these declarations so we can transform it to some another shape of data like ts declaration. But... but with telegraf we have another story - it written in js even without jsdoc or flow type description. So we have one way - support ts declaration manually. Otherwise, telegraf isn't complicated library and can be easealy rewrite to typescript in 1-2 days. Anyway, updating ts declaration now is better than nothing

PR #526, opened 2018-10-06
Added context property to the Telegraf interface

vladimir-fed:
# Description

Added context property to the `Telegraf` interface. Moved typescript to devDependencies.

## Type of change

**Test Configuration**:
* Node.js Version:
* Operating System:

# Checklist:

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -336,8 +336,8 @@ export interface ContextMessageUpdate extends Context {
 
 }
 
-export interface Middleware<C extends ContextMessageUpdate> {
-  (ctx: C, next?: () => any): any
+export interface Middleware<TContext extends ContextMessageUpdate> {
+  (ctx: TContext, next?: () => any): any
 }
 
 export type HearsTriggers = string[] | string | RegExp | RegExp[] | Function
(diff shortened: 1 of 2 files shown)


typings/index.d.ts
```diff
@@ -844,6 +844,11 @@ export interface Telegraf<C extends ContextMessageUpdate> extends Composer<C> {
    */
   telegram: Telegram
 
+  /**
+   * Use this property to extend context and support interface C
+   */
+  context: C
```
Piterden: Why `C`? Why not `Cat` or `Carpet`?

This definitely should ba called `TelegrafContext` just like it called in the sources. Anyway not `C`!

vladimir-fed: Because `Telegraf<C extends ContextMessageUpdate>`

Piterden: There is also something `Composer`

Piterden: image

Piterden: Yeah I see C

Piterden: That's why I don't like TS actually (

vladimir-fed: I can update whole interface `Telegraf` and change `C` to the `TelegrafContext`, if you want.

Piterden: What about `R extends ContextMessageUpdate`???

Piterden: > I can update whole interface Telegraf and change C to the TelegrafContext, if you want.

Ask @dotcypress about it, please.

vladimir-fed: @dotcypress, I see three options:
* change `C` to `TelegrafContext`
* change `C` to `TContext`
* leave it as it is.

For me, one of the last two options is better, because it'll be more understandable that it is generic.
What do you think about this?

dotcypress: @vladimir-fed I think second option is totally fine(`TContext`)

Piterden: Y do u sink so?

typings/index.d.ts
```diff
@@ -336,8 +336,8 @@ export interface ContextMessageUpdate extends Context {
 
 }
 
-export interface Middleware<C extends ContextMessageUpdate> {
-  (ctx: C, next?: () => any): any
+export interface Middleware<TContext extends ContextMessageUpdate> {
+  (ctx: TContext, next?: () => any): any
```
Piterden: But middleware should have next, either you can't continue, isn't it?

PR #553, opened 2018-11-10
fix typings for basic Markup/Extra methods

whitecolor:
There are minor changes to typings, that fixes some issues with basic examples.

Actually, typings do not reflect the state of the code, an none of the examples converted to TS could be compiled correctly because of it.

Advanced things like Stage/Scene/Wizard seem to be not typed at all.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -9,6 +9,11 @@ import * as tt from './telegram-types.d'
 
 export interface TelegramOptions {
   /**
+   * Telegram endpoint API, allow to override default https://api.telegram.org
+   */
+  apiRoot?: Agent
+
+  /**
    * https.Agent instance, allows custom proxy, certificate, keep alive, etc.
    */
   agent?: Agent
@@ -38,6 +43,8 @@ export interface Context {
 }
 
 export interface ContextMessageUpdate extends Context {
+
+  match: string[]
 
   /**
    * Use this method to add a new sticker to a set created by the bot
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -1001,37 +1017,37 @@ export class Markup {
...
 
-  keyboard(buttons: (Buttons | string)[], options?: object): tt.InlineKeyboardMarkup;
+  keyboard(buttons: (Buttons | string)[], options?: object): Markup;
 
   resize(value?: boolean): Markup;
 
   oneTime(value?: boolean): Markup;
 
-  inlineKeyboard(buttons: CallbackButton[] | CallbackButton[][], options: object): tt.InlineKeyboardMarkup;
+  inlineKeyboard(buttons: CallbackButton[] | CallbackButton[][], options?: object): Markup;
```
rubybrowncoat: I'd like to add that inlineKeyboard's buttons also accepts `UrlButton[] | UrlButton[][]`

rubybrowncoat: Or maybe even simply `Buttons[] | Buttons[][]`

ne0nx:
Also, i found that `Markup.keyboard` method has wrong type for first parameter.
Now it's 
`keyboard(buttons: (Buttons | string)[], options?: object): tt.InlineKeyboardMarkup;`
you can call keyboard method only with array but in JS you can pass array of arrays. [https://github.com/telegraf/telegraf/blob/128aa04b9e0cc9b624a780dff99aa9155ccd9943/docs/examples/keyboard-bot.js#L20](example)

it can be `keyboard(buttons: (Buttons | string)[][] | (Buttons | string)[], options?: object): tt.InlineKeyboardMarkup;`

PR #629, opened 2019-03-21
Update README.md

kavayzin:
Judging by the documentation, the sendVoice method takes a voice of the File type, when in practice it is only necessary to transmit file_id. I propose to make corrections to the documentation.


dotcypress:
Regarding [official documentation](https://core.telegram.org/bots/api#sendvoice), `sendVoice` supports files as well.
Could you define `in practice` and how it's relates to API documentation?

kavayzin:
Yes, the mistake was on my side. SendVoice accepts both File, file id and also HTTP URL. Made changes to commit

PR #677, opened 2019-05-30
Add example to inline Keyboard

vladviolenty:


EdJoPaTo:
What is the benefit of this example? Maybe it should be have a more explaining command than 'random2'.

vladviolenty:
command random2 with 2 col in inline message

EdJoPaTo:
Wouldn't it be better to rename the existing random to something like `inlinecolumns` and add the columns there? Then there isn't something with a bad / useless name anymore. Also one example contains columns then.

PR #734, opened 2019-08-19
Stage/BaseScene/Session typings

### LINKED ISSUE #506: Typescript submodules typings
There is no typings for such classess as Markup, Scene, Session, etc.
So how do i declare them without ruinning Telegraf typings ?

I've made telegraf.d.ts and declared them there, but ts still warns me about there is no such module.
This is happening only when i'm also declaring telegraf to extend its interfaces to add props as scene and whatever else.

Any options how to make it right?

mnb3000:
# Description

- Added Stage/BaseScene/Session typings
- Exported Stage and BaseScene from main file

Fixes #506 

## Type of change

- [x] Typings

# How Has This Been Tested?

Simple example scene/stage bot

**Test Configuration**:
* Node.js Version: 10
* Operating System: Linux

# Checklist:

### DIFF
--- a/telegraf.js
+++ b/telegraf.js
@@ -5,6 +5,8 @@ const Composer = require('./composer')
 const Markup = require('./markup')
 const session = require('./session')
 const Router = require('./router')
+const Stage = require('./stage')
+const BaseScene = require('./scenes/base')
 const Context = require('./core/context')
 const generateCallback = require('./core/network/webhook')
 const crypto = require('crypto')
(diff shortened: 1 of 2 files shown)


telegraf.js
```diff
@@ -5,6 +5,8 @@ const Composer = require('./composer')
 const Markup = require('./markup')
 const session = require('./session')
 const Router = require('./router')
+const Stage = require('./stage')
+const BaseScene = require('./scenes/base')
```
dotcypress: I think we don't need to reexport Stage stuff here because node will parse these code even if we don't need it

mnb3000: But we can't add separate typedef for stages/scenes because of the way typescript works

PR #752, opened 2019-09-16
[WIP] Fix weird export

hosein2398:
# Description

Hey,
I've created this PR to discuss weird export in telegraf.js file, so that if we reach to a conclusion I'll fix it and update.(current commit is not necessarily the  fix)
You're using `Object.assign` when exporting, a function which all of it's parameter must be object, but interestingly you're passing a function (class) to it !
```js
module.exports = Object.assign(Telegraf, {
  Composer,
  Extra,
  Markup,
  Router,
  Telegram,
  session
})
```
It's OK for second argument which there are classes in an object and due to object property shorthand those get converted the right way to:
```
{
  Extra: function(){...
}
```
but...
about first argument it's really weird, you're passing a class as a first argument of Object.assign which only accepts object.
And if you just log the output of export you can see the strange output.

#### So if this is wrong why it's just working well ?!
I really have no freaking idea that why this is even working, and what's behavior of Object.assign in that situation.(This could be an interesting case study tho :smile:)

#### How to fix?
I think there would be three solutions.
Since as much as I've seen in codes and docs you're not using any of those exported object classes and only default one which is Telegraf class, so we can:
- Only export the Telegraf (which PROBABLY has anything which is needed, if not, let me know)

Or:
- Set other classes to Telegraf's prototype if those are needed and must be used in user side and only export Telegraf again.

Or:
- Export Telegraf class as default export and others as named export.

What do you think?

## Type of change

Bug fix (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/telegraf.js
+++ b/telegraf.js
@@ -198,22 +198,6 @@ class Telegraf extends Composer {
   }
 }
 
-module.exports = Object.assign(Telegraf, {
-  Composer,
-  Extra,
-  Markup,
-  Router,
-  Telegram,
-  session
-})
+module.exports = Telegraf
 
-module.exports.default = Object.assign(Telegraf, {
-  Composer,
-  Extra,
-  Markup,
-  Router,
-  Telegram,
-  Stage,
-  BaseScene,
-  session
-})
+module.exports.default = Telegraf


hosein2398:
It seems that's for adding static methods to Telegraf class all at once.
And since functions are kinda like objects in js that's why it's working.
But this still looks abnormal to use Object.assign for that, sort of a hack.
Looks unstable to me yet :man_shrugging: , but feel free to close this if you want.

Issue #672, opened 2019-05-28
Heroku incompatibility


TorlanDelta:
* Telegraf.js Version: latest (3.29.0)
* Node.js Version: 10.8.0
* Operating System: win 8.1 64b (but deployed on Heroku)

The following code snippet misbehaves. 

```javascript
amountScene.hears(/^\s*(\d+)\s*$/, async ctx=>{
    console.log(ctx.message);
    await ctx.telegram.deleteMessage(ctx.message.chat.id, ctx.message.message_id)
    console.log('woooosh')
    if (ctx.match[1]>=ctx.session.service[0].MinOrder){
      ctx.session.amount = ctx.match[1]
      ctx.scene.enter('service')
    } else {
      tempMessage(ctx, resp.failedAmount(ctx.session.lang))
    }
  })
```
This handler is supposed to receive a text message like " 120", delete it, and if it complies with certain parameters, return to the previous scene while setting the `session.amount` to the received value. 

Instead this code freezes the handler and it does not  execute `console.log('woooosh')` or anything past it. 

`ctx.message` always has the correct contents and gets output every time. 

**Removing `await` from `ctx.telegram.deleteMessage() `fixes the freeze, but the message still does not get deleted.**

However, if I restart the app right before this handler gets used, the app deletes the message just fine. So like... it only deletes a message once every session. 

I am using `telegraf-session-mysql` btw, but I don't see how that's relevant to `ctx.telegram` working in a weird way.

**With long polling** from local machine **everything works smoothly**, despite a huge latency instilled by my VPN setup. 

Using `bot.launch` method as well as the old way does not make a difference.

Requesting support. 

PS: This one-time "pool" seems to be shared between different telegram functions. For example `ctx.answerCbQuery("hai bishh", true)` also does not get executed except for the first time. And executing it also prevents `deleteMessage()` from working properly. But all my `editMessageMedia` calls do work, interestingly enough. Or `sendPhoto`. So it's a very localized problem.

TorlanDelta:
Will all due respect, bump.

TorlanDelta:
Also to note, the following block of code faces the same complications as the aforementioned ones.
```javascript
userBot.on('text', (ctx) => {
    console.log('WOW SUCH MUCH BIG');
    ctx.reply(ctx.message.text)
  })
```
It works first time, but after that it requires a `heroku restart` to work again. What.Even.

dotcypress:
[...]

PR #756, opened 2019-09-25
Fix answering to webhook query in express app with compression middleware

### LINKED ISSUE #610: Heroku deployment
After deployment to Heroku the bot receives all messages, but doesn't answer.
In case the bot sends two messages consequently in a short period of time, only the second one gets delivered.
When polling is used, all the messages sent by the bot get delivered. I also tried using webhooks without express, but nothing changed.
I found several related issues, but in all of them the author ended up using node-telegram-bot library which solved the problem.

The code used for setting webhook:
```js
const express = require("express");
const expressApp = express();

const bot = new Telegraf(process.env.BOT_TOKEN);
bot.telegram.setWebhook(`${process.env.BOT_URL}/bot${process.env.BOT_TOKEN}`);
expressApp.use(bot.webhookCallback(`/bot${process.env.BOT_TOKEN}`));

bot.use(Telegraf.log());
bot.use(session({
    ttl: 100
}));

expressApp.get("/", (req, res) => {
    res.send("Hello World!");
});
expressApp.listen(process.env.PORT, () => {
    console.log(`Example app listening on port ${process.env.PORT}!`);
});
```

stasusov:
When express with the [compression](https://www.npmjs.com/package/compression) middleware is used, `response.end` gets overridden by [this](https://github.com/expressjs/compression/blob/master/index.js#L92) signature, that doesn't have a callback parameter. Therefore [this promise](https://github.com/telegraf/telegraf/blob/develop/core/network/client.js#L186) never resolves and execution stops.

I found an [old issue](https://github.com/expressjs/compression/issues/46) in the compression Github repo concerning adding the callback parameter, but it ended in a [PR](https://github.com/expressjs/compression/pull/101) from 2017 that is not going to be merged soon.

What is more, there is [shrink-ray](https://www.npmjs.com/package/shrink-ray) that also can be used behaves the same as `compression`.

That PR that simply checks the number of arguments for the `response.end` function and resolve the promise right after `end` if callback is not available.

**May fix** #610, should be tested
**Looks like it fixes** #672, issue creator should test his case.

## Type of change

# How Has This Been Tested?

Express project with webhook handling using express and not the default `.startWebhook()` method. Testing was done with the telegraf code changed directly in the node_modules.
- [x] with `compression` middleware enabled
- [x] with `compression` middleware disabled

**Test Configuration**:
* Node.js Version: 10.13.0
* Operating System: MacOS 10.14.6

# Checklist:

- [ ] I have made corresponding changes to the documentation _(I don't think should make any)_
- [ ] I have added tests that prove my fix is effective or that my feature works _(Unfortunately I have no idea how to test this. However, if you look at the changes, there is pretty toggle)_

### DIFF
--- a/core/network/client.js
+++ b/core/network/client.js
@@ -183,9 +183,15 @@ function answerToWebhook (response, payload = {}) {
     if (!response.headersSent) {
       response.setHeader('content-type', 'application/json')
     }
-    return new Promise((resolve) =>
-      response.end(JSON.stringify(payload), 'utf-8', () => resolve(WEBHOOK_REPLY_STUB))
-    )
+    return new Promise((resolve) => {
+      const params = [JSON.stringify(payload), 'utf-8', () => resolve(WEBHOOK_REPLY_STUB)]
+      if (response.end.length === 2) {
+        response.end(...params.slice(0, 2))
+        params[2]()
+      } else {
+        response.end(...params)
+      }
+    })
   }
 
   return buildFormDataConfig(payload)


stasusov:
@dotcypress, I didn't find any info about who I should assign PR to (and do not have rights for this), but looking at the previous PRs it seems to me that you are a responsible person :)

Please take a look at this PR or assign to one who should do this.

TorlanDelta:
So, when the release? Or will this be another 4 months before these couple of lines will finally be implemented? 

Will be adding it to my live versions to test this out.

dotcypress:
Thank you for PR but decide to fix it [other way](https://github.com/telegraf/telegraf/commit/03683ccedb278628487e0772cb9ee7ee5b82a44e)

Fix will be released this week

PR #781, opened 2019-10-26
fix reply_markup was not passed correctly to body

feed4rz:
# Description

When using telegram client (`telegraf/telegram.js`) method editMessageReplyMarkup directly, reply_markup was not being passed to API call body. Instead it passed Extra instance, which telegram rejected with an error:
```
Bad Request: message is not modified: specified new message content and reply markup are exactly the same as a current content and reply markup of the message
```

Since the code is a little inconsistent, I ve decided to just copy/paste the fix from `editMessageMedia` method.

## Type of change

# How Has This Been Tested?

It was not tested, but it should work, since currently I use this method this way:
```js
await telegram.editMessageReplyMarkup(
  chat_id,
  message_id,
  null,
  Extra.markup(...).reply_markup
)
```

# Checklist:

### DIFF
--- a/telegram.js
+++ b/telegram.js
@@ -300,7 +300,7 @@ class Telegram extends ApiClient {
       chat_id: chatId,
       message_id: messageId,
       inline_message_id: inlineMessageId,
-      reply_markup: markup
+      reply_markup: extra.reply_markup ? extra.reply_markup : extra
     })
   }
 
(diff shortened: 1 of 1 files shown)


dotcypress:
@feed4rz thank you for PR, but I can't merge it, cause it will break API and Typescript typings as well.

Anyway, you can use Markup directly, without Extra wrapper:
```js
await telegram.editMessageReplyMarkup(
  chat_id,
  message_id,
  null,
  Markup.keyboard(...)
)
```

PR #828, opened 2019-12-08
Create self-cleaning-scene.js (working patch)

Piterden:
Added the example


dotcypress:
@Piterden not sure that we need this in examples.

PR #871, opened 2020-01-25
Move typings to devdependencies.

vlakam:
# Description
Moved @types/node and telegram-typings package to devDependencies

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [X] Run Ava

**Test Configuration**:
* Node.js Version: v12.14
* Operating System: Win 10

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,15 +35,14 @@
   },
   "types": "./typings/index.d.ts",
   "dependencies": {
-    "@types/node": "^13.1.0",
     "debug": "^4.0.1",
     "minimist": "^1.2.0",
     "module-alias": "^2.2.2",
     "node-fetch": "^2.2.0",
-    "sandwich-stream": "^2.0.1",
-    "telegram-typings": "^3.6.0"
+    "sandwich-stream": "^2.0.1"
   },
   "devDependencies": {
+    "@types/node": "^13.1.0",
     "ava": "^3.0.0",
     "eslint": "^6.2.2",
     "eslint-config-standard": "^14.1.0",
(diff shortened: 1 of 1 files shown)


EdJoPaTo:
Using `telegraf` with typescript without having `telegram-typings` as a dependency does just fail on compiling typescript as it can not find some types. When this is included in the next release it will break every typescript project using telegraf.

`@types/node` is not required as a dependency so it can stay a devDependency. `telegram-typings` is needed to be installed in order to use telegraf with TypeScript. I suggest adding `telegram-typings` back again to the dependencies.

How to recreate:
1. clone telegraf
2. `npm i`
3. `npm pack` creates the telegraf-3.36.0.tgz
4. go to any typescript project using telegraf
5. `npm i ../path/to/telegraf/telegraf-3.36.0.tgz`
6. `npx tsc` → fails with compiling the TypeScript into JavaScript

Besides that: Testing something that is not a dependency anymore with a devDependency (ava) is not proving anything as you have to install the devDependencies for using Ava in the first place. So the installed packages in total haven't changed at all.
In order to check if a package is really not needed for production / working without devDependencies you have to check that without having them. Because of this ava cant be the right way of testing this.

vlakam:
My bad.
Was sure that i had been running ava from global installation, not locally

EdJoPaTo:
> My bad.
> Was sure that i had been running ava from global installation, not locally

Even if thats the case, I don't think it would make a difference when this dependency is required when using telegraf from another project.
I don't want to blame you, I thought about the very same thing some time ago and found out its just breaking stuff… nearly fell into the same pit :)

PR #877, opened 2020-02-01
formatHTML cosmetic

LetItCode:
# Description

ugly case code

# Checklist:

- [x] Existing unit tests pass locally with my changes

### DIFF
--- a/markup.js
+++ b/markup.js
@@ -167,8 +167,10 @@ class Markup {
     const chars = ['', ...text.split('')]
     entities.forEach(entity => {
       const tag = getHTMLTag(entity)
-      chars[entity.offset] += tag.open
-      chars[entity.offset + entity.length] += tag.close
+      const openPos = entity.offset
+      const closePos = entity.offset + entity.length + 1
+      chars[openPos] += tag.open
+      chars[closePos] = tag.close + chars[closePos]
     })
     return chars.join('')
   }
--- a/markup.js
+++ b/markup.js
@@ -225,7 +225,7 @@ const HTMLTags = new Map([
 
 function getHTMLTag (entity) {
   const tag = HTMLTags.get(entity.type || 'unknown')
-  if (!tag) throw new TypeError('Unknown entity')
+  if (!tag) return { open: '', close: '' }
   return typeof tag === 'function' ? tag(entity) : tag
 }
 
--- a/markup.js
+++ b/markup.js
@@ -164,7 +164,7 @@ class Markup {
   }
 
   static formatHTML (text, entities) {
-    const chars = ['', ...text.split()]
+    const chars = ['', ...text.split('')]
     entities.forEach(entity => {
       const tag = getHTMLTag(entity)
       chars[entity.offset] += tag.open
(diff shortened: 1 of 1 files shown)


markup.js
```diff
@@ -267,4 +201,39 @@ function buildKeyboard (buttons, options) {
...
+function openTag (tag, params) {
+  return params
+    ? `<${tag} ${Object.entries(params)
+        .map(param => `${param[0]}="${param[1]}"`)
+        .join(' ')}>`
+    : `<${tag}>`
+}
+
+function closeTag (tag) {
+  return `</${tag}>`
```
Piterden:
```suggestion
  return `</${tag || ''}>`
```
Maybe this?

LetItCode: I think this is not necessary, the tag will always not be empty. In addition, everyone has a closing pair.

markup.js
```diff
@@ -164,83 +164,15 @@ class Markup {
...
-          case 'text_mention':
-          case 'text_link':
-            result.push('</a>')
-            break
-        }
-        opened.splice(index, 1)
-      }
-    }
-    return result.join('')
+    const chars = ['', ...text.split(''), '']
```
GingerPlusPlus: Great code, however each character from `text` should be html-escaped:

```js
function escapeChar (c) {
  switch (c) {
    case `&`: return '&amp;'
    case `"`: return '&quot;'
    case `'`: return '&#39;'
    case `<`: return '&lt;'
    default : return c
  }
}
```

```suggestion
    const chars = ['', Array.prototype.map.call(text, escapeChar), '']
```

markup.js
```diff
@@ -271,4 +203,30 @@ function buildKeyboard (buttons, options) {
   return result
 }
 
+function tag (name, params) {
+  return {
+    open: params
+      ? `<${name} ${Object.entries(params).map(([key, value]) => `${key}="${value}"`).join(' ')}>`
```
GingerPlusPlus: Attribute values should be escaped as well:
```suggestion
      ? `<${name} ${Object.entries(params).map(([key, value]) => `${key}="${value.replace(/[<&"]/g, escapeChar)}"`).join(' ')}>`
```

On 2nd thought, no need to escape apostrophe.

markup.js
```diff
@@ -271,4 +203,30 @@ function buildKeyboard (buttons, options) {
...
+  ['pre', tag('pre')],
+  ['strikethrough', tag('s')],
+  ['underline', tag('u')],
+  ['text_link', ({ url }) => tag('a', { href: url })],
+  ['text_mention', ({ user }) => tag('a', { href: `tg://user?id=${user.id}` })]
+])
+
+function getHTMLTag (entity) {
+  const tag = HTMLTags.get(entity.type || 'unknown')
+  if (!tag) throw new TypeError('Unknown entity')
```
GingerPlusPlus: This will throw if [`entity.type`] is “mention”, “hashtag”, “cashtag”, “bot_command”, “url”, “email” or “phone_number”.
```suggestion
  if (!tag) return { open: '', close: '' }
```

[`entity.type`]: https://core.telegram.org/bots/api#messageentity

markup.js
xamgore: Maybe `['', ...text.split(''), ''].map(escapeChar)`?

Issue #847, opened 2019-12-30
Change type in RestrictChatMember


B4TT3RY:
## Context

* Telegraf.js Version: 3.34.1
* Node.js Version: 12.13.1
* Operating System: Windows 10

## Expected Behavior

![image](https://user-images.githubusercontent.com/12368101/71565881-381d7f00-2af6-11ea-9e9f-ed878ef188fc.png)

## Current Behavior

![image](https://user-images.githubusercontent.com/12368101/71565894-4d92a900-2af6-11ea-9d15-636019e66771.png)

until_date isn't a boolean. Please change it to number.

B4TT3RY:
When will it change?

PR #887, opened 2020-02-04
Fix until date type in index.d.ts

Piterden:
https://github.com/telegraf/telegraf/issues/847


Piterden:
One line === one commit

PR #915, opened 2020-02-21
When entering a scene, while in a scene: push the former onto a stack

arthurphilippe:
# Description

I found that it can be interesting not to leave a scene when entering a new one but rather save its state onto a stack for future reuse.

I do not not if that feature is of interest for most people, but thought I'd share.

## Type of change

# How Has This Been Tested?

- [x] Added unit test file for scenes under `tests/scenes.js`: contains one test that triggers entrance to two scenes and expect the first scene to be active even after having entered and left the second.

**Test Configuration**:
* Node.js Version:
* Operating System:

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "telegraf",
-  "version": "3.36.0",
+  "name": "telegraf_acp_fork",
+  "version": "3.36.2",
   "description": "Modern Telegram Bot Framework",
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


wojpawlik:
I don't dislike this idea, just, it would need to be reimplemented in TS. While at it, why not introduce separate stack methods (`push`, `pop`...) instead of optional flag argument?

PR #929, opened 2020-02-29
using static method Date.now()

LetItCode:
save energy, stop Greta Thunberg panic


jacobwaller:
Good catch. Date.now() is about twice as fast. (Not that the half a millisecond matters that much lol)

PR #922, opened 2020-02-27
[Typescript] Fix types for Telegraf.Context and Composer handlers

frontendmonster:
# Description

Hi, and thank you all for this great library.

I've used the Telegraf for a while with typescript and find that typescript typings are poor.

I had two major problems with typing, 
* Custom context #920 
* pass composer as a handler #921. 

I've created this PR to cover those issues.

### Custom Context Typings

Custom context solved by adding `contextType` to `TelegrafOptions` type.  i've I've change `TelegrafOptions` to generitc type to pass `Context` to `contextType`.

(also, I found `retryAfter` and `handlerTimeout` are missing `TelegrafOptions` props)

finally, I've added `TelegrafConstructor.Context`.

### Handler Typings

For #921, I found that all middleware args can be also a `composer` too. So, I create a `Handler` type and replace it with all handler arguments in the `Composer` members. 

```typescript 
export type Handler<T extends ContextMessageUpdate> = Middleware<T> | Composer<T>

// ...

export interface Composer<TContext extends ContextMessageUpdate> {
  // ...
  use(middleware: Handler<TContext>, ...middlewares: Array<Handler<TContext>>): Telegraf<TContext>
  // ...
}
```

Fixes #920 #921 

## Type of change

Please delete options that are not relevant.

- [ ] Documentation (typos, code examples or any documentation update)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

I've added all the use-cases to the `test.ts` type and it compiled successfully.

- [x] Handler Types
- [x] Custom Composer Types

**Test Configuration**:
* Node.js Version: 13.5.0
* Operating System: GNU/Linux 5.4.18
* Typescript Version: 3.0.1

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #445, opened 2018-06-19
Add default export to support Typescript import

DenisCarriere:
# Description

Add default export to support Typescript import

**Typescript syntax**
```ts
import Telegraf from "telegraf";

const bot = new Telegraf(process.env.BOT_TOKEN)
```

**Error message (Issue)**

```
$ ts-node index.ts

/Users/mac/Github/telegraf-bot/index.ts:3
const bot = new Telegraf(process.env.BOT_TOKEN)
            ^
TypeError: telegraf_1.default is not a constructor
```

## Type of change

Fixes Typescript, without this change you can't use Typescript with Telegraf

PR #960, opened 2020-04-01
Overhaul typings

### LINKED ISSUE #763: VSCode intelliSense isn't working with telegraf
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Telegraf.js Version: 3.32.0
* Node.js Version: 8.11.3

## Expected Behavior

Showing hints and autocompleting code

## Current Behavior

Not showing hints for telegraf.js, it works for any other node.js module

### LINKED ISSUE #920: [TypeScript] Property 'Context' does not exist on type 'TelegrafConstructor'
## Context

* Telegraf.js Version: 3.36.0
* Node.js Version: 13.5.0
* Operating System: GNU/Linux 5.4

## Expected Behavior

`Telegraf.Context` type should be defined as `ContextMessageUpdate`.

## Current Behavior

Typescript can not resolve the right type for `Telegraf.Context`.

### Steps to Reproduce

1. follow [custom-context example](https://github.com/telegraf/telegraf/blob/develop/docs/examples/custom-context-bot.js) in ts file.

### Failure Logs

Property 'Context' does not exist on type 'TelegrafConstructor'

### LINKED ISSUE #921: [Typescript] bot.use Argument of type 'Composer<Context>' is not assignable to parameter of type 'Middleware<Context>'.
## Context

* Telegraf.js Version: 3.36.0
* Node.js Version: 13.5.0
* Operating System: GNU/Linux 5.4

## Expected Behavior

`bot.use` signature should accept the `Composer` type as arguments.

## Current Behavior

Typescript type definition for `bot.use` arguments only accepts `Middleare`.

### Steps to Reproduce

1. create a `Composer` instance.
2. create a `Telegraf` instance.
3. try to add composer to bot middlewares.

### Failure Logs

Argument of type 'Composer<Context>' is not assignable to parameter of type 'Middleware<Context>'.

### LINKED ISSUE #943: typings: incorrect bindings
The typings are not working correctly as expected
the IntelliSense is working when `const Telegraf = require('telegraf').Telegraf` is done which is not correct and its throwing error
but it does shows suggestions in vscode
image
 
working behavior should be without `require('telegraf').Telegraf` this extra `.Telegraf` thing

* Telegraf.js Version: 3.36.0
* Node.js Version: 13.11.0
* Operating System: ArchLinux

## Expected Behavior
it should work without `.Telegraf import`. but in this case its not showing with normal method provided in documentation 
image

## Current Behavior
Current behaviour throws error with `const Telegraf = require('telegraf').Telegraf`
```js
const bot = new Telegraf(process.env.BOT_TOKEN);
            ^

TypeError: Telegraf is not a constructor
    at Object.<anonymous> (/home/sonisins/prujekts/node/gamerary-tg-bot/bot.js:4:13)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
    at Module.load (internal/modules/cjs/loader.js:996:32)
    at Function.Module._load (internal/modules/cjs/loader.js:896:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47

```

### LINKED ISSUE #970: `replyWithPoll` options not available in TypeScript
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Telegraf.js Version:
* Node.js Version:
* Operating System:

## Expected Behavior

`replyWithPoll` option available on ctx object.

## Current Behavior

Copying [example](https://github.com/telegraf/telegraf/blob/master/docs/examples/poll-bot.js) errors when trying to run with Typescript.

### Steps to Reproduce

1. Create bot
```ts
import Telegraf, { Extra } from "telegraf";

bot.start((ctx) => ctx.reply("Welcome"));
bot.command('poll', (ctx) =>
  ctx.replyWithPoll(
    'Your favorite math constant',
    ['x', 'e', 'π', 'φ', 'γ'],
    { is_anonymous: false }
  )
)
bot
  .launch()
  .then(() => console.log("Bot Started!"))
  .catch((e) => console.error("Uh oh, bot didn't start: ", e.toString()));
```

2. Run bot
`$ ts-node src/bot.ts`

### Failure Logs
```
TSError: ⨯ Unable to compile TypeScript:
src/bot.ts:44:7 - error TS2551: Property 'replyWithPoll' does not exist on type 'ContextMessageUpdate'. Did you mean 'replyWithPhoto'?

44   ctx.replyWithPoll(
         ~~~~~~~~~~~~~

  node_modules/telegraf/typings/index.d.ts:260:3
    260   replyWithPhoto(photo: tt.InputFile, extra?: tt.ExtraPhoto): Promise<tt.MessagePhoto>
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    'replyWithPhoto' is declared here.
```

### LINKED ISSUE #987: typescript: buttons require hide argument though it is optional
https://github.com/telegraf/telegraf/blob/11c3105e69e2878c5979cf54215dd8f3c9960097/markup.js#L42-L48
From the source code it's not hard to see that `options` argument in 
https://github.com/telegraf/telegraf/blob/11c3105e69e2878c5979cf54215dd8f3c9960097/typings/index.d.ts#L1383
is optional.
Although, it's not.

### LINKED ISSUE #991: typescript: editMessageText doesn't return edited message on success
From `index.d.ts`:
https://github.com/telegraf/telegraf/blob/11c3105e69e2878c5979cf54215dd8f3c9960097/typings/index.d.ts#L350-L356
Right now, an edited message isn't returned.

I expect this function to return an edited message as stated in docs.

wojpawlik:
# Description

Follow up to https://github.com/telegraf/telegraf/issues/943#issuecomment-606540135.

Closes #922.
Closes #932 (backported).
Closes #992 (backported).
Closes #988 (backported).
Fixes #763.
Fixes #920.
Fixes #921.
Fixes #943.
Fixes #970.
Fixes #991.
Fixes #987.

PS. Consider auto-formatting typings with `prettier --write "**.ts"` (`{ "semi": false, "singleQuote": true }`)

## Type of change

- Documentation (types)
- Breaking change (incompatible with current types)

# How Has This Been Tested?

- `npm run typecheck`
- `npx -p gist:885dbe80be82e706e9e84c967269e17b mistypes telegraf.js --- typings/index.d.ts`
- Fixes 27 typing errors in https://github.com/thedevs-network/the-guard-bot

**Test Configuration**:
* Node.js Version: v12.3.1
* Operating System: Windows_NT x64 10.0.18363

# Checklist:


wojpawlik:
Why the holdup @dotcypress?

wojpawlik:
@dotcypress I merged #964 and #965. I'd appreciate getting this merged so that I can start building on top of it instead of resolving merge conflicts again. I'm happy to address any questions or concerns you may have.

typings/markup.d.ts
```diff
@@ -0,0 +1,172 @@
...
+
+interface Buttons {
+  url?: string
+  pay?: boolean
+  text: string
+  callback_data?: string
+  callback_game?: string
+  switch_inline_query?: string
+  switch_inline_query_current_chat?: string
+}
```
EdJoPaTo, line 64:
```suggestion
export type Buttons = Button | ContactRequestButton | LocationRequestButton | UrlButton | CallbackButton | SwitchToChatButton | SwitchToCurrentChatButton | GameButton | PayButton
```

This should be a simple 'any of the buttons' type

typings/markup.d.ts
```diff
@@ -0,0 +1,172 @@
+import * as tt from './telegram-types.d'
+
+interface Button {
```
EdJoPaTo, line 3: Each of these buttons should be exported in order to use the interface externally.

Super handy when creating keyboards like this:
```ts
const buttons: UrlButton[][] = []
buttons.push(…)
Markup.inlineKeyboard(buttons)
```

typings/telegram.d.ts
```diff
@@ -0,0 +1,660 @@
+import * as tt from './telegram-types.d'
+
+import * as https from 'https'
+import * as http from 'http'
+
+interface AdminPerms {
```
EdJoPaTo, line 6: I think interfaces like this should be exported in order to be declared before used with types

```ts
const param: AdminPerms = {…}
ctx.method(…, params, …)
```

typings/index.d.ts
```diff
@@ -1,1567 +1,31 @@
...
-   */
-  channelMode?: boolean
-
-  retryAfter?: number
-
-  handlerTimeout?: number
 }
 
-export default Telegraf
+export = Telegraf
```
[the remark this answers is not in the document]
wojpawlik: This would unnecessarily export a lot of internal types. I curated `index.d.ts` to only export types for stuff exported by `telegraf.js`, and `Middleware`.
Also, the workaround is not needed in your code since there's no declaration merging.

typings/index.d.ts
```diff
@@ -1,1567 +1,31 @@
...
+import { Telegraf as _Telegraf } from './telegraf'
+declare class Telegraf<TContext extends ContextMessageUpdate> extends _Telegraf<
+  TContext
+> {}
+
+declare namespace Telegraf {
+  export {
+    BaseScene,
+    Composer,
+    ContextMessageUpdate as Context,
```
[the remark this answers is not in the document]
wojpawlik: It's not "not perfect", it's outright incorrect: #920.

I never tried to hide that my changes are breaking. I'm open to exporting it as both `Context` and `ContextMessageUpdate`, but only if @dotcypress allows `telegraf.js` to do the same.

typings/index.d.ts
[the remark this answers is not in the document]
EdJoPaTo: I dont only create bots, I also create middlewares for Telegraf which sometimes need internal types. Also things like Markup contain a lot of types which have to be exported. Having to list them each manually here to is prone to errors.

typings/index.d.ts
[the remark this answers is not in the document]
EdJoPaTo: You are right: telegraf.js is exporting it as `Context` so ContextMessageUpdate is wrong in the Typescript typings.
https://github.com/telegraf/telegraf/blob/252513a1e80aa444c88e2a33e056cf96023c61d2/telegraf.js#L210

Looks like it has to be breaking the types as it has to do the same as the JavaScript Code does.
As it is `Context` the interface `ContextMessageUpdate` should be renamed to `Context` instead of using `export as` in my opinion.

Using an additional `export {Context as ContextMessageUpdate}` would be nice to not break everything with a simple update

typings/index.d.ts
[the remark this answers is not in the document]
EdJoPaTo: Also: whats the problem with exporting types and interfaces? Typescript is there to ensure type safety. Not being able to use types because they are inaccessible is TypeScript as useless as JavaScript. When they are exported and you dont need them you still have the option to not import them.

typings/markup.d.ts
wojpawlik: I'll do it. Just tell me, what is the `hide` field in `interface Button`?

typings/markup.d.ts
wojpawlik: Wouldn't the union of them all be enough?

typings/index.d.ts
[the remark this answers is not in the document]
wojpawlik: It clutters suggestions, making the library look more intimidating that it actually is.

typings/markup.d.ts
EdJoPaTo: When hide is true the buttons are not placed in the keyboard.
https://github.com/telegraf/telegraf/blob/252513a1e80aa444c88e2a33e056cf96023c61d2/markup.js#L184-L186

```ts
Markup.inlineKeyboard([
  [Markup.callbackButton('hey', 'x', Math.random() > 0.5)]
])
```
is the equivalent of
```ts
const buttons: Buttons = []
if (Math.random() > 0.5) {
  buttons.push([Markup.callbackButton('hey', 'x')])
}
Markup.inlineKeyboard(buttons)
```

typings/test.ts
```diff
@@ -1,10 +1,11 @@
 // This is a test file for the TypeScript typings.
 // It is not intended to be used by external users.
-import Telegraf, { Markup, Middleware, ContextMessageUpdate, Extra } from './index';
+import Telegraf = require('./index')
+import { Markup, Middleware, Context } from './index';
```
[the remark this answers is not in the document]
wojpawlik: I tried (and failed) to add `Telegraf as default` to exports, PR to my PR welcome.

typings/markup.d.ts
EdJoPaTo: Being more specific is often helpful.
For example: UrlButtons are kept when a message is forwarded. If there are other Buttons than UrlButtons within the same keyboard, all buttons are dropped when forwarded.

I like to use a typed language to have specific types so the compiler can prevent me doing mistakes later on. The use cases for these are often not clear when designing a library so the user of the library should have as many options available as possible. It shouldn't prevent the users doing what they like to do unless its not possible for some reason.

.mailmap
```diff
@@ -0,0 +1 @@
+Wojciech Pawlik <woj.pawlik@gmail.com>
```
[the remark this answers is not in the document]
wojpawlik: [It's useful](https://www.git-scm.com/docs/git-check-mailmap#_mapping_authors).

typings/index.d.ts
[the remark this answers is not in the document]
wojpawlik: > using `export = Telegraf` will require all repositories which use Telegraf to include `esModuleInterop` in the tsconfig.

I am `require`'ing the module with improved typings without it, no issues.
Also, see https://github.com/telegraf/telegraf/issues/763#issuecomment-538671190 and #943 -- situation is similar to `Context`.

typings/markup.d.ts
wojpawlik: Alright. I'm fine with them being exported from `markup.d.ts`, but not `index.d.ts`.

.mailmap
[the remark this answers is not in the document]
EdJoPaTo: I know mailmaps but I dont think its something needed on GitHub projects as GitHub keeps the mapping of mails to users anyway: https://github.com/telegraf/telegraf/graphs/contributors

Also its unrelated to the point of this PR: improving the type situation of telegraf.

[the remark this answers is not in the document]
wojpawlik: Yes, Github won't benefit, but git cli and good git tools will.
I just don't wanna look like 2 people to those, is that too much to ask?

I totally agree that it's unrelated, but PR adding just a `.mailmap` entry would look stupid.

typings/index.d.ts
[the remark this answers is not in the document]
EdJoPaTo: I think the `export default Telegraf` is correct as long as a library has a `module.export.default` (if i understand [this](https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default) correctly). The default was added in #445 to use it as a default import.

Based on that im not sure the default export is what is useful here and probably also confusing VSCode Intellisense.

Doing a step back from Telegraf and looking at the way it is done elsewhere. In JavaScript / CommonJS I would use a module like this: 
```js
const fs = require('fs')
fs.writeFileSync(…)
```

ES6 modules are more into this way of doing things:
```js
import * as fs from 'fs'
fs.writeFileSync(…)
// or
import {writeFileSync} from 'fs'
writeFileSync(…)
```

Back to Telegraf. Seeing this I would assume I should import it like this?
```js
import {Telegraf, Markup} from 'telegraf'
// or
import * as Telegraf from 'telegraf'
```

When there is a default export like we have here I should be able to do this:
```js
import Telegraf, {Markup} from 'telegraf'
```

Seeing that i would assume there is a default export besides the other exports like this:
```js
export default Telegraf;
export {Markup} from …;
```

Also I don't understand why the namespace is there. We are already in the TypeScript module 'Telegraf' so why is there an additional namespace for that? What is its benefit?
Taking a look at the TypeScript Module Documentation it also lists top level namespaces as a [red flag](https://www.typescriptlang.org/docs/handbook/modules.html#red-flags).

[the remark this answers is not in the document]
wojpawlik: Thank you for your persistence on this.
Since it's impossible to provide maximally convenient typings for both CommonJS and TypeScript consumers, I deprioritized CommonJS and significantly simplified my typings as a result. I also added `Telegraf` and `ContextMessageUpdate` aliases for backwards compatibility.

telegraf.js
```diff
@@ -208,11 +208,13 @@ class Telegraf extends Composer {
 
 module.exports = Object.assign(Telegraf, {
   Context,
+  ContextMessageUpdate: Context,
```
EdJoPaTo, line 211: Im not sure the JavaScript files should export a type which was wrongly introduces by TypeScript?

typings/context.d.ts
```diff
@@ -0,0 +1,480 @@
+import * as tt from './telegram-types'
+
+import { AdminPerms, Telegram } from './telegram'
+
+export declare class ContextMessageUpdate {
```
EdJoPaTo, line 5: I still think this should be renamed to Context internally as the JavaScript Code calls it Context.

The name `ContextMessageUpdate` should only be in `export {Context as ContextMessageUpdate}` and thats only for legacy reasons.
As you pointed out correctly the name does not exist in the JavaScript Code so it shouldnt exist in TypeScript either.

typings/telegraf.d.ts
```diff
@@ -0,0 +1,213 @@
...
+   */
+  allowedUpdates?: tt.UpdateType[] | tt.UpdateType | null
+
+  /**
+   * Polling stop callback
+   */
+  stopCallback?: () => void | null
+}
+
+interface LaunchWebhookOptions {
```
EdJoPaTo, line 48: I think these interfaces should also be exported (in this file, not necessarily from index)

readme.md
```diff
@@ -41,7 +41,7 @@ $ yarn add telegraf
 ### Examples
   
 ```js
-const Telegraf = require('telegraf')
+const { Telegraf } = require('telegraf')
```
EdJoPaTo, line 44: I think with `export {Telegraf as default}` introduced in e3d3589196587ff88a4d8231bfe268fcb4e21dce we can still recommend usage as `const Telegraf = require('telegraf')`. Or do I miss something?

EdJoPaTo:
> Additional change request in a file you didnt changed so I could pin it to a file: tsconfig should use include instead of files as there are more files now.
> 
> ```json
>   "include": [
>     "typings"
>   ]
> ```

Please also update the tsconfig.json so typescript will include all the files instead of only the old three which are partially gone now.

.mailmap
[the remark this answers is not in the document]
EdJoPaTo: A mailmap for a repo with 100 contributors where most of the code / commits were introduced by a single person containing a different person only looks narcissistic to me.
But yeah, someone has to start with it. As its something unrelated to the purpose of the PR I still think it shouldn't be added here. But if @dotcypress actively wants to have it, why not?

typings/index.d.ts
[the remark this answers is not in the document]
EdJoPaTo: Looking on other JavaScript projects like express with typings added via @types packages they mostly are used this way:

```ts
import express as * from 'express'
const app = new express()
```

DefinitelyTyped is relatively strict about not using a default export.

But as Telegraf already had a default export for years I think the state we are in with this PR is quite good now.

readme.md
wojpawlik: Actual CommonJS eqivalent is weirdly looking `const Telegraf = require('telegraf').default`. 

```js
const Telegraf = require('telegraf')
new Telegraf(token)
```
runs correctly, but makes TS complain:
```
This expression is not constructable.
  Type 'typeof import("telegraf/typings/index")' has no construct signatures.ts(2351)
```

This is exactly what I was trying to prevent with my overly complex `index.d.ts`.

typings/telegraf.d.ts
wojpawlik: I don't think it's necessary for a function you only call once. But whatever, I'll do it anyway.

typings/context.d.ts
wojpawlik: I agree. However I'm worried about making this PR unreviewable (and thus unmergable) by changing too much.

readme.md
EdJoPaTo: If it runs correctly for JavaScript users I would continue to suggest it that way.
For TypeScript users I wouldn't suggest using require as import from is the way to go. 

But im not sure what VisualStudio Code autocomplete does when the simple require() is used in JavaScript.

If we change the overall usage towards `const {Telegraf} = require('telegraf')` or `import {Telegraf} from 'telegraf'` then the default import can be removed...

I think the main goal should be to have as little changes to JavaScript users (while autocomplete and stuff like that works) as possible. The second goal should be to have TypeScript types which are reflecting as close as possible what the JavaScript code does. If the types are not working `as any` is a common way to handle telegraf currently anyway (Markup and Extra for example) as it falls back to the stable working JavaScript code.

typings/telegraf.d.ts
EdJoPaTo: I don't care about the specific interfaces. They have some content. If I have to initialize larger objects I start to split them up. While doing so I want to have type checking as early as possible.
I don't know if that might happen with exactly these but they are publicly facing interfaces meant to be used by the user so they should be public.

telegraf.js
wojpawlik: Without this, typings would lie about the class being accessible under this name.

typings/context.d.ts
wojpawlik: `SceneContextMessageUpdate` would have to be renamed as well, probably to `SceneContext`.

readme.md
wojpawlik: > But im not sure what VisualStudio Code autocomplete does when the simple require() is used in JavaScript.

It sees it as a namespace, but not a class.

To make it see it as a class as well, `export=` has to be used, but then no other exports are allowed, which forced me to awkwardly squeeze them into otherwise unneccesary namespace.

wojpawlik:
I noticed your suggestion to update `tsconfig`. My answer: works, don't touch :P

wojpawlik:
Thank you for your reviews @EdJoPaTo. I'd appreciate feedback on `composer.d.ts`.

typings/telegram-types.d.ts
```diff
@@ -179,6 +179,32 @@ export type InputFile =
  */
 export type InputFileVideoNote = Exclude<InputFile, InputFileByURL>
 
+export interface AdminPerms {
```
Piterden, line 182: Why does it call `AdminPerms`? There isn't that type in bots API docs, where the same called `ChatPermissions`.

Even if Admin word is more relative to the context than Chat, anyway shortening is not a good when all other types called with their full names!

typings/telegram-types.d.ts
```diff
@@ -522,6 +580,10 @@ export interface MessageDice extends TT.Message {
   dice: Dice
 }
 
+export interface MessagePoll extends TT.Message {
+  poll: Poll
+}
+
 export interface NewInvoiceParams {
```
Piterden, line 587: The same to here! Unneeded shortening.

EdJoPaTo:
> I noticed your suggestion to update `tsconfig`. My answer: works, don't touch :P

`telegram-types.d.ts` Is referenced which isn’t existing anymore.

This PR tries to bring types to a better state so it should do it.
The current types are „working somehow“ too, so don’t touch them. :P

typings/telegraf.d.ts
EdJoPaTo: If a user is able to use the interface it should be exported.

typings/context.d.ts
EdJoPaTo: I think reviewing coming from JavaScript is easier when looking into each file at a time. Comparing `context.js` with `context.d.ts` and so on. Having the same names only makes it easier to do so.

As this is splitting one file onto many it is already hard to review just by looking into the changes.

Looking into specific commits doing only small things also helps a lot.

Renaming Types does not change much from the functionality side of view so I think it should be done now in a bunch. Then only this PR is breaking and requiring manual work to do after updating to it.

telegraf.js
EdJoPaTo: The current set of typings are lying all over the place with their current names regarding the JavaScript code and their names in the Telegram API.

This PR should try to fix this situation.
It should only change types and do not interfere with JavaScript. If types have to change the JavaScript then the types have made something very wrong.

If you think this is required to add to module.exports then I prefer not to `export {Context as ContextMessageUpdate}` at all and only export the correct names of things.

wojpawlik:
[I read up](https://github.com/microsoft/TypeScript/wiki/Performance#configuring-tsconfigjson-or-jsconfigjson), and I don't see any benefit in switching from just entry point in `files` to `include`.

typings/telegram-types.d.ts
wojpawlik: You mean, you'd prefer `AdminPermissions`, or maybe `ExtraPromote`, for consistency with `telegram-types.d.ts`?

EdJoPaTo: https://core.telegram.org/bots/api#chatpermissions

EdJoPaTo:
> The primary difference between the two is that files expects a list of file paths to source files, and include/exclude` use globbing patterns to match against files.

I prefer including the typings folder over each file in the typings folder as an array as its wrong again when adding / removing a file later on.
That's why I suggested it.

typings/telegram-types.d.ts
EdJoPaTo: right.. its not ChatPermissions, its extra of [promoteChatMember](https://core.telegram.org/bots/api#promotechatmember)

so `ExtraPromoteChatMember` seems more fitting as @wojpawlik suggested

typings/telegram-types.d.ts
wojpawlik: Note that I didn't picked up this name, it was already like this. Now that you mentioned it, I will improve it while renaming other stuff, but please don't expect me to fix the whole world in a single PR.

wojpawlik:
TS auto-includes each imported file, so it's only necessary to list entry points in `files`.

typings/composer.d.ts
```diff
@@ -0,0 +1,197 @@
...
+type HearsTriggers = string[] | string | RegExp | RegExp[] | Function
+
+export interface MiddlewareFn<TContext extends ContextMessageUpdate> {
+  /*
+  next's parameter is in a contravariant position, and thus, trying to type it
+  prevents assigning `MiddlewareFn<ContextMessageUpdate>`
+  to `MiddlewareFn<CustomContext>`.
+  Middleware passing the parameter should be a separate type instead.
+  */
+  (ctx: TContext, next: () => Promise<unknown>): unknown
```
EdJoPaTo, line 15: I am thinking about best practices here...

As a first idea I would say this should be kind of strict: `(ctx: TContext: next: () => Promise<void>): void | Promise<void>`.
But its neat to do quick one liners like this:
```js
bot.command('start', ctx => ctx.reply('go away'))
```

In that case `ctx.reply` returns `Promise<Message>` which can not be applied to `Promise<void>`.
So `Promise<unknown>` would be better for these one liners.
I currently do not see any point for something returned which isnt a Promise. So `void | Promise<unknown>` as a return type seems quite good for me.

Looking at next: next should not return something? Yeah, its possible and `next` can return something but that feels hacky and not intended. ([If someone wants to get stuff from one middleware to the next they should use `ctx.state`.](https://telegraf.js.org/#/?id=state)) So I think `() => Promise<void>` would represent the purpose of `next` better?

I do not really understand the comment you added about the next parameter. It seems unrelated to the next parameter? The next parameter is not changing when the TContext is a different one.

Based on that I would suggest this:

```suggestion
  (ctx: TContext, next: () => Promise<void>): void | Promise<unknown>
```

wojpawlik: [`void` is weird](https://github.com/microsoft/TypeScript/issues/20006#issuecomment-344410732); I don't understand it and thus I don't trust it enough to write `Promise<void>`.

If you want stricter type, why allow returning `void` from `MiddlewareFn`? Why not just `Promise<unknown>`?

EdJoPaTo: First void: Its the absence of any type at all. Return type -> You except something to return. void as return type -> you expect the absence of anything. Which sounds more fitting to me than unknown: there is something but i don't know what it is. next shouldnt return anything. Technically it can when Middlewares are badly implemented but its not planned that way.

Second returning void.  When doing stuff locally or when doing long tasks in parallel for example:
```ts
bot.action('whatever', ctx => {
  ctx.session.lastAction = 'whatever'
})

bot.action('whatever', ctx => {
  setTimeout(() => ctx.reply('hey you!'), 1000)
})
```

wojpawlik: I see. Still I'm not convinced if the benefits outweight the temptation to just not `await`:

```ts
ctx => {
  const x = "result of some calculation"
  ctx.reply(x) // no await, no return, error can't be caught
}
```

`state` is currently untyped and I have no idea how it could ever be well-typed; I'm not sure if that's the best way for middlewares in TypeScript to communicate.

EdJoPaTo: You can use linters like typescript-eslint with [rules preventing floating promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md)

Also something like .catch can be used to prevent unhandled rejections:
```ts
ctx => {
  const x = 'result'
  ctx.reply(x).catch(error => console.error(error))
}
```

I use this when debouncing many user messages.
The user sends a bunch of messages which get saved and then replied to when there was no new message for some time.

Another use case is my chat record bot: Look at each message in a group and save it. It does not respond on any message. When the user uses /finish after an hour or so the whole chat log is written into a file and sent to the chat. (In this case saving is done asynchron so I want to await but this can be done with writeFileSync or something like that too.)

EdJoPaTo: The type of state is based on middlewares unrelated to telegraf itself. Telegraf itself does not use it so it shouldnt type it.

I would expect something like this on the user side:

```ts
import {Context as TelegrafContext} from 'telegraf'
interface MyContext extends TelegrafContext {
  state: {
    userSeemsFine: boolean;
  }
}

const bot = new Telegraf<MyContext>(token)
bot.use(async (ctx, next) => {
  ctx.state = {
    userSeemsFine = Math.random() > 0.5
  }
  return next()
})
bot.command('whatever', async ctx => ctx.reply(ctx.state.userSeemsFine ? 'yaay' : 'oh no'))
```

Personally I prefer to keep things externally as it does not require the context of the specific Telegram update so its something unrelated to the Context -> keep it separate.

```ts
const userSeemsFine: Map<number, bool> = new Map()

bot.use(async (ctx, next) => {
  if (ctx.from) {
    userSeemsFine[ctx.from.id] = Math.random() > 0.5
  }

  return next()
})
bot.command('whatever', async ctx => ctx.reply(userSeemsFine[ctx.from!.id] ? 'yaay' : 'oh no'))
```

typings/context.d.ts
```diff
@@ -350,17 +350,6 @@ export declare class ContextMessageUpdate {
   // ------------------------------------------------------------------------------------------ //
   // ------------------------------------------------------------------------------------------ //
 
-  /**
-   * Use this method to send answers to an inline query. On success, True is returned.
-   * No more than 50 results per query are allowed.
-   * @param results Array of results for the inline query
-   * @param extra Extra optional parameters
-   */
-  answerInlineQuery(
```
EdJoPaTo: answerInlineQuery exists: https://github.com/telegraf/telegraf/blob/646476197d2c6b7e02bea0ba6a9f03da67ac2809/context.js#L176-L179

typings/context.d.ts
EdJoPaTo: I just realized `Context` or `ContextMessageUpdate` is named `TelegrafContext` in the Sourcecode
https://github.com/telegraf/telegraf/blob/646476197d2c6b7e02bea0ba6a9f03da67ac2809/context.js#L53

typings/context.d.ts
EdJoPaTo: Wait... the typings included it twice so a472f80698e6e87a5655da62327cc88af14b92dd was only deduplicating it. My bad

typings/composer.d.ts
wojpawlik: > I do not really understand the comment you added about the _`next`_ parameter. [...] The _`next`_ parameter is not changing when the TContext is a different one.

`next`_'s_ parameter.

wojpawlik:
Notice me @dotcypress

wojpawlik:
This will never go anywhere without your approval, @dotcypress, so what's wrong?

If you're reluctant to deal with typings, how about I move them to `@types/telegraf`?

Please don't force me to maintain a fork.

dotcypress:
@wojpawlik I'm ready to merge this PR any time. Just do a rebase and ping me here

wojpawlik:
@dotcypress I merged develop.

Is rebasing really necessary? I'm worried I may need to resolve conflicts with `develop` again, or even with myself, in order to obtain clear history out of this. I'd squash and merge.

dotcypress:
@wojpawlik Thank you! 🚀

The next release is coming soon!

typings/markup.d.ts
wojpawlik: How about exporting this from `index.d.ts`?

```ts
export namespace KeyboardButton {
  export interface Text {...}
  export interface ContactRequest {...}
  export interface LocationRequest {...}
  export interface PollRequest {...}
}
export type KeyboardButton =
  | KeyboardButton.Text
  | KeyboardButton.ContactRequest
  | KeyboardButton.LocationRequest
  | KeyboardButton.PollRequest
  | string
```

All union members would be available but wouldn't clutter suggestions.

PR #1000, opened 2020-04-23
Fixup poll typings

EdJoPaTo:
# Description

Typings suggest that every update contains a poll which isn't the case. This was probably just overlooked in a hurry. @YouTwitFace <3

Fixup for #999

## Type of change

# Checklist:


EdJoPaTo:
> Oh, yes, thank you. Not sure why it didn't warn me about that.

Guess you only get warnings when you do something like this:
```ts
const context: Context = {}
```

and even that is kind of horrible as there are ~50 Methods...

so basically we are back at human reviews

PR #1017, opened 2020-04-30
Added return of an object describing the connection result.

an-ivannikov:
# Description

Added return of an object describing the connection result.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Node.js Version:
* Operating System:

# Checklist:

### DIFF
--- a/telegraf.js
+++ b/telegraf.js
@@ -108,7 +108,10 @@ class Telegraf extends Composer {
           const { timeout, limit, allowedUpdates, stopCallback } = config.polling || {}
           return this.telegram.deleteWebhook()
             .then(() => this.startPolling(timeout, limit, allowedUpdates, stopCallback))
-            .then(() => debug('Bot started with long-polling'))
+            .then(() => {
+              debug('Bot started with long-polling')
+              return { ok: true, error: null, type: 'long-polling' }
+            })
         }
         if (typeof config.webhook.domain !== 'string' && typeof config.webhook.hookPath !== 'string') {
           throw new Error('Webhook domain or webhook path is required')
@@ -126,7 +129,15 @@ class Telegraf extends Composer {
         }
         return this.telegram
           .setWebhook(`https://${domain}${hookPath}`)
-          .then(() => debug(`Bot started with webhook @ https://${domain}`))
+          .then(() => {
+            debug(`Bot started with webhook @ https://${domain}`)
+            return { ok: true, error: null, type: 'webhook' }
+          })
+      })
+      .catch((err) => {
+        console.error('Launch failed')
+        console.error(err.stack || err.toString())
+        return { ok: false, error: { message: 'Launch failed', code: err.code ? err.code : null } }
       })
   }
 
--- a/telegraf.js
+++ b/telegraf.js
@@ -137,7 +137,7 @@ class Telegraf extends Composer {
       .catch((err) => {
         console.error('Launch failed')
         console.error(err.stack || err.toString())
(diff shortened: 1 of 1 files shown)


an-ivannikov:
Честно я не понимаю зачем вы это сделали:
https://github.com/telegraf/telegraf/pull/1004

Когда я предлагал и говорил совсем о другом: о том чтобы возвращать ответ с описанием. сейчас ничего не возвращается. и вы убрали перехват ошибки. Не понимаю.

an-ivannikov:
это например 
```js
      .catch((err) => {
        console.error('Launch failed')
        console.error(err.stack || err.toString())
        return { ok: false, error: { message: 'Launch failed', code: err.code ? err.code : null } }
      })
```
можно заменить на выражение с 'throw'
```js
      .catch((err) => {
        console.error('Launch failed')
        console.error(err.stack || err.toString())
        throw new Error({ ok: false, error: { message: 'Launch failed', code: err.code ? err.code : null } })
      })
```

wojpawlik:
What's the pragmatic advantage over current code?

an-ivannikov:
This code exposes the object describing the result of the connections. You can specify an error code in it, they are different for the state when there is no connection and when the server rejects the connection.

Nothing is returned in the current code. if successful, returns 'undefined'.

to your code in case of an error also returned 'undefined'.

After your code, the program will simply crash if an error is not caught.

an-ivannikov:
Your code was created in a simple way - just delete what bothers you. On the contrary, I expanded the functionality of the code.

wojpawlik:
Crashing is the correct thing to do if error is not handled, isn't it?

> to your code in case of an error also returned 'undefined'.

Not true, it idiomatically returns rejected promise with an error.

an-ivannikov:
`to your code ...` -> before your code ...

wojpawlik:
I fail to see how any `.catch` can be beneficial here

an-ivannikov:
Any `.catch` can be useful here for working out errors at the framework level. In my code, there is also a `throw new Error` which will nullify the presence of `.catch`.

wojpawlik:
How is "Launch failed" any more helpful than, well, any error returned from `.launch`?

an-ivannikov:
The framework handles the error and throws a new exception. Processing at the framework level can be any.

It is important for me to get an error code so that I can understand the reason: the network is not working or is it a connection problem associated with blocking telegram servers in Russia.

At the moment, the framework does not return anything if successful. The fremvor returned nothing to your code in case of an error. The framework intercepted the event, wrote to the console about the presence of an error, but did not return the error and the object.

My code suggestion makes the framework handle success and error but always returns a correct description.

wojpawlik:
I wouldn't oppose returning something useful on success.
However, I don't think that any error processing is currently necessary.

> It is important for me to get an error code so that I can understand the reason

`.catch(({ code }) => {...})` should work currently.

an-ivannikov:
I wouldn't oppose returning something useful on success.
However, I don't think that any error processing is currently necessary.

I do not agree with you. I believe that starting a bot should return the result anyway.

This code is currently working. I know this without you. Prior to your code changes (removing interception) this could not work.

In any case, I do not care because I use a post-installed script to replace your file with my file with my changes.

If you think that it will be better for your project not to return the result of the action, then this is your problem. You can break the whole project into electronic functions and each user will have to create his own framework.

I will just make my fork and will provide a convenient framework that works out of the box and in which each action returns a result.

Success in the development of the project. I'm tired of this useless conversation. You can continue to remove from the project those problem areas that you cannot solve.

PR #1056, opened 2020-06-04
typings: improve HearsTriggers

h0tw4t3r:
Let users handle triggers with boolean callbacks.

Use-case: You don't rely on RegEx but on the scoped data or some variable.
https://github.com/telegraf/telegraf/blob/39e89c4d27692dbab7880718831840254980f0ef/composer.js#L390 Allows this

# Description
Let users handle triggers with boolean callbacks.

## Type of change

# Checklist:

BTW, is it possible to trigger on async callback? That would allow custom triggers that rely on i.e. Database values.

### DIFF
--- a/typings/composer.d.ts
+++ b/typings/composer.d.ts
@@ -3,7 +3,7 @@
 import * as tt from './telegram-types.d'
 import { TelegrafContext } from './context'
 
-type HearsTriggers<TContext> = string[] | string | RegExp | RegExp[] | ((value: string, ctx: TContext) => RegExpExecArray | null)
+type HearsTriggers<TContext> = string[] | string | RegExp | RegExp[] | ((value: string, ctx?: TContext) => boolean)
 type BranchPredicate<TContext> = boolean | ((ctx: TContext) => boolean | Promise<boolean>)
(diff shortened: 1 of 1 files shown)


EdJoPaTo:
Take a look on these two code snippets:
https://github.com/telegraf/telegraf/blob/d2a29d292d915df5b175dde584de4a1e1b27a5db/composer.js#L381-L400
https://github.com/telegraf/telegraf/blob/d2a29d292d915df5b175dde584de4a1e1b27a5db/composer.js#L258-L261

This basically means the trigger function has to return what ends up in `ctx.match`.
Yes, due to the truthy check of `if (ctx.match)` you can return a `RegExpExecArray` or `true` and both are truthy but the user expects a `RegExpExecArray` as `ctx.match`. Thats why i think this change is not correct to the actual behaviour of the JavaScript code.

To the Async Trigger Question:
As Composer.optional can support async function the triggers can be async too when the trigger function here is awaited:
https://github.com/telegraf/telegraf/blob/d2a29d292d915df5b175dde584de4a1e1b27a5db/composer.js#L258

PR #1060, opened 2020-06-07
Add MarkdownV2 to Extras

Ashesh3:
# Description

This adds MarkdownV2 to telegraf/extras

## Type of change
Currently only `Extra.markdown()` is possible, there's no direct support for MarkdownV2 in Extras

# How Has This Been Tested?

Tried sending a message with `MarkdownV2` parse_mode and can confirm i can use all basic markdown plus nested markdown too (only a feature of v2)
image

**Test Configuration**:
* Node.js Version: v14.1.0
* Operating System: Windows 10 x64

# Checklist:

### DIFF
--- a/extra.js
+++ b/extra.js
@@ -41,6 +41,11 @@ class Extra {
     this.parse_mode = value ? 'Markdown' : undefined
     return this
   }
+  
+  markdownv2 (value = true) {
+    this.parse_mode = value ? 'MarkdownV2' : undefined
+    return this
+  }
 
   caption (caption = '') {
     this.caption = caption
@@ -73,6 +78,10 @@ class Extra {
 
   static markdown (value) {
     return new Extra().markdown(value)
+  }
+
+  static markdownv2(value) {
+    return new Extra().markdownv2(value)
   }
 
   static caption (caption) {


Loskir:
Typings would be nice to have

Loskir:
Also I don't really like `markdownv2` name, but can't come up with better ideas (`MarkdownV2` maybe, but I don't like it either)

wojpawlik:
`markdownV2`, for consistency? https://github.com/telegraf/telegraf/blob/cce84b355d80363c3620f3283d7f66558d08b201/context.js#L561

Ashesh3:
> `markdownV2`, for consistency?
> 
> https://github.com/telegraf/telegraf/blob/cce84b355d80363c3620f3283d7f66558d08b201/context.js#L561

Hey, it is already 'markdownV2' as you can see [here](https://github.com/telegraf/telegraf/pull/1060/commits/878b95217d61183fdfd094b22660ef1e7e755537)
What is pending to get it out of the draft? :)

wojpawlik:
I mean, "v" should be upper case. I don't see this PR getting out of draft state, I think it should be redone after #1076.

KnorpelSenf:
@Ashesh3 that is a good idea, check out https://github.com/telegraf/telegraf/blob/develop/src/extra.ts and reapply your changes there. The newly added dependency to [typegram](https://www.npmjs.com/package/typegram) already makes sure that `'MarkdownV2'` is a permitted value for `parse_mode` and it would make sense if `telegraf` had support for that.

PR #1064, opened 2020-06-12
Composer inline query & Context state & ExtraAnimation types

ejnshtein:
Add typing for handling inline queries in composer and state property in telegraf context.


typings/context.d.ts
```diff
@@ -4,7 +4,7 @@ import * as tt from './telegram-types'
 
 import { Telegram } from './telegram'
 
-export declare class TelegrafContext {
+export declare class TelegrafContext<StateType = any> {
```
wojpawlik, line 7: 👎, this is just implicit casting, `TelegrafContext<T>` is assignable to `TelegrafContext<any>` for any `T`. Do `state: any` for now.

wojpawlik:
```suggestion
export declare class TelegrafContext {
```

typings/context.d.ts
```diff
@@ -26,6 +26,7 @@ export declare class TelegrafContext {
   pollAnswer?: tt.PollAnswer
   preCheckoutQuery?: tt.PreCheckoutQuery
   shippingQuery?: tt.ShippingQuery
+  state?: StateType
```
wojpawlik, line 29:
```suggestion
  state: any
```

PR #1082, opened 2020-07-11
Rework and re-enable `webhookReply`

wojpawlik:
# Description

Typings of `sendMessage` indicate that it always returns sent `Message` on success, and our code relies on that to be able to delete it later (https://github.com/thedevs-network/the-guard-bot/issues/118#issuecomment-623101479).

After this change, only methods that always return `true` on success, providing no useful information, are eligible for `webhookReply`.

# How Has This Been Tested?

- `npm test` on Node.js v14.0.0

# Checklist:


test/telegraf.js
```diff
@@ -285,8 +284,8 @@ test.cb('should work with context extensions', (t) => {
 
 test.cb('should handle webhook response', (t) => {
   const bot = new Telegraf()
-  bot.on('message', async ({ reply }) => {
-    const result = await reply(':)')
+  bot.on('message', async (ctx) => {
+    const result = await ctx.replyWithChatAction('typing')
     t.deepEqual(result, { webhook: true })
```
wojpawlik, line 289: This test seems to never fail!

wojpawlik: ...because of `await` 😕

KnorpelSenf:
Brief summary of how this could be solved:
- remove all current webhook reply configuration
- always use webhook reply as the first API request of the middleware, and use regular requests for follow-up API  calls, BUT
- only do this for methods where it makes sense (no meaningful return data or swallowed errors)
- let the user adjust what these methods are

wojpawlik:
I'll leave those to a future pull request I think

> * remove all current webhook reply configuration
> * let the user adjust what these methods are

KnorpelSenf (approved):
Looks about right :+1:

Issue #1074, opened 2020-06-26
Moving to Typescript


wojpawlik:
### Non-breaking

- [x] Merge/close all pull requests.
- [x] Move `.js` files to `src/`, `tsc`, add `{ "exports": { "./": "./lib/" } }` to `package.json`.
- [x] ~~Add stub files to avoid breaking for Node <12~~ Drop Node <12 compat
- [x] Change extension to `.ts` and add type annotations from `.d.ts` files.
  + [x] `router.ts`
  + [x] `telegram.ts`
  + [x] `context.ts`
  + [x] `composer.ts`
  + [x] `core/network/client.ts`
  + [x] `telegraf.ts`
- [x] Configure formatting/linting of `.ts` files.
- [ ] Move the main documentation to `typedoc` (#1205)
- [ ] Update examples (#1177, #1194)

### Help wanted

- [x] `scenes/`
- [x] `stage.ts`
- [x] `extra.ts` (#1076)
- [x] `session.ts` (#914)
- [x] Complete type annotations, `noImplicitAny`.
- [ ] Fix eslint warnings and https://github.com/telegraf/telegraf/search?q=ts-expect-error.

### Breaking

- [x] `index.ts` ~~(awaiting https://github.com/microsoft/TypeScript/issues/39464)~~
- [x] `mjs-entry`
- [x] `"exports": { "import": "./lib/index.mjs", "require": "./lib/index.js" }`
- [x] `markup.ts` (rewrite)
- [x] Start building `.d.ts` files from source.
- [ ] Resolve [related issues](https://github.com/telegraf/telegraf/labels/typings).
- [ ] (optional) [denoify](https://github.com/garronej/denoify/issues/13#issuecomment-689076929).

wojpawlik:
`export=` makes it difficult if not impossible to export related interfaces, so it has to go (breaking).

wojpawlik:
I'm done for now. Would appreciate help with the rest of the files, or green light for breaking changes.

KnorpelSenf:
Are you planning to do a full migration to TS, that is, eventually dropping `allowJs`? If so, I have a few questions regarding this issue, sorted by importance, highest first.

[...]

PR #1084, opened 2020-07-13
Feature/ts improvements

trgwii:
# Description

Adds improved typings to src/telegraf.ts

# How Has This Been Tested?

All tests pass

**Test Configuration**:
* Node.js Version: v14.4.0

# Checklist:

### DIFF
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -1,25 +1,26 @@
-/* eslint @typescript-eslint/no-var-requires: "warn" */
-import * as http from 'http'
-import * as https from 'https'
 import type * as tt from '../typings/telegram-types'
 import type ApiClient from './core/network/client'
+import BaseScene from './scenes/base'
 import Composer from './composer'
 import Context from './context'
 import crypto from 'crypto'
+import Debug from 'debug'
+import Extra from './extra'
 import generateCallback from './core/network/webhook'
+import http from 'http'
+import https from 'https'
+import Markup from './markup'
 import { promisify } from 'util'
+import Router from './router'
+import session from './session'
+import Stage from './stage'
 import Telegram from './telegram'
 import type { TlsOptions } from 'tls'
 import { URL } from 'url'
-const debug = require('debug')('telegraf:core')
-const Extra = require('./extra')
-const Markup = require('./markup')
-const session = require('./session')
-const Router = require('./router')
-const Stage = require('./stage')
-const BaseScene = require('./scenes/base')
 
-const DEFAULT_OPTIONS: Telegraf.Options<Context> = {
+const debug = Debug('telegraf:core')
+
+const DEFAULT_OPTIONS: Telegraf.Options<any> = {
   telegram: {},
   retryAfter: 1,
   handlerTimeout: 0,
@@ -27,7 +28,7 @@ const DEFAULT_OPTIONS: Telegraf.Options<Context> = {
   contextType: Context,
 }
 
-const noop = () => {}
+const noop: () => void = () => undefined
 function always<T>(x: T) {
   return () => x
 }
@@ -96,14 +97,15 @@ class Telegraf<TContext extends Context = Context> extends Composer<TContext> {
 
   handleError: (err: any, ctx: TContext) => void = (err) => {
     console.error()
-    console.error((err.stack || err.toString()).replace(/^/gm, '  '))
+    console.error(
+      ('stack' in err ? err.stack : err.toString()).replace(/^/gm, '  ')
+    )
     console.error()
     throw err
   }
 
   constructor(token: string, options?: Partial<Telegraf.Options<TContext>>) {
     super()
-    // @ts-expect-error
     this.options = {
       ...DEFAULT_OPTIONS,
       ...options,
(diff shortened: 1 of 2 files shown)


src/telegraf.ts
```diff
@@ -1,33 +1,34 @@
...
 import type ApiClient from './core/network/client'
+import BaseScene from './scenes/base'
 import Composer from './composer'
 import Context from './context'
 import crypto from 'crypto'
+import Debug from 'debug'
+import Extra from './extra'
 import generateCallback from './core/network/webhook'
+import http from 'http'
+import https from 'https'
```
wojpawlik, line 11: What's the benefit of doing this? The drawback is obvious -- it'll need to be changed back if `esModuleInterop` is disabled.

src/telegraf.ts
```diff
@@ -1,33 +1,34 @@
...
+const DEFAULT_OPTIONS: Telegraf.Options<any> = {
   telegram: {},
   retryAfter: 1,
   handlerTimeout: 0,
   channelMode: false,
   contextType: Context,
 }
 
-const noop = () => {}
+const noop: () => void = () => undefined
```
wojpawlik, line 31: This is just more verbose way of doing the same thing, isn't it?

src/telegraf.ts
```diff
@@ -168,12 +173,13 @@ class Telegraf<TContext extends Context = Context> extends Composer<TContext> {
   ) {
     const webhookCb = this.webhookCallback(hookPath)
     const callback: http.RequestListener =
-      cb && typeof cb === 'function'
+      cb != null && typeof cb === 'function'
```
wojpawlik, line 176:
```suggestion
      typeof cb === 'function'
```

src/telegraf.ts
```diff
@@ -183,20 +189,25 @@ class Telegraf<TContext extends Context = Context> extends Composer<TContext> {
   launch(config: Telegraf.LaunchOptions = {}) {
     debug('Connecting to Telegram')
     return this.telegram.getMe().then((botInfo) => {
-      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-      debug(`Launching @${botInfo.username}`)
+      debug(
+        `Launching ${botInfo.username == null ? 'bot' : '@' + botInfo.username}`
```
wojpawlik, line 193: Dead code, bots always have a username as far as I'm aware. The real solution here is to tighten the type returned by `getMe` and stored in `botInfo` to something like
```ts
interface BotUser {
  id: number
  is_bot: true
  first_name: string
  username: string
  can_join_groups: boolean
  can_read_all_group_messages: boolean
  supports_inline_queries: boolean
}
```

src/telegraf.ts
```diff
@@ -96,14 +97,15 @@ class Telegraf<TContext extends Context = Context> extends Composer<TContext> {
 
   handleError: (err: any, ctx: TContext) => void = (err) => {
     console.error()
-    console.error((err.stack || err.toString()).replace(/^/gm, '  '))
+    console.error(
+      ('stack' in err ? err.stack : err.toString()).replace(/^/gm, '  ')
+    )
```
wojpawlik, line 102: Nothing wrong with this, but it also doesn't seem to fix anything.

src/telegraf.ts
```diff
@@ -1,33 +1,34 @@
...
-const Markup = require('./markup')
-const session = require('./session')
-const Router = require('./router')
-const Stage = require('./stage')
-const BaseScene = require('./scenes/base')
 
-const DEFAULT_OPTIONS: Telegraf.Options<Context> = {
+const debug = Debug('telegraf:core')
+
+const DEFAULT_OPTIONS: Telegraf.Options<any> = {
```
wojpawlik, line 23: 👍

src/telegraf.ts
trgwii: I prefer to make it as tidy as possible for now, esModuleInterop changes later are easy to fix

src/telegraf.ts
trgwii: No, this fixes a type error further down, where another function expects a void callback, and not an undefined callback

src/telegraf.ts
trgwii: Sure, we'll fix that later, I'd enjoy seeing some extending here, since BotUser would just be an extension of User in this case

src/telegraf.ts
trgwii: It checks that err.stack is a property on err, instead of checking if it's nullish

src/telegraf.ts
wojpawlik: How? `() => {}` is already a void callback.

trgwii: Weird, NOW it is for me too, previously this wasn't a void callback but the type turned into () => undefined.. uhh

KnorpelSenf:
Any ETA for this? What is the current state?

I'm waiting for this because it is blocking #1074.

wojpawlik:
I don't think it's blocking, as long as you don't touch `src/telegraf.ts`. Still, feel free to make any changes, I'll close this if serious conflicts arise, as @trgwii seems to be ignoring some of my feedback.

trgwii:
@wojpawlik would it be possible for me to give you write access to my fork? I'm not very interested in arguing about these tiny things I really don't care about, but I'm fìne with letting you have final say on all of these as long as my code gets merged. I enjoy making larger contributions than these changes

KnorpelSenf:
Save your time you two. I'll just pull your work into my fork and then continue the migration journey from there. I'll most likely need to touch most lines of the codebase anyway. After all, half of it is still JS. I applied a reasonably tight TS config for a project of this size and I'm at nearly 300 compilation errors now. It's not just gonna take a week before this is done. I'll submit a separate PR once I have a first proper step to share.

KnorpelSenf:
@trgwii please check out if something slipped through on my side or if you're missing something else in my PR, I'll happily add it

Issue #1076, opened 2020-06-30
`Extra::caption` ovewrties itself


wojpawlik:
https://github.com/telegraf/telegraf/blob/359f4f106dddccabfa6523c3bbd6ddf79a8b826a/extra.js#L45-L46

This cannot be ported to idiomatic Typescript.

KnorpelSenf:
That's because the type of the class member `caption` changes at runtime.  You _can_ type it correctly by this:
```ts
  caption: string | ((caption: string) => Extra) = (caption = '') => {
    this.caption = caption
    return this
  }
```
This would properly reflect that it is never really known for a given instance of `Extra` what type its property `caption` has.

However, this would of course break existing code like `new Extra().caption('asdf')` which is undesired. If we want to capture that `caption` changes its type exactly once (when called), we need to take into account that this also changes the type of `Extra` as it is in a sense the sum of its parts. Consequently, we need to define a second version of `Extra`, namely one where `caption` was already called and thus has type `string`.
```ts
interface CaptionedExtra extends Omit<Extra, 'caption'> {
  caption: string
}
```
We can now use `Extra` as before and tell the compiler that we know what we're doing when changing the type of `Extra` at runtime:
```ts
  caption(caption: string): CaptionedExtra {
    const me = this as unknown as CaptionedExtra
    me.caption = caption
    return me
  }
```
This catches all typing errors as desired.

One can argue that this is still not entirely idiomatic because we preferably want to avoid explicit type casting. I would counter this with saying that the JS version itself is not idiomatic either, but fortunately TS has enough flexibility to be bent around such that we can have type safety even in this somewhat odd case.

Note that the declaration of an interface could also be skipped, even though that does not really contribute to readability and also prevents us from adding a full function signature (without repeating code):
```ts
  caption(caption: string) {
    const me = this as unknown as Omit<Extra, 'caption'> & { caption: string }
    me.caption = caption
    return me
  }
```

KnorpelSenf:
Now _that's_ what I call impossible to type:

```js
// found in class Markup from markup.js:
  keyboard (buttons, options) {
    const keyboard = buildKeyboard(buttons, { columns: 1, ...options })
    if (keyboard && keyboard.length > 0) {
      this.keyboard = keyboard
    }
    return this
  }
```

[...]

PR #1108, opened 2020-08-18
Migrate to TS

KnorpelSenf:
# Description

Tighten TS config, eliminate all JS files, add missing types for most files. Also, the project did not follow its own prettier config, I reformatted the code in these places (except where disabled explicitly by `prettier-ignore` comment of course).

Fixes
- #1074 (not yet)

PRs incorporated
- #1084

Currently, there is still a few things that need to be done. It's those things that need to be discussed first.

1) `src/core/replicators.ts` has a lot of errors because I don't really know what type `message` has in all these methods. Who would invoke them anyway?
1) `src/markup.ts` also has a few errors. I fixed https://github.com/telegraf/telegraf/issues/1076#issuecomment-657288263, but what about https://github.com/telegraf/telegraf/issues/1076#issuecomment-675119452? This calls for breaking changes. Some other dependant methods have errors as a side-effect, they need to be fixed, too.
1) `src/scenes/context.ts` has a few lines where `this.current` is accessed—I really don't know how that is supposed to behave.
1) Also confer https://github.com/telegraf/telegraf/issues/1074#issuecomment-657558679 for a few more things that are planned. They could be done here, as well.

Please give some advice on how these problems are supposed to be solved.

## Type of change

As of today, this PR changes only typing. However, it would make much sense to introduce breaking changes, too (see above).

# How Has This Been Tested?

- [X] `npm run typecheck`

Otherwise, thorough testing is still an open task for this PR. It needs to be verified that no breaking changes other than the intended ones are introduced.

# Checklist:

- [X] I have made corresponding changes to the documentation (none)
- [ ] ~My changes generate no new warnings~ (they do, that's why we're in draft state)
- [ ] ~I have added tests that prove my fix is effective or that my feature works~ (does not apply)
- [ ] ~New and existing unit tests pass locally with my changes~ (not until all above problems are solved)


wojpawlik (changes requested):
Formatting-only changes are not welcome.

Do you have any reason against using `import type`?

KnorpelSenf:
I know that formatting-only changes are not welcome. We avoid them using Prettier. However, if someone commits code to the repo that does not follow the formatting conventions, I believe it is better to fix the formatting once and do it properly rather than permitting that people keep on submitting formattings that follow their own rules. That way, we can keep everyone in sync.

I don't personally have something against import type declarations, I understand their purpose. I suggest following the recommendations from the TS team. I quote:
> This feature is something most users may never have to think about; however, if you’ve hit issues under --isolatedModules, TypeScript’s transpileModule API, or Babel, this feature might be relevant.

What they're basically saying is that this should only be used when necessary. `telegraf` applied them inconsistently and without having a reason to do so. I'm completely fine with using them everywhere if that is desired for some reason, however, as they were used only sparingly, I decided that the consistent state of not using them at all is better than the consistent state of using them wherever possible.

src/composer.ts
```diff
@@ -215,9 +215,11 @@ class Composer<TContext extends Context> implements Middleware.Obj<TContext> {
     return (ctx, next) => next()
   }
 
-  private static safePassThru() {
-    // prettier-ignore
-    return (ctx, next) => typeof next === 'function' ? next(ctx) : Promise.resolve()
+  static safePassThru(): Middleware.Fn<Context> {
+    return (ctx, next) =>
+      typeof next === 'function'
+        ? (next as (ctx: Context) => void)(ctx)
```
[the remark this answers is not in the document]
KnorpelSenf: I concur with you wholeheartedly. However, it is imported at another place in a library, thus marking it `private` causes compilation errors.

Should we eliminate either of the two versions of the pass-through function?

src/context.ts
```diff
@@ -65,6 +66,8 @@ class TelegrafContext {
   readonly updateSubTypes: ReadonlyArray<typeof MessageSubTypes[number]>
   /** @deprecated */
   private contextState?: any
+  scene: any
+  wizard!: WizardContext<any>
```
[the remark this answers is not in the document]
KnorpelSenf: Good question! I don't think so.

First of all, any usage of `any` can and should be improved because `any` is always just an intermediate migration type.

Second, `wizard` is regularly assumed to be a member that is truthy. Most often, no checks are performed. If it is possible that these values are nullish, this would unveil potential bugs that need fixing. I mainly focused on `typing the library in the ways it works as of today` and not `typing the library in the way it makes sense`. Sometimes that coincides, sometimes it does …

Do you have a suggestion here?

src/session.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+    Object.defineProperty(ctx, options.property, {
+      get: function () {
+        return session
+      },
+      set: function (newValue) {
+        session = { ...newValue }
+      },
+    })
+    await (next as (ctx: TContext) => Promise<void>)(ctx)
```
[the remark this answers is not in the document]
KnorpelSenf: Is it correct to remove the argument from the function call? If so, I'd commit your suggestion.

src/core/network/client.ts
```diff
@@ -60,23 +62,27 @@ const WEBHOOK_REPLY_STUB = {
 }
 
 // prettier-ignore
```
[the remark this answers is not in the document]
KnorpelSenf: Should I just go ahead remove it wherever possible? Maybe in a separate commit or even a separate PR so we have an isolated change?

src/core/network/client.ts
```diff
@@ -60,23 +62,27 @@ const WEBHOOK_REPLY_STUB = {
...
-function includesMedia (payload) {
+function includesMedia (payload: object) {
   return Object.keys(payload).some(
     (key) => {
-      const value = payload[key]
+      const value = (payload as { [k: string]: unknown })[key]
```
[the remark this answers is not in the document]
KnorpelSenf: Good idea, thanks

src/core/network/client.ts
```diff
@@ -267,12 +298,18 @@ class ApiClient {
...
     const payload = Object.keys(data)
       .filter((key) => typeof data[key] !== 'undefined' && data[key] !== null)
-      .reduce((acc, key) => ({ ...acc, [key]: data[key] }), {})
+      .reduce(
+        (acc, key) => ({
+          ...acc,
+          [key]: data[key],
+        }),
+        {}
+      )
```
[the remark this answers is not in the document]
KnorpelSenf: See my new commit—is that what you're aiming at?

KnorpelSenf:
Is there anything you can say about the three major points I listed in the description? I'd be happy to see that solved

src/composer.ts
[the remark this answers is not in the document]
wojpawlik: I see. Try using `passThru` instead, or at least mark `safePassThru` as deprecated, I want it gone in 4.0.

src/session.ts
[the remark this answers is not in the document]
wojpawlik: Yes. The behavior is identical, just way easier to type: https://github.com/telegraf/telegraf/blob/f1403a8f400086cee3b16ff3574efefef2a87888/src/types.ts#L8-L11

src/core/network/client.ts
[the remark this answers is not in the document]
wojpawlik: No. Only remove them if you're touching adjacent code, to preserve `git blame`. I'd have tried [`precise-commits`](https://github.com/nrwl/precise-commits) instead if it was still maintained.

src/core/network/client.ts
```diff
@@ -301,15 +296,11 @@ class ApiClient {
...
-        (acc, key) => ({
-          ...acc,
-          [key]: data[key],
-        }),
-        {}
-      )
+    const payload = Object.fromEntries(
+      Object.keys(data)
+        .filter((key) => typeof data[key] !== 'undefined' && data[key] !== null)
+        .map((key) => [key, data[key]])
```
wojpawlik, line 325: Not bad, but can be improved:
```suggestion
      Object.entries(data)
        .filter(([key, value]) => value != null)
```

src/composer.ts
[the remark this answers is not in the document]
KnorpelSenf: I completely removed `safePassThru` and replaced it by `passThru`.

src/util.ts
```diff
@@ -0,0 +1,22 @@
...
+  Z extends
+    | 'string'
+    | 'number'
+    | 'bigint'
+    | 'boolean'
+    | 'symbol'
+    | 'undefined'
+    | 'object'
+    | 'function'
+>(obj: X, prop: Y, type: Z): obj is X & Record<Y, Z> {
```
wojpawlik, line 19: Isn't `Z` a literal string instead of a type?

wojpawlik (changes requested):
Sorry for being terse, there's a lot to cover!

By "formatting-only changes" meant the extra newlines in payload objects in `telegraf.ts`, please revert those.

> 1\. `src/core/replicators.ts` has a lot of errors because I don't really know what type `message` has in all these methods. Who would invoke them anyway?

Replicators are used by `Telegram::sendCopy`. You can try using `Required<Pick<Message, ...>>` as param types.

> 3\. `src/scenes/context.ts` has a few lines where `this.current` is accessed—I really don't know how that is supposed to behave.

What's the problem? https://github.com/telegraf/telegraf/blob/66b795a38ab7c0bb8bbc29b05ddcc1d955d26205/src/scenes/context.js#L33-L36

> 4\. Also confer [#1074 (comment)](https://github.com/telegraf/telegraf/issues/1074#issuecomment-657558679) for a few more things that are planned. They could be done here, as well.

Please don't expand the scope of this PR, it's huge already! I'm considering requesting splitting it up.

src/composer.ts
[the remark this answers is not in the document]
wojpawlik: That might have been too aggressive for a non-breaking release. It's also incorrect, some of these should use `anoop` instead of `safePassThru`.

src/core/network/client.ts
KnorpelSenf: Was wondering why you wanted `fromEntries` to appear there.

src/core/network/error.ts
```diff
@@ -0,0 +1,21 @@
...
+  error_code: number
+  description: string
+  parameters?: {}
+}
+class TelegramError extends Error {
+  code: number
+  response: ErrorPayload
+  description: string
+  parameters: {}
+  constructor(readonly payload: ErrorPayload, readonly on?: unknown) {
```
wojpawlik, line 11: My mistake.
```suggestion
  constructor(payload: ErrorPayload, readonly on = {}) {
```

src/util.ts
KnorpelSenf: It's a literal string representing a type.

wojpawlik:
```ts
Record<'foo', 'string'> !== Record<'foo', string>
```

src/composer.ts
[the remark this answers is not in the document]
KnorpelSenf: In that case I need you to elaborate. When was a `safePassThru` used even though a `anoop` was meant to be used?

src/util.ts
KnorpelSenf: Should I provide documentation how this util function works?

KnorpelSenf:
1) I removed the accidental reformattings.
1) That does not work because some properties like `reply_markup` are not even declared on the message type in the first place. I assume this will work anyway at runtime, but I could not find an existing union type or something that could serve as the base for these methods.
1) The problem is `Object is possibly 'null' or 'undefined'.` and there are no checks performed. Is that a bug that should be fixed? Should I change some types?
1) I'm not really considering to extend the scope actually, this whole PR is solely about migrating to TS, just as is #1074.

src/composer.ts
```diff
@@ -198,7 +198,7 @@ class Composer<TContext extends Context> implements Middleware.Obj<TContext> {
     const handler = Composer.unwrap(middleware)
     return (ctx, next) => {
       // eslint-disable-next-line @typescript-eslint/no-misused-promises
-      setImmediate(handler, ctx, Composer.safePassThru())
+      setImmediate(handler, ctx, Composer.passThru())
       return next()
```
wojpawlik, line 202:
```suggestion
return Promise.all([handler(ctx, anoop), next()])
```

KnorpelSenf: I see where you're going with this, but you cannot return an array here. Should the left or the right value be returned?
```js
return Promise.all([handler(ctx, anoop), next()]).then(([l, r]) => l)
```
OR
```js
return Promise.all([handler(ctx, anoop), next()]).then(([l, r]) => r)
```

KnorpelSenf: Oh wait, it's the right one of course

wojpawlik: Users are not supposed to rely on values returned by middleware.

src/util.ts
wojpawlik: Nope, it doesn't do what you want it to do, documenting it doesn't fix that.

src/composer.ts
KnorpelSenf: Exactly :shrug:

wojpawlik:
1. Thanks.
2. Oh, right.
3. ¯\_(ツ)_/¯
4. That's too big scope for a single PR.

src/util.ts
KnorpelSenf: Well actually it does that very well, but perhaps it's not what you were expecting it to do? I added docs now to resolve the misunderstanding

KnorpelSenf: Okay, I think I see what you mean now. The function behaves exactly the way I want, though. I'm currently just not using it in places where the return type needs to be more accurate than `boolean`, but if someone was actually trying to utilize it in a place where flow analysis matters, it would yield the wrong type. I'll quickly fix that, thanks

src/scenes/context.ts
```diff
@@ -0,0 +1,97 @@
...
+  reenter() {
+    return this.enter(this.session.current, this.state)
+  }
+
+  async leave() {
+    debug('Leave scene')
+    const handler =
+      this.current && this.current.leaveMiddleware
+        ? this.current.leaveMiddleware()
+        : passThru()
```
wojpawlik, line 91:
```suggestion
        : Composer.passThru()
```

src/scenes/context.ts
```diff
@@ -0,0 +1,97 @@
...
+  ) {
+    this.ctx = ctx
+    this.scenes = scenes
+    this.options = options
+  }
+
+  get session() {
+    const sessionName = this.options.sessionName
+    let session =
+      (this.ctx as { [sessionName: string]: any })[sessionName].__scenes || {}
```
wojpawlik, line 30:
```suggestion
    let session = (this.ctx as any)[sessionName].__scenes ?? {}
```

src/scenes/context.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (session.expires < now()) {
+      session = {}
+    }
+    ;(this.ctx as { [sessionName: string]: any })[
+      sessionName
+    ].__scenes = session
```
wojpawlik, line 36:
```suggestion
    (this.ctx as any)[sessionName].__scenes = session
```

src/scenes/context.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  get current() {
+    const sceneId = this.session.current || this.options.default
+    return sceneId && this.scenes.has(sceneId) ? this.scenes.get(sceneId) : null
+  }
+
+  reset() {
+    const sessionName = this.options.sessionName
+    delete (this.ctx as { [sessionName: string]: any })[sessionName].__scenes
```
wojpawlik, line 56:
```suggestion
    delete (this.ctx as any)[sessionName].__scenes
```

src/scenes/context.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  enter(sceneId: string, initialState: any, silent?: boolean) {
+    if (!sceneId || !this.scenes.has(sceneId)) {
```
wojpawlik, line 60:
```suggestion
    if (!this.scenes.has(sceneId)) {
```

src/scenes/wizard/context.ts
```diff
@@ -0,0 +1,35 @@
+import TelegrafContext from '../../context'
+import { Middleware } from '../../types'
+
+class WizardContext<TContext extends TelegrafContext> {
+  ctx: TContext
+  steps: Array<Middleware<TContext>>
```
wojpawlik, line 6:
```suggestion
  steps: ReadonlyArray<Middleware<TContext>>
```

src/core/network/client.ts
```diff
@@ -221,13 +266,15 @@ function answerToWebhook(response, payload, options: ApiClient.Options) {
...
+          response.set(key, (headers as { [k: string]: unknown })[key])
+        )
         response.body = body
         return Promise.resolve(WEBHOOK_REPLY_STUB)
       }
       if (!response.headersSent) {
         Object.keys(headers).forEach((key) =>
-          response.setHeader(key, headers[key])
+          response.setHeader(key, (headers as { [k: string]: unknown })[key])
         )
```
wojpawlik, line 278:
```suggestion
        for (const [key, value] of Object.entries(headers)) {
          response.set(key, value)
        }
```

src/scenes/wizard/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    super()
+    this.id = id
+    this.options =
+      typeof options === 'function'
+        ? { steps: [options, ...steps], leaveHandlers: [] }
+        : { steps: steps, leaveHandlers: [], ...options }
+    this.leaveHandler = compose(this.options.leaveHandlers)
+  }
+
+  set ttl(value) {
```
wojpawlik, line 29:
```suggestion
  set ttl(value: number | undefined) {
```

src/scenes/wizard/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  middleware() {
+    return compose([
+      (ctx, next) => {
+        const wizard = new WizardContext(ctx, this.options.steps)
+        ctx.wizard = wizard
+        return next()
+      },
+      super.middleware() as Middleware.Fn<TelegrafContext>,
```
wojpawlik, line 53:
```suggestion
    return Composer.compose<TContext>([
      (ctx, next) => {
        const wizard = new WizardContext(ctx, this.options.steps)
        ctx.wizard = wizard
        return next()
      },
      super.middleware(),
```

src/stage.ts
```diff
@@ -0,0 +1,68 @@
...
+    return this
+  }
+
+  middleware(): Middleware.Fn<TContext> {
+    const handler = compose([
+      (ctx: TContext, next) => {
+        ctx.scene = new SceneContext<TContext>(ctx, this.scenes, this.options)
+        return next()
+      },
+      super.middleware() as Middleware.Fn<TelegrafContext>,
```
wojpawlik, line 40:
```suggestion
      super.middleware(),
```

src/stage.ts
```diff
@@ -0,0 +1,68 @@
...
+      lazy((ctx) => ctx.scene.current || passThru()),
+    ])
+    return optional(
+      (ctx: { [key: string]: any }) => ctx[this.options.sessionName],
+      handler
+    ) as Middleware.Fn<TContext>
```
wojpawlik, line 46:
```suggestion
    return Composer.unwrap(
      Composer.optional(
        (ctx) => (ctx as any)[this.options.sessionName],
        handler
      )
    )
```

`Composer.unwrap` would be unnecessary if this was removed: https://github.com/telegraf/telegraf/blob/2e4a9d4ca45175c4de01e8a38706f4eb0bdd05cc/src/composer.ts#L242-L244

src/telegram.ts
```diff
@@ -187,7 +187,7 @@ class Telegram extends ApiClient {
     longitude: number,
     title: string,
     address: string,
-    extra
+    extra: { [key: string]: any }
```
wojpawlik, line 190:
```suggestion
    extra: tt.ExtraVenue
```

src/telegram.ts
```diff
@@ -218,7 +218,7 @@ class Telegram extends ApiClient {
     chatId: number | string,
     phoneNumber: string,
     firstName: string,
-    extra
+    extra: { [key: string]: any }
```
wojpawlik, line 221:
```suggestion
    extra: tt.ExtraContact
```

src/telegram.ts
```diff
@@ -970,8 +970,9 @@ class Telegram extends ApiClient {
     if (!message) {
       throw new Error('Message is required')
     }
-    // prettier-ignore
-    const type = Object.keys(replicators.copyMethods).find((type) => message[type])
+    const type = Object.keys(replicators.copyMethods).find(
+      (type) => ((message as unknown) as { [key: string]: string })[type]
```
wojpawlik, line 974:
```suggestion
      (type) => (message as any)[type]
```

src/scenes/wizard/index.ts
```diff
@@ -0,0 +1,65 @@
+import Composer from '../../composer'
+import TelegrafContext from '../../context'
+import { Middleware } from '../../types'
+import WizardContext from './context'
+const { compose, unwrap } = Composer
+
+type WizardSceneOptions<TContext extends TelegrafContext> =
+  | Middleware.Fn<TContext>
+  | Array<Middleware.Fn<TContext>>
```
wojpawlik, line 9: Unused.
```suggestion
```

src/markup.ts
```diff
@@ -1,173 +1,212 @@
...
     return {
       text: text,
       login_url: { ...opts, url: url },
-      hide: hide
+      hide: hide,
     }
   }
 
-  static formatHTML (text = '', entities = []) {
+  static formatHTML(text = '', entities = []) {
```
wojpawlik, line 205:
```suggestion
  static formatHTML(text = '', entities: tt.MessageEntity[] = []) {
```

src/markup.ts
```diff
@@ -1,173 +1,212 @@
...
     const chars = text
     const available = [...entities]
-    const opened = []
-    const result = []
+    const opened: any = []
```
wojpawlik, line 208:
```suggestion
    const opened: number[] = []
```

src/context.ts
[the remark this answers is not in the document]
wojpawlik: They're assigned right here: https://github.com/telegraf/telegraf/blob/5dec6927eab19df2efe02394cfa51cc6ce0eb825/src/scenes/wizard/index.ts#L50 https://github.com/telegraf/telegraf/blob/f78c6eff3065ce5348397d64c139c2fe52754089/src/stage.ts#L37

src/telegram.ts
wojpawlik: Or
```suggestion
      (type) => type in message
```

src/context.ts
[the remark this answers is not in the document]
wojpawlik: I believe this (from #1113) would be helpful: https://github.com/telegraf/telegraf/blob/3e2d4adb328457d2841a4c3a2ed51b56cead8caf/src/types.ts#L21-L26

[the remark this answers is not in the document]
wojpawlik: I'm working on this.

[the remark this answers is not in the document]
wojpawlik: I hope they won't turn out contravariant

foodornt:
Can I somehow use these typings in my project?

KnorpelSenf:
@foodornt sure, check out [typegram](https://github.com/KnorpelSenf/typegram)

foodornt:
@KnorpelSenf thank you man.

btw, is there any wizardscene typings you know of?
or can I somehow use the develop branch to make my bot.

KnorpelSenf:
Even though I migrated the module, I'm not an expert on wizards, so I cannot really help you here.

Feel free to open an issue and remember to include an if possible. People won't necessarily read an old PR that's already been merged some time ago.

foodornt:
@KnorpelSenf I think I worked it out by changing some types .
Much appreciated, thank you for helping!

PR #1112, opened 2020-08-19
formatHTML() misses the entities offset when special characters are used #1109

### LINKED ISSUE #1109: Markup.formatHTML() misses the entities offset when special characters are used
## Context

i have a message object which contains an emoji and when i use sendCopy() to forward the message the inlne links starting point gets changed.

### original text
image

### received with sendCopy() 
image

* Telegraf.js Version: 3.38.0
* Node.js Version: 12.18.3
* Operating System: windows 10 Home (version 2004)

## Expected Behavior

the message should be exactly as the first one.

dagema1:
# Description

when using formatHTML the usage of a spread operator to split string causes multi character emojis to mess up the offset of markups.

Fixes #1109 

## Type of change

**Test Configuration**:
* Node.js Version: v12.18.3
* Operating System: Windows 10 home

# Checklist:


wojpawlik:
Add a test showing that your fix is correct and that current code is wrong.

dagema1:
i think  #1105 fixed it.

PR #1115, opened 2020-08-23
Ditch telegram-typings dependency

KnorpelSenf:
# Description

The dependency `telegram-typings` is no longer maintained. It causes a few errors because of missing types that [need to be suppressed](https://github.com/telegraf/telegraf/blob/2d6582a59c15b19a55c8bee6326d28af0bdd10ba/src/context.ts#L136) in turn. This PR gets rid of it. Instead, a more up to date dependency is used. As a result, a lot of code can be simplified as well as updated to API version 4.9.

## Type of change

- [X] Bug fix (non-breaking changes which fix a few typing issues)
- [X] New feature (non-breaking change which adds new types)

# How Has This Been Tested?

```bash
npm run typecheck
```

# Checklist:

### DIFF
--- a/typings/telegram-types.d.ts
+++ b/typings/telegram-types.d.ts
@@ -3,7 +3,6 @@
 import * as TT from 'typegram'
 export * from 'typegram'
 
-
 export type ChatAction = TT.Opts<'sendChatAction'>['action']
 
 export type ChatType = TT.Chat['type']
@@ -51,7 +50,6 @@ export type MessageSubTypes =
   | 'connected_website'
   | 'animation'
 
-
 export type MessageMedia = TT.InputMedia
 
 export interface StickerData {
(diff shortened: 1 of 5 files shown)


KnorpelSenf:
Most likely it is possible to improve the typings of the library much more now, however, I did not consider this to be in scope of this PR. I might have a look at this later.

typings/telegram-types.d.ts
```diff
@@ -62,308 +52,39 @@ export type MessageSubTypes =
...
-  media: InputFile
-  thumb?: string | InputFile
-  caption?: string
-  parse_mode?: string
-  width?: number
-  height?: number
-  duration?: number
-  supports_streaming?: boolean
-}
+export type InlineQueryResult = TT.InlineQueryResult
```
wojpawlik, line 55: No-op due to `export * from 'typegram'` at the top; remove.
```suggestion
```

typings/telegram-types.d.ts
```diff
@@ -1,23 +1,13 @@
 /** @format */
 
-import * as TT from 'telegram-typings'
-export * from 'telegram-typings'
+import * as TT from 'typegram'
+export * from 'typegram'
 
-export type ParseMode = 'Markdown' | 'MarkdownV2' | 'HTML'
+export type ParseMode = TT.ParseMode
```
wojpawlik, line 6: No-op due to `export * from 'typegram'` at the top; remove.
```suggestion
```

wojpawlik (approved):
Great work! How about using this as an opportunity to sort `Extra*` types?

Also, how would you feel about improving accuracy of `typegram` by using unions instead of optional properties?

KnorpelSenf:
> Great work! How about using this as an opportunity to sort `Extra*` types?

You mean you simply want a different ordering of the lines with type declarations?

> Also, how would you feel about improving accuracy of `typegram` by using unions instead of optional properties?

I was considering this when I created the library. More explicitly, one could for example make the update type or the message type a large union type where only exactly possible combinations of fields are specified, depending on the type of update or message, respectively.

Pros:
- Never accidentally assign two fields

Cons:
- Cannot `extend` types anymore.
- Accidental reads on two fields do not occur anyway because the properties are optional, thus making checks necessary.
- Cannot not use discriminated unions without breaking API conformity, this could make it difficult to narrow down types.
- Requires A LOT of manual work because this needs to be decided individually on a per-type basis, no automation can be used for that.
- The Telegram API docs decided to draw the same line between where to use unions (`InlineQueryResult`) and when not to.

That's why I decided not to do this.
However, if you have solutions to the above problems, see more advantages, or want to invest the time, I'll happily review your PR.

wojpawlik:
One more (big) advantage: no non-null assertions (in theory).

> * Cannot `extend` types anymore.

Why would extending be useful?

> * Cannot not use discriminated unions without breaking API conformity, this could make it difficult to narrow down types.

Not sure what you mean. Types can be narrowed with `in` type guard.

<hr>

One more suggestion: some properties are "returned only in `getChat`" or "...`getMe`". How about removing them from the base interfaces and put them only in the return value they're present in?

KnorpelSenf:
> One more suggestion: some properties are "returned only in `getChat`" or "...`getMe`". How about removing them from the base interfaces and put them only in the return value they're present in?

That sounds like a possible thing to do, I'll open an issue as a reminder

wojpawlik:
I already PRed.

PR #1116, opened 2020-08-24
Migrate replicators to TS

KnorpelSenf:
# Description

Migrates replicators.js to replicators.js. Now possible due to `typegram`.

Continues work on #1074 

A few oddities are unveiled that look like bugs in the library to me.

1) `thumb` should always be sent as string (`file_id`), it is not possible to just copy the property from the message (would be `{ file_id: string; /* ... */ }`. This happens four times. This PR suggests to access the `file_id` property instead.
2) `Property 'supports_streaming' does not exist on type 'Video'.` Consequently, it cannot be specified when sending a copy of a video. That also makes a lot of sense, because how would you respecify this when sending the same file to another chat? This PR removes the property.

## Type of change

# How Has This Been Tested?

`npm run typecheck`

# Checklist:


wojpawlik:
I'm not sure what should be done about these errors, for now I can only point you to https://github.com/telegraf/telegraf/blob/e618f587885b988a968de1b2d9b56be24c9a2f2c/src/core/network/client.ts#L64-L82

`Replicate` produces some nasty-looking types...

src/core/replicators.ts
```diff
@@ -0,0 +1,190 @@
+import * as tt from '../../typings/telegram-types'
+import Markup from '../markup'
+const { formatHTML } = Markup
+
+type Replicate<X, K extends keyof (tt.Message & X)> = Required<
+  Pick<tt.Message & X, K>
+>
+type Copy<M extends keyof tt.Telegram> = tt.MakeExtra<M, never>
```
wojpawlik, line 8: `never` could be the default value for 2nd type param of `tt.MakeExtra`, allowing it to be used directly.

KnorpelSenf:
If I understand correctly, that part of the code is only called when answering webhooks, so it would not fix the bugs. Missing thumbnails might be hard to detect, it would make sense if that bug had not been spotted yet. Passing the file_id as string should not be breaking things, this is explicitly documented behaviour. I therefore changed the code accordingly.

typings/telegram-types.d.ts
```diff
@@ -65,9 +65,9 @@ export type DiceEmoji = '🎲' | '🎯' | '🏀'
  */
 export type InputFileVideoNote = Exclude<TT.InputFile, TT.InputFileByURL>
 
-type MakeExtra<
+export type MakeExtra<
   M extends keyof TT.Telegram, // method name
-  P extends keyof Omit<TT.Opts<M>, 'chat_id'> // fields to skip
+  P extends keyof Omit<TT.Opts<M>, 'chat_id'> = never // fields to skip
 > = Omit<TT.Opts<M>, 'chat_id' | P>
```
wojpawlik, line 71: How about improving the param names instead of commenting?
```suggestion
export type MakeExtra<
  MethodName extends keyof TT.Telegram,
  SkipFields extends keyof Omit<TT.Opts<MethodName>, 'chat_id'> = never
> = Omit<TT.Opts<MethodName>, 'chat_id' | SkipFields>
```

wojpawlik (changes requested):

typings/telegram-types.d.ts
KnorpelSenf: I'm hesitant towards using long names as type parameters, but I added proper docs now

src/core/replicators.ts
```diff
@@ -0,0 +1,186 @@
+import * as tt from '../../typings/telegram-types'
+import Markup from '../markup'
+const { formatHTML } = Markup
+
+type Replicate<X, K extends keyof (tt.Message & X)> = Required<
+  Pick<tt.Message & X, K>
+>
```
wojpawlik, line 7: This magic is the last thing I don't like. I want to use the `Message` types after my next PR to `typegram`.

Forgot to mention, wait with releasing `typegram` until the next PR is settled.

wojpawlik:
I'll get back to this after fixing Telegraf type errors caused by unionization.

PR #1119, opened 2020-08-25
Update typegram to version 2.0.0

KnorpelSenf:
Advance the dependency by one commit to the release 2.0.0.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -3,6 +3,7 @@
 import * as tt from '../typings/telegram-types.d'
 import { Middleware, NonemptyReadonlyArray } from './types'
 import Context from './context'
+import { hasProp } from './util'
 
 type MaybeArray<T> = T | T[]
 type MaybePromise<T> = T | Promise<T>
@@ -454,7 +455,9 @@ class Composer<TContext extends Context> implements Middleware.Obj<TContext> {
       const text =
         getText(ctx.message) ??
         getText(ctx.channelPost) ??
-        ctx.callbackQuery?.data ??
+        (hasProp(ctx?.callbackQuery, 'data')
+          ? ctx.callbackQuery?.data
+          : undefined) ??
         ctx.inlineQuery?.query
       if (text === undefined) return next()
       for (const trigger of triggers) {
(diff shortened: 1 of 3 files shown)


src/composer.ts
```diff
@@ -454,7 +455,9 @@ class Composer<TContext extends Context> implements Middleware.Obj<TContext> {
       const text =
         getText(ctx.message) ??
         getText(ctx.channelPost) ??
-        ctx.callbackQuery?.data ??
+        (hasProp(ctx?.callbackQuery, 'data')
+          ? ctx.callbackQuery?.data
+          : undefined) ??
```
wojpawlik, line 460: This code is obnoxious, unionizing `CallbackQuery` might have been a mistake, since there isn't anything like `Composer::mount` or `Composer.on` to help distinguish between the variants... If you still think this should remain unionized, at the very least
```suggestion
        getProp(ctx.callbackQuery, 'data') ??
```

wojpawlik: Alternatively,
```ts
function collapseUnion<T extends object>(t: T): T & UnionToIntersection<Partial<T>> {
  return t
}
```
```ts
collapseUnion(ctx.callbackQuery ?? {})?.data ??
```

Or just slap a `// @ts-expect-error`.

wojpawlik: Just saying, deunionizing `CallbackQuery` would be non-breaking

KnorpelSenf: > This code is obnoxious

I agree this (c|sh)ould be prettier, I'll fix it

> Just saying, deunionizing `CallbackQuery` would be non-breaking

That is plain wrong. There is no difference between unionising this type or any other. You could have the exact same issues with e.g. the `Update` type in another library consuming `typegram`, in other words, we already introduced a breaking change as soon as we unionised the very first type. That's btw also the reason for bumping major.

Maybe it makes more sense to support different types of callback queries in `telegraf` instead of typing the API with inconsistent accuracy in `typegram`?

However, I'm not too happy with the union types in general. First, from my experience, most union and intersection types tend to cause problems sooner or later. I try to avoid them wherever possible. Second, I already ran into issues with those beasts several times within the first 24 hours of using it.
> Expression produces a union type that is too complex to represent. ts(2590)

I'm considering reverting the changes or at least providing the old types along with the unions. I'll investigate if some simple type transformations to map between the two versions can solve the problems. I'm still not sure if the benefits outweigh this hassle.

There's a good reason for most other type systems to not provide untagged union types (that prominently).

PR #1123, opened 2020-08-30
Fix formatHTML to work with emoji

ExposedCat:
Fix wrong offset when formatting text with emoji

# Description

formatHTML() works wrong with emoji:
image

## Type of change


wojpawlik:
What about the tests broken by your change?

PR #1130, opened 2020-09-05
Type `ApiClient::callApi`

wojpawlik:
# Description

1. Type `ApiClient::callApi`,
2. Remove return type annotations from `Telegram`, rely on return type of `ApiClient::callApi`,
3. Fix encountered problems, clean up.

## Type of change

- Types

# How Has This Been Tested?

Precommit hook.

# Checklist:

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -248,7 +248,7 @@ function answerToWebhook(
   response: Response,
   payload: Record<string, unknown>,
   options: ApiClient.Options
-) {
+): Promise<typeof WEBHOOK_REPLY_STUB> {
   if (!includesMedia(payload)) {
     if (isKoaResponse(response)) {
       response.body = payload
@@ -347,6 +347,8 @@ class ApiClient {
     }
 
     debug('HTTP call', method, payload)
+    // I hate these `?? {}`, but I don't know how to
+    // make empty payload `{}` instead of `undefined`
     const buildConfig = includesMedia(payload ?? {})
       ? buildFormDataConfig({ method, ...(payload ?? {}) }, options.agent)
       : buildJSONConfig(payload)
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -60,7 +60,7 @@ const WEBHOOK_REPLY_STUB = {
   webhook: true,
   details:
     'https://core.telegram.org/bots/api#making-requests-when-getting-updates',
-}
+} as const
 
 function includesMedia(payload: Record<string, unknown>) {
   return Object.values(payload).some((value) => {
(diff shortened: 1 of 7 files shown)


src/core/network/client.ts
```diff
@@ -342,8 +347,10 @@ class ApiClient {
     }
 
     debug('HTTP call', method, payload)
-    const buildConfig = includesMedia(payload)
-      ? buildFormDataConfig({ method, ...payload }, options.agent)
+    // I hate these `?? {}`, but I don't know how to
+    // make empty payload `{}` instead of `undefined`
+    const buildConfig = includesMedia(payload ?? {})
+      ? buildFormDataConfig({ method, ...(payload ?? {}) }, options.agent)
```
wojpawlik, line 353: @KnorpelSenf?

KnorpelSenf: Regarding line 352, I suggest you simply allow `undefined` in the method signature of `includesMedia` and in turn go with `return payload !== undefined && Object.values(payload).some(` …

The second one in line 353 is a little problematic to avoid. I suggest you either
- leave the `?? {}` in there and just avoid reading the code ever again, or maybe
- do a QND like this:
   ```ts
    const m: Record<string, unknown> = { method }
    const buildConfig = includesMedia(payload)
      ? buildFormDataConfig({ ...m, ...payload }, options.agent)
      : buildJSONConfig(payload)
   ```
I can't really come up with a nice and clean solution here, but it might just be missing something.

wojpawlik: To be clear, I was wondering whether it would be possible to make for example `Opts<'getChat'>` return `{}` instead of `undefined`.

KnorpelSenf: I see. I'll have a look if I find a way to do this, I agree it would make things more comfortable.

KnorpelSenf: @wojpawlik check v2.2.0

KnorpelSenf: `Opts<'getChat'>` isn't `undefined` btw, but other methods are, I'm assuming you're referring to them.

wojpawlik: I must've meant `Opts<'getMe'>`, I need a break 😅

Sadly, v2.2.0 doesn't help. I did try writing my own `Opts` before I asked you, and no matter what I did, I couldn't convince TS that the value can't be `undefined`.

KnorpelSenf: You're right, sorry. I cloned your fork, reproduced and fixed the issue. The problem is that conditional types are transformed to union types on their way. I published this in 2.2.1, could you try that version? Do let me know if you're having further issues, perhaps I make a mistake while fixing it or something.

wojpawlik: It works now, I'm impressed.
If you have time to spare, it'd be nice to port `Markup` to TS (breaking changes allowed).

PR #1134, opened 2020-09-07
Markup to TS

KnorpelSenf:
# Description

Migrate markup.js to markup.ts.

Markup was refactored so it no longer overwrites its own methods. Therefore it is now required to unwrap the instance again after the building, this is done by accessing `reply_markup`. It might be desired to change the rest of the lib accordingly, so that also plain instances of `Markup` are accepted by all other methods and then unwrapped accordingly if necessary. This is not (yet) part of this PR, though.

## Type of change

- Typings
- Put all properties of Markup class inside member

# How Has This Been Tested?

`npm run typecheck && npm t`

# Checklist:


src/markup.ts
```diff
@@ -0,0 +1,394 @@
...
+
+type MarkupBuilder = Partial<
+  InlineKeyboardMarkup & ReplyKeyboardMarkup & ReplyKeyboardRemove & ForceReply
+>
+
+type Hideable<B> = B & { hide: boolean }
+type HideableKBtn = Hideable<KeyboardButton>
+type HideableIKBtn = Hideable<InlineKeyboardButton>
+
+class Markup {
```
wojpawlik, line 33:
```suggestion
export class Markup {
```

src/markup.ts
```diff
@@ -0,0 +1,394 @@
...
+  public readonly reply_markup: MarkupBuilder = {}
```
wojpawlik, line 34: Ideally, this should be similar to `InlineKeyboardMarkup | ReplyKeyboardMarkup | ReplyKeyboardRemove | ForceReply`... I'll try to make it happen when I'll have time to spare.

src/markup.ts
```diff
@@ -0,0 +1,394 @@
...
+    this.reply_markup.remove_keyboard = !value ? undefined : value
+    return this
+  }
+
+  selective(value = true) {
+    this.reply_markup.selective = value
+    return this
+  }
+
+  extra(options: MarkupBuilder): { reply_markup: MarkupBuilder } {
```
wojpawlik, line 51: Wouldn't it be cool if this returned `Extra`?

wojpawlik (changes requested):

KnorpelSenf:
@wojpawlik I won't have capacity to work on telegraf in the time ahead, but feel free to apply your suggestions by adding your own commits. Otherwise, I'll just take care of it when I'm available again.

src/markup.ts
```diff
@@ -0,0 +1,394 @@
...
+    currentRow.push(btn)
+    index++
+  }
+  if (currentRow.length > 0) {
+    result.push(currentRow)
+  }
+  return result
+}
+
+export = Markup
```
wojpawlik, line 394:
```suggestion
export default Markup
```

src/markup.ts
wojpawlik: I'd make `Markup` generic and immutable to achieve this...

src/markup.ts
```diff
@@ -0,0 +1,303 @@
...
+    hide = false
+  ): Hideable<InlineKeyboardButton.LoginButton> {
+    return {
+      text: text,
+      login_url: { ...opts, url: url },
+      hide: hide,
+    }
+  }
+
+  static formatHTML(text = '', entities: MessageEntity[] = []) {
```
wojpawlik, line 178: This function feels out of place here, how about moving it to `core/replicators`?

wojpawlik: Ideally, the file should export only `removeKeyboard`, `forceReply`, `keyboard` and `inlineKeyboard`.

KnorpelSenf: I factored out this and other helper methods to `core/helpers`. I also took the freedom to force push to clean up the history a bit.

src/markup.ts
```diff
@@ -0,0 +1,204 @@
...
+  resize(this: Markup<ReplyKeyboardMarkup>, value = true) {
+    this.reply_markup.resize_keyboard = value
+    return this
+  }
+
+  oneTime(this: Markup<ReplyKeyboardMarkup>, value = true) {
+    this.reply_markup.one_time_keyboard = value
+    return this
+  }
+}
```
wojpawlik, line 41: I'm not certain, but I believe immutabel chaining is simpler, what do you think?
```suggestion
  selective<T extends ForceReply | ReplyKeyboardMarkup>(
    this: Markup<T>,
    value = true
  ) {
    return new Markup({ ...this.reply_markup, selective: value })
  }

  resize(this: Markup<ReplyKeyboardMarkup>, value = true) {
    return new Markup({ ...this.reply_markup, resize_keyboard: value })
  }

  oneTime(this: Markup<ReplyKeyboardMarkup>, value = true) {
    return new Markup({ ...this.reply_markup, one_time_keyboard: value })
  }
```

wojpawlik:
I cleaned up `Markup`:
obraz

Previously:
obraz


Restoring it back to class would be non-breaking.

If you're okay with this, I'll strenghten some types and merge.

wojpawlik:
Last nitpick, `utils` feels redundant since the files are already in `helpers/` directory: `src/core/helpers/check-utils.ts -> src/core/helpers/check.ts`, `src/core/helpers/formatting-utils.ts -> src/core/helpers/formatting.ts`.

src/markup.ts
KnorpelSenf: I do not see how it's simpler on our end, but I generally like immutability. It allows for reusing one markup instance and making slightly different modifications to it when sending multiple messages. Mutability would cause weird behaviour, so I applied your suggestion in 855f480

KnorpelSenf:
I'm always a bit hesitant towards overloading because it can cause ambiguity and it can sometimes make things harder to understand:
image
How do `arguments`, `bind`, `call`, `login`, and `locationRequest` end up in one list? I find that counterintuitive.

Also, as a new user I'd expect this to work:
image

If you want to keep the shortcut of directly invoking `button`, I'd roll with it, but I suggest that we at least make it an alias for a new function `button.text`. Do you agree?

wojpawlik (approved):
Anything else?

KnorpelSenf:
Only the necessary documentation updates.

I'll write a quick sample bot to validate our work before the 4.0 release.

wojpawlik:
API reference should ideally be generated from source, which is out of scope of this PR, merging.

Issue #1110, opened 2020-08-19
Maximum call stack size exceeded


tourman:
## Context

Unable to understand the reason of the thrown exception.

* Telegraf.js Version: 3.32.0
* Node.js Version: 13.8.0
* Operating System: CentOS 7

## Expected Behavior

No unexpected exception or a proper call stack at least with ability to understand the real reason.

## Current Behavior

```
Exception in PromiseRejectCallback:
/opt/project/node_modules/telegraf/composer.js:355
          return Promise.reject(err)
                         ^

RangeError: Maximum call stack size exceeded

  RangeError: Maximum call stack size exceeded
      at execute (/opt/project/node_modules/telegraf/composer.js:338:24)
      at /opt/project/node_modules/telegraf/composer.js:337:14
      at execute (/opt/project/node_modules/telegraf/composer.js:352:13)
      at /opt/project/node_modules/telegraf/composer.js:337:14
      at execute (/opt/project/node_modules/telegraf/composer.js:352:13)
      at /opt/project/node_modules/telegraf/composer.js:337:14
      at execute (/opt/project/node_modules/telegraf/composer.js:352:13)
      at /opt/project/node_modules/telegraf/composer.js:337:14
      at execute (/opt/project/node_modules/telegraf/composer.js:352:13)
      at /opt/project/node_modules/telegraf/composer.js:337:14
```
## Failure Information (for bugs)

I'm not sure if it is a bug or I did something wrong. I just want to understand the reason or help to fix it.
For some time Telegraf works with no violation. After an undetected time span it fails.

### Steps to Reproduce

Unknown so far. Trying to update to the latest version and observe.

### Failure Logs

See above.

sagadav:
Do you use composer in your code? If yes, attach the code. If not, try using a VPN

vrenda16:
I have the same issue, but in my code i don't use composer. Some solution?

KnorpelSenf:
Does this still happen when building from source? This might have been fixed by #1159

wojpawlik:
If the issue persists, try reproducing it with low `--stack-trace-limit`.

Issue #1128, opened 2020-09-03
Maximum call stack exceeded, uncaughtable


razenization:
## Context

* Telegraf.js Version: ^3.38.0
* Node.js Version: 14.3.0
* Operating System: Win10

## Current Behavior

Receiving the following exception: 
```
RangeError: Maximum call stack size exceeded
    at Object.object.<computed> [as isNative] (/usr/lib/node_modules/pm2/node_modules/source-map-support/source-map-support.js:332:55)
    at Object.CallSiteToString [as toString] (/usr/lib/node_modules/pm2/node_modules/source-map-support/source-map-support.js:265:12)
    at Function.prepareStackTrace (/usr/lib/node_modules/pm2/node_modules/source-map-support/source-map-support.js:416:37)
    at maybeOverridePrepareStackTrace (internal/errors.js:86:29)
    at prepareStackTrace (internal/errors.js:65:5)
    at formatError (internal/util/inspect.js:1085:19)
    at formatRaw (internal/util/inspect.js:892:14)
    at formatValue (internal/util/inspect.js:746:10)
    at inspect (internal/util/inspect.js:287:10)
    at formatWithOptions (internal/util/inspect.js:1910:40)
    at Object.Console.<computed> (internal/console/constructor.js:304:10)
    at Object.warn (internal/console/constructor.js:314:61)
    at Object.consoleCall (<anonymous>)
    at execute (/var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:374:12)
    at /var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:355:14
    at execute (/var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:370:13)
    at /var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:355:14
    at execute (/var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:370:13)
    at /var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:355:14
    at execute (/var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:370:13)
    at /var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:355:14
    at execute (/var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:370:13)
    at /var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:355:14
    at execute (/var/www/html/ponchesme/backend/node_modules/telegraf/composer.js:370:13)
[...]

PR #1156, opened 2020-10-14
Re-enable short middleware notation

KnorpelSenf:
# Description

With the current version of the code base, it is not possible to write something like this:

```ts
bot.on('photo', ctx => {
  ctx.session.counter ??= 0
  ctx.session.counter++
})
```

Middleware is required to return `Promise<void>`. It cannot return `void`. This is nasty. It requires the counterintuitive

```ts
bot.on('photo', async ctx => {
  ctx.session.counter ??= 0
  ctx.session.counter++
})
```
with an `async` middleware that does not even use `await`.

Furthermore, with the current codebase, it is not possible to write something like this:

```ts
bot.command('start', ({ reply }) => reply('Welcome! I count your photos. /help'))
```

The problem is that middleware must return `Promise<void>`, so even adding `async` and `await` does not work here.
We have to allow returning arbitrary values from the middleware function in order to allow this convenient pattern.
Otherwise, strange workarounds like this would be necessary:

```ts
bot.command('start', async ({ reply }) => void reply('Welcome! I count your photos. /help'))
// or even
bot.command('start', async ({ reply }) => {
  reply('Welcome! I count your photos. /help')
})
```

Fixes the above issue.

May help with
- #1110 
- #1128

## Type of change

# How Has This Been Tested?

`npm t && npm run typecheck`

This PR is a result of the test development performed after https://github.com/telegraf/telegraf/pull/1134#issuecomment-708249167

# Checklist:

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -643,9 +643,7 @@ export class Composer<TContext extends Context>
       return await execute(0, ctx)
       async function execute(i: number, context: TContext): Promise<unknown> {
         if (!(context instanceof Context)) {
-          return await Promise.reject(
-            new Error('next(ctx) called with invalid context')
-          )
+          throw new Error('next(ctx) called with invalid context')
         }
         if (i <= index) {
           return await Promise.reject(new Error('next() called multiple times'))
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -638,9 +638,9 @@ export class Composer<TContext extends Context>
     if (middlewares.length === 1) {
       return Composer.unwrap(middlewares[0])
     }
-    return async (ctx, next) => {
+    return (ctx, next) => {
       let index = -1
-      return await execute(0, ctx)
+      return execute(0, ctx)
       async function execute(i: number, context: TContext): Promise<void> {
         if (!(context instanceof Context)) {
           throw new Error('next(ctx) called with invalid context')
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -653,15 +653,9 @@ export class Composer<TContext extends Context>
         if (!handler) {
           return
         }
-        try {
-          return await Promise.resolve(
-            handler(context, async (ctx = context) => {
-              await execute(i + 1, ctx)
-            })
-          )
-        } catch (err) {
-          return await Promise.reject(err)
-        }
+        await handler(context, async (ctx = context) => {
+          await execute(i + 1, ctx)
+        })
       }
     }
   }
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -653,7 +653,7 @@ export class Composer<TContext extends Context>
         index = i
         const handler = middlewares[i] ? Composer.unwrap(middlewares[i]) : next
         if (!handler) {
-          return await Promise.resolve()
+          return
         }
         try {
           return await Promise.resolve(
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -638,28 +638,31 @@ export class Composer<TContext extends Context>
     if (middlewares.length === 1) {
       return Composer.unwrap(middlewares[0])
     }
-    return (ctx, next) => {
+    return async (ctx, next) => {
       let index = -1
-      return execute(0, ctx)
-      function execute(i: number, context: TContext): Promise<void> {
+      return await execute(0, ctx)
+      async function execute(i: number, context: TContext): Promise<unknown> {
         if (!(context instanceof Context)) {
-          // prettier-ignore
-          return Promise.reject(new Error('next(ctx) called with invalid context'))
+          return await Promise.reject(
+            new Error('next(ctx) called with invalid context')
+          )
         }
         if (i <= index) {
-          return Promise.reject(new Error('next() called multiple times'))
+          return await Promise.reject(new Error('next() called multiple times'))
         }
         index = i
         const handler = middlewares[i] ? Composer.unwrap(middlewares[i]) : next
         if (!handler) {
-          return Promise.resolve()
+          return await Promise.resolve()
         }
         try {
-          return Promise.resolve(
-            handler(context, (ctx = context) => execute(i + 1, ctx))
+          return await Promise.resolve(
+            handler(context, async (ctx = context) => {
+              await execute(i + 1, ctx)
+            })
           )
         } catch (err) {
-          return Promise.reject(err)
+          return await Promise.reject(err)
         }
       }
     }
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -641,12 +641,12 @@ export class Composer<TContext extends Context>
     return async (ctx, next) => {
       let index = -1
       return await execute(0, ctx)
-      async function execute(i: number, context: TContext): Promise<unknown> {
+      async function execute(i: number, context: TContext): Promise<void> {
         if (!(context instanceof Context)) {
           throw new Error('next(ctx) called with invalid context')
         }
         if (i <= index) {
-          return await Promise.reject(new Error('next() called multiple times'))
+          throw new Error('next() called multiple times')
         }
         index = i
         const handler = middlewares[i] ? Composer.unwrap(middlewares[i]) : next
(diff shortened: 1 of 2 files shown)


src/composer.ts
```diff
@@ -638,28 +638,31 @@ export class Composer<TContext extends Context>
...
         }
         if (i <= index) {
-          return Promise.reject(new Error('next() called multiple times'))
+          return await Promise.reject(new Error('next() called multiple times'))
         }
         index = i
         const handler = middlewares[i] ? Composer.unwrap(middlewares[i]) : next
         if (!handler) {
-          return Promise.resolve()
+          return await Promise.resolve()
```
KnorpelSenf, line 656:
```suggestion
          return
```

wojpawlik (changes requested):
There's a lot to unpack here.

> with the current codebase, it is not possible to write something like this:
> 
> ```ts
> bot.command('start', ({ reply }) => reply('Welcome! I count your photos. /help'))
> ```

Currently, the correct way is:

```ts
bot.command('start', ({ reply }) => {
  await reply('Welcome! I count your photos. /help')
})
```

It's not an issue to me personally, but I can see how it can be unintuitive or inconvenient to others. To account for this, return type can be changed to `Promise<unknown>`, bare `unknown` is overkill.

<hr />

> Middleware is required to return `Promise<void>`. It cannot return `void`. This is nasty. It requires the counterintuitive
> 
> ```ts
> bot.on('photo', async ctx => {
>   ctx.session.counter ??= 0
>   ctx.session.counter++
> }
> ```

From my perspective, this is only an issue if you use `session` or `state` or other synchronous storage and https://eslint.org/docs/rules/require-await. Again, `unknown` is overkill here, adding `| void` suffices.

So, the return type avoiding both "problems" is `Promise<unknown> | void`; [we've come full circle](https://github.com/telegraf/telegraf/pull/1108/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aL17) :)

<hr />

Anyway, perhaps changing `execute` to async function will clarify traceback enough to understand and fix #1110 / #1128.

KnorpelSenf:
I agree that using `async`/`await` for `execute` is orthogonal to the initial idea of this pull request, thanks for pointing this out. I applied it.

I do not think that `unknown` as return type is overkill. Imagine having a handler function `h` that returns a value for some reason. Maybe it updates some global variable and then returns that value because whatever, the function is used in some other place.

The return type `void | Promise<unknown>` would then still interdict what I demonstrated in the second example, just with `h` instead of `reply` (i.e. `ctx => h(ctx)` as middleware). This `void | Promise<unknown>` incidentally allows all handler functions if they return a `Promise`, but not others. This is making `telegraf` less approachable.

I would probably not realistically write such code myself, I see that I'm also being nitpicky here. However, on the other side, I fail to see a reason why it helps library consumers spot bugs early because their middleware is returning a value of the wrong type. Only allowing a promisified `unknown` (or a regular `void`) but disallowing regular values creates this weird edgecase that we do not need. In what scenario does it make sense do explicitly forbid non-promise non-void values but allow all others?

wojpawlik:
Nitpick: `Promise<unknown>` clearly expresses that returning [`Fluture`](https://github.com/fluture-js/Fluture) or other Promise-alternative is not supported.

KnorpelSenf:
Valid point. We will soon run into a `NitpickStackOverflowException`.

So what are we going to do about this?

wojpawlik:
> ```ts
> bot.on('photo', ctx => {
>   ctx.session.counter ??= 0
>   ctx.session.counter++
> })
> ```

This code is unlikely to be written, it probably should `return next()` if not `return await next()`.

KnorpelSenf:
So how would you actually develop the example bot you're referring to? Do you find it idiomatic to write
```
return Promise.resolve() // serves no purpose, only fixes compilation error
```
or force users to `@ts-expect-error` their middleware?

What if I told you that I have a bot running in production that needs exactly this because it silently stores something in the session for certain types of messages? I don't see a reason why this bot is unlikely to exist

wojpawlik:
I'd love to see that piece of code.

`return Promise.resolve()` is a solution, as is turning the function `async`.

KnorpelSenf:
I'd love to be allowed to show you. The example snippet above gets pretty close to it (storing info about most recent message instead of photo counting).

It is a solution to an artificial problem you're creating because of … what reason again? I hope your comment about promise alternatives was a joke (I took it as such) but now I'm no longer sure …

wojpawlik:
While I doubt anyone would try using https://github.com/fluture-js/Fluture with Telegraf, I cannot rule it out. TypeScript should still ensure correctness.

I'm bringing this up because of #1262.

Why are you storing information about a message and not allowing it to finish processing by `return next()`?

KnorpelSenf:
Not sure I understand the question. Synchronous functions finish processing “immediately” so to say, i.e. after less than a millisecond in my case (see snippet above). Can you rephrase?

wojpawlik:
In `express`, you need to call `res.end()` exactly once, explicitly or implicitly. Therefore, each middleware needs to call `next()` xor `res.end()`. If you do neither, response remains open and client will time out.

`telegraf` automatically closes responses if needed, so consequences aren't that severe. Still, it feels like a latent bug: if you don't call `next()`, and later add another middleware listening for the same update, it won't be called, and you'll have to troubleshoot.

Why do you need to effectively swallow the update? Why should `telegraf` make it easy to do it?

KnorpelSenf:
`telegraf` and `express` are not really comparable. `express` has an open connection during middleware execution, while for `telegraf`, this is only true when using webhooks.

Calling the missing `next` call a “bug” reminds me of the old discussion of whether to use fallthrough cases in `switch` statements.

It is very nice that `telegraf` can be used this way:

```ts
bot.url('example.com')   // very specific
bot.on('text')           // somewhat specific
bot.on('message')        // pretty general
bot.use()                // everything else
```

This looks somewhat like a `switch` statement in the sense that the conditions are checked one by one until the first one matches. This is different from the switch statement that we “`break`” automatically so to say. Calling `next` can be seen as an _explicit_ fallthrough.

If you structure your bots differently, that's fine, but this has in fact proven well for me and I see no reason to be so opinionated about how I may or may not use middleware.

Why should `telegraf` make it harder to use middleware by constraining types in a way that suits you but not necessarily other people?

wojpawlik:
It's perfectly fine not to call `next()` iff you invoke some API methods. Either way, it's easy to return a `Promise<unknown>`. What's an use-case for doing neither? Except `.filter` / `.drop`, which the lib already covers.

KnorpelSenf:
Removing `| void` boils down to disallowing synchronous middleware.

I believe I already named a number of good use-cases (there is another one below), and I told you that existing code (running in production) relies on middleware that does not return a promise.

I have three questions about your suggestion.

1. How would you rewrite the following snippet, and how would the new version be an improvement to the original?

```ts
bot.on('left_chat_member', (ctx) => {
  // delete session if bot is removed from group
  if (ctx.message.left_chat_member.id === ctx.botInfo.id) {
    ctx.session = {}
  }
})
```

`telegraf` has zero problems with synchronous middleware as of today.

2. What changes do you want to introduce in the future that asynchronous middleware a requirement?

We can easily make supplied middleware asynchronous by wrapping it with a `Promise.resolve` call if necessary.

3. How do we benefit from users migrating all of their synchronous middleware to asynchronous one, instead doing it once in a single place by applying the above suggestion?

wojpawlik:
> How would you rewrite the following snippet

```ts
bot.on('left_chat_member', async (ctx, next) => {
  // delete session if bot is removed from group
  if (ctx.message.left_chat_member.id === ctx.botInfo.id) {
    ctx.session = {}
  }
  return await next()
})
```

> how would the new version be an improvement to the original?

It allows to attach middleware that _actually_ responds to the message:

```ts
bot.on('left_chat_member', async (ctx) => {
  if (ctx.message.left_chat_member.id !== ctx.botInfo.id) {
    await ctx.reply(`Bye ${ctx.message.left_chat_member.first_name}!`)
  }
})
```

PR #1159, opened 2020-10-20
Potentially improve stacktraces by moving to async/await

KnorpelSenf:
# Description

The `async`/`await` pattern can sometimes improve stack traces. However, it is not applied consistently in the library. This PR performs some automatic code transformations to reasonable places such that `async`/`await` is used instead of `Promise`s.

(Also, two  `ts-expect-error` comments were explained with a comment each.)

## Type of change

- [X] Could make future stack traces more readable, does not introduce functional changes

# How Has This Been Tested?

`npm run lint && npm t`

# Checklist:

### DIFF
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -21,12 +21,13 @@ class Telegram extends ApiClient {
   /**
    * Get download link to a file
    */
-  async getFileLink(
-    fileId: string | tt.File | Promise<string> | Promise<tt.File>
-  ) {
-    fileId = await fileId
-    if (typeof fileId === 'string') return await this.getFile(fileId)
-    if (fileId.file_path !== undefined) return fileId
+  async getFileLink(fileId: string | tt.File) {
+    if (typeof fileId === 'string') {
+      return await this.getFile(fileId)
+    }
+    if (fileId.file_path !== undefined) {
+      return fileId
+    }
     const file = await this.getFile(fileId.file_id)
     // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
     return `${this.options.apiRoot}/file/bot${this.token}/${file.file_path}`
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -23,14 +23,12 @@ class Telegram extends ApiClient {
    */
   async getFileLink(fileId: string | tt.File) {
     if (typeof fileId === 'string') {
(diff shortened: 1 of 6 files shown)


wojpawlik (changes requested):
Try enabling https://github.com/telegraf/telegraf/blob/63d5e25b1c6130b4b8c0043d2c74b8d0fe131156/.eslintrc#L16

KnorpelSenf:
> Try enabling
> 
> https://github.com/telegraf/telegraf/blob/63d5e25b1c6130b4b8c0043d2c74b8d0fe131156/.eslintrc#L16

This is a bad idea.

`✖ 244 problems (162 errors, 82 warnings)`

wojpawlik:
Just now I realized that your `getFileLink` is incorrect, it should always return absolute URL. Revert your changes if you don't have a good idea how to accomplish it.

KnorpelSenf:
You're right! That's crazy, I used automatic code transformation built into VSCode to do this. I expected that you can trust that tool, so I didn't consider the logic to be completely different. I submitted https://github.com/microsoft/vscode/issues/109401 because of this.

In the meantime, I did the transformation here manually. I will check all other changes once again, please do so, too, if you haven't already, just in case I'll miss something when I do it now.

src/telegram.ts
```diff
@@ -21,18 +21,14 @@ class Telegram extends ApiClient {
...
-      })
-      .then(
-        (file) =>
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          `${this.options.apiRoot}/file/bot${this.token}/${file.file_path}`
-      )
+  async getFileLink(fileId: string | tt.File) {
+    if (typeof fileId === 'string') {
+      fileId = await this.getFile(fileId)
+    } else if (fileId.file_path === undefined) {
```
wojpawlik, line 27:
```suggestion
    } else if (fileId.file_path !== undefined) {
```

KnorpelSenf: In the old code,
```ts
if (fileId.file_path !== undefined) return fileId
return this.getFile(fileId.file_id)
```
made sure that `getFile` is only called when `fileId.file_path !== undefined` is `false`.

You're suggesting to call `getFile` when `fileId.file_path !== undefined` is `true`.

PR #1166, opened 2020-10-27
Fix duplicate overload

uWynell:
There were 2 identical overloads for 1 method and I removed one

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -618,9 +618,6 @@ export class Composer<TContext extends Context>
   ): Middleware.Fn<TContext>
   static compose<TContext extends Context>(
     middlewares: ReadonlyArray<Middleware<TContext>>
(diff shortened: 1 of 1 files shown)


uWynell:
They are the identical, why tests are failing!?

wojpawlik:
https://www.typescriptlang.org/docs/handbook/functions.html#overloads

uWynell:
> https://www.typescriptlang.org/docs/handbook/functions.html#overloads
Why there are **two absolutely identical** overloads?

wojpawlik:
The second one is not an overload. It's an _internal_ signature, not exposed to callers by default.

uWynell:
> The second one is not an overload. It's an _internal_ signature, not exposed to callers by default.
where can I read about it?

PR #1177, opened 2020-11-11
Typecheck examples

wojpawlik:
https://github.com/telegraf/telegraf/issues/1074#issuecomment-724467372

I decided that the best way to verify that examples are correct is to typecheck them. This revealed several typing issues, some of which I fixed.

## Type of change

- Documentation (typos, code examples or any documentation update)


wojpawlik:
Unsure what to do with the rest, please take over @KnorpelSenf

KnorpelSenf:
Depends on https://github.com/KnorpelSenf/typegram/issues/11.

@wojpawlik how is it even possible that sending files works? We're passing `{ "url": "http://example.com/file", "filename": "my-file" }` to the Telegram servers. Where does it say that it works that way? The docs [state](https://core.telegram.org/bots/api#sending-files) that we can only pass a string as URL. What am I missing here?

docs/examples/shop-bot.js
```diff
@@ -13,9 +12,9 @@ const invoice = {
     { label: 'Working Time Machine', amount: 4200 },
     { label: 'Gift wrapping', amount: 1000 }
   ],
-  payload: {
+  payload: JSON.stringify({
```
KnorpelSenf, line 15: @wojpawlik can you confirm that this is correct?

docs/examples/shop-bot.js
```diff
@@ -32,14 +31,14 @@ const shippingOptions = [
...
-]).extra()
+  Markup.button.pay('💸 Buy'),
+  Markup.button.url('❤️', 'http://telegraf.js.org')
+])
 
 const bot = new Telegraf(process.env.BOT_TOKEN)
 bot.start(({ replyWithInvoice }) => replyWithInvoice(invoice))
 bot.command('buy', ({ replyWithInvoice }) => replyWithInvoice(invoice, replyOptions))
-bot.on('shipping_query', ({ answerShippingQuery }) => answerShippingQuery(true, shippingOptions))
+bot.on('shipping_query', ({ answerShippingQuery }) => answerShippingQuery(true, shippingOptions, undefined))
```
KnorpelSenf, line 41: We may need to redefine that function's signature. It would be nice if `undefined` is not required here.

package.json
```diff
@@ -27,13 +27,16 @@
     "telegraf": "./bin/telegraf"
   },
   "scripts": {
-    "lint": "eslint src/",
-    "pretest": "tsc",
+    "typecheck": "tsc --noEmit",
+    "build": "npm run build:lib && npm run build:docs",
+    "build:lib": "npm run prepare",
+    "build:docs": "tsc --build docs/examples && npm run typedoc",
```
KnorpelSenf, line 33: It is very convenient to have those shortcuts here. If we should try not to use `package.json` as a build script container, we may want to have some kind of alternative that supports a similarly easy way to build the project.

src/telegram-types.ts
```diff
@@ -1,11 +1,54 @@
...
+export interface InputFileByReadableStream {
+  source: NodeJS.ReadableStream
+}
+export interface InputFileByBuffer {
+  source: Buffer
+}
+export interface InputFileByURL {
+  url: string
+  filename?: string
+}
```
KnorpelSenf, line 31: Is this correct? I'm not sure if this accurately describes the `InputFile`s that `telegraf` can handle.

src/telegram-types.ts
```diff
@@ -140,8 +185,9 @@ export type ExtraVideo = MakeExtra<'sendVideo', 'video'>
 export type ExtraVideoNote = MakeExtra<'sendVideoNote', 'video_note'>
 export type ExtraVoice = MakeExtra<'sendVoice', 'voice'>
 
-export type IncomingMessage = TT.Message
+export type IncomingMessage = Message
 
+/** @deprecated */
```
KnorpelSenf, line 190: This is a partial redefinition of the arguments of `sendInvoice` in `typegram`. Should this be removed?

src/telegram.ts
```diff
@@ -269,7 +269,7 @@ class Telegram extends ApiClient {
    */
   sendSticker(
     chatId: number | string,
-    sticker: tt.InputFile,
+    sticker: tt.Opts<'sendSticker'>['sticker'],
```
KnorpelSenf, line 272: It is important to note here that `string` must not be part of `InputFile`. It is not possible to use a URL or a file_id when sending a sticker, and `typegram` already models this correctly. The correct type is simply `InputFile` as it is defined.

src/telegram.ts
```diff
@@ -282,7 +282,7 @@ class Telegram extends ApiClient {
    */
   sendVideo(
     chatId: number | string,
-    video: tt.InputFile,
+    video: tt.Opts<'sendVideo'>['video'],
```
KnorpelSenf, line 285: `typegram` makes sure that `string` is included in this definition. Videos can be sent via file_id.

KnorpelSenf (commented):
I added scripts to `package.json` that make it simple to build the docs.

All typing errors are resolved, except in two areas: scenes and wizards. It looks to me like the library itself needs some tweaks here before the examples can work.

src/telegram-types.ts
wojpawlik: @dotcypress ?

package.json
wojpawlik: I understand `build`, but what's the advantage of `build:lib` over `prepare` and `build:docs` over `typedoc`?

wojpawlik: I'd be fine with renaming `typedoc` to `build:docs` actually. `prepare` could be changed to `tsc --build docs/examples && mjs-entry src/index.ts > lib/index.mjs`, it'd still build the lib: https://github.com/telegraf/telegraf/blob/732f2c67bf8f55c9b541bddc45ff0c3e2c593dc3/docs/examples/tsconfig.json#L11

docs/examples/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "composite": true,
+    "checkJs": true,
+    "esModuleInterop": true,
+    "noEmit": true,
+    "paths": { "telegraf": ["../../"] },
+    "baseUrl": "."
```
wojpawlik, line 9: `allowJs` and `baseUrl` are unnecessary in typescript 4.1.1-rc.
```suggestion
```

src/core/network/client.ts
```diff
@@ -346,9 +346,17 @@ class ApiClient {
...
+
+    const hasMedia = includesMedia(payload)
+    let buildConfig: Promise<RequestInit>
+    const configPayload: { [key: string]: unknown } = payload
+    if (hasMedia) {
+      configPayload[method] = method
+      buildConfig = buildFormDataConfig(configPayload, options.agent)
+    } else {
+      buildConfig = buildJSONConfig(payload)
+    }
```
wojpawlik, line 358: ...what does it improve?

KnorpelSenf: It makes the code compile. I have no idea why it suddenly didn't anymore. Do let me know what the issue was in case you decide to investigate this.

package.json
KnorpelSenf: I'm very open for restructurings. That said, the idea was the following:
`prepare`: runs the build automatically before `npm publish`
`typedoc`: generates docs

`build`: no-brainer, builds everything from scratch, e.g. after fresh `npm i`
`build:lib`: builds the lib without the doc generation overhead, currently reuses the prepare hook for that
`build:docs`: checks the docs for types and then builds them, currently reuses the doc generation script for the build

Therefore, the difference between between `build:lib` and `prepare` is that `build:lib` is a name for humans to use and `prepare` is a name for npm to use. The other thing is that `build:docs` typechecks the docs before generating them with `typedoc`.

I like to have a naming scheme that makes sense to invoke manually, even if it is just aliases for npm hooks. I don't have a preference in which direction that alias should go, and again, we can also make something totally different from what I suggested here :)

src/core/network/client.ts
KnorpelSenf: `Type '{ method: M; } & Opts<M>' is not assignable to type '{ [key: string]: unknown; }'.ts(2322)` was the error, perhaps related to https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#k-string-unknown-is-no-longer-a-wildcard-assignment-target

src/core/network/client.ts
```diff
@@ -346,9 +346,17 @@ class ApiClient {
...
     debug('HTTP call', method, payload)
-    const buildConfig = includesMedia(payload)
-      ? buildFormDataConfig({ method, ...payload }, options.agent)
-      : buildJSONConfig(payload)
+
+    const hasMedia = includesMedia(payload)
+    let buildConfig: Promise<RequestInit>
+    const configPayload: { [key: string]: unknown } = payload
+    if (hasMedia) {
+      configPayload[method] = method
```
wojpawlik, line 354:
```suggestion
      configPayload.method = method
```

docs/examples/shop-bot.js
wojpawlik: BY adding `true` and `false` overloads?

package.json
wojpawlik: Please open a separate pull request. I'll drop these upon rebasing to keep discussion focused and avoid merge conflicts.

wojpawlik: > `build`: no-brainer, builds everything from scratch, e.g. after fresh `npm i`

`prepare` runs automatically after `npm i`.

KnorpelSenf: > Please open a separate pull request.

Confer #1191. Note that it is not possible to include docs typechecks in that PR as it depends on this one.

docs/examples/shop-bot.js
KnorpelSenf: How are you going to make `true` and `false` overloads for `string | undefined`?

KnorpelSenf: Nvm, you mean overloads for `ok`. Yes, that makes a lot of sense.

src/telegram.ts
KnorpelSenf: Also confer https://github.com/KnorpelSenf/typegram/issues/15

wojpawlik:
Everything's alright @KnorpelSenf? I'll merge this (without squashing) if yes, even tho last 2 examples aren't checked.

PR #1186, opened 2020-11-18
Add AWS Lambda Example Code

septs:
# Description

Add AWS Lambda Example code

Fixes # (issue)

## Type of change

# Checklist:


nezort11:
@septs  @KnorpelSenf  , am I right that in order for `sendChatAction` (with `replyToWebhook`) work in serverless environment (Lambda, Google functions, etc.) i need to add `webhookReply: false` ?

KnorpelSenf:
Yes. In fact, if you ever want to perform two API calls for the same incoming request (no matter if you call `sendChatAction` or something else) then you must disable webhook reply that way. 

This is because AWS doesn't allow your code to run after the request was completed by sending a response.

Moreover, note that this means that the request will likely take a few seconds until it is completed. That is acceptable for very small bots with just a few hundred users, but it leads to drastic performance problems for all larger projects. Telegram only ever sends up to 100 updates in parallel so if each of them takes 10 seconds to process, your throughput will drop to 10 messages/second.

Finally, if your requests take even longer than that, your bot may receive the same updates many times, which leads to an entirely new type of problem. I explained it in depth here: https://grammy.dev/guide/deployment-types#ending-webhook-requests-in-time (I don't think telegraf has docs for this)

KnorpelSenf:
Not all serverless environments work that way btw, but AWS does

nezort11:
Thanks a lot for the clarifications.

PR #1187, opened 2020-11-19
Fix telegram.getFileLink with local bot API instances

AndrewLaneX:
# Description

Local bot API instances return the absolute path to the file and `getFileLink` breaks since it adds `{apiRoot}/file/bot{token}/` to the path

## Type of change

- [x] This change requires a documentation update

# Checklist:

- [ ] ~I have added tests that prove my fix is effective or that my feature works~

### DIFF
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -31,7 +31,7 @@ class Telegram extends ApiClient {
 
     // Local bot API instances return the absolute path to the file
     if (fileId.file_path && isAbsolute(fileId.file_path)) {
-      return 'file://' + fileId.file_path
+      return new URL(fileId.file_path.replace(/^\/?/, '/'), 'file:')
     }
 
     // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -19,7 +19,7 @@ class Telegram extends ApiClient {
   }
 
   /**
-   * Get download link or path to a file
+   * Get download link to a file
    */
   async getFileLink(fileId: string | tt.File) {
     if (typeof fileId === 'string') {
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -1,6 +1,7 @@
 import * as replicators from './core/replicators'
 import * as tt from './telegram-types'
 import ApiClient from './core/network/client'
+import { isAbsolute } from 'path'
 
 class Telegram extends ApiClient {
   /**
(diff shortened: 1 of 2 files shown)


src/telegram.ts
```diff
@@ -19,14 +19,20 @@ class Telegram extends ApiClient {
...
   async getFileLink(fileId: string | tt.File) {
     if (typeof fileId === 'string') {
       fileId = await this.getFile(fileId)
     } else if (fileId.file_path === undefined) {
       fileId = await this.getFile(fileId.file_id)
     }
+
+    // Local bot API instances return the absolute path to the file
+    if (fileId.file_path?.startsWith('/')) {
+      return fileId.file_path
```
wojpawlik, line 33: Why not return a `file:` URL?
```suggestion
      return 'file://' + fileId.file_path
```

wojpawlik (approved):

src/telegram.ts
```diff
@@ -19,14 +19,20 @@ class Telegram extends ApiClient {
...
    */
   async getFileLink(fileId: string | tt.File) {
     if (typeof fileId === 'string') {
       fileId = await this.getFile(fileId)
     } else if (fileId.file_path === undefined) {
       fileId = await this.getFile(fileId.file_id)
     }
+
+    // Local bot API instances return the absolute path to the file
+    if (fileId.file_path?.startsWith('/')) {
```
wojpawlik, line 32: What about Windows' paths? They don't start with `/`, do they?

wojpawlik (changes requested):
Undo your changes to docs.

src/telegram.ts
wojpawlik: [`path.isAbsolute`](https://nodejs.org/api/path.html#path_path_isabsolute_path)?

wojpawlik:
`file:///` implies `localhost`, which is wrong when local API instance is running on different machine/container...

AndrewLaneX:
I believe the bot API only returns absolute paths if you run it with `--local`.

wojpawlik:
That's correct, but does `--local` prevent connections from outside `localhost`?

AndrewLaneX:
I don't think so.

src/telegram.ts
```diff
@@ -27,6 +28,12 @@ class Telegram extends ApiClient {
     } else if (fileId.file_path === undefined) {
       fileId = await this.getFile(fileId.file_id)
     }
+
+    // Local bot API instances return the absolute path to the file
+    if (fileId.file_path && isAbsolute(fileId.file_path)) {
+      return 'file://' + fileId.file_path
```
wojpawlik, line 34: In case Windows paths don't start with a slash:

```suggestion
      return 'file://' + fileId.file_path.replace(/^\//, '/')
```

AndrewLaneX: I don't understand. If it doesn't start with a slash, nothing will happen.

wojpawlik: One slash will be missing

wojpawlik:
Don't worry about unhappy CI.

PR #1191, opened 2020-11-24
Add build scripts

KnorpelSenf:
# Description

Adds build scripts for convenient building of the library and the docs.

- `prepare`: npm hook that calls the build script both before `npm publish` and after `npm install`
- `typecheck`: checks the types of both lib and docs
- `typecheck:lib`: checks the types of the library
- `typecheck:docs`: checks the types of the docs (examples bots there)
- `build`: builds both lib and docs
- `build:lib`: builds the library
- `build:docs`: builds the docs
- `pretest`: npm hook that builds the lib before `npm test`
- `test`: runs the tests
- `lint`: lints the project
- `clean`: remove generated files

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "typecheck": "npm run typecheck:lib && npm run typecheck:docs",
     "typecheck:lib": "tsc --noEmit",
     "typecheck:docs": "tsc --build docs/examples",
+    "pretest": "npm run build:lib",
     "test": "ava test/*",
     "lint": "eslint .",
     "clean": "git clean -fX docs/build/ lib/ typings/"
--- a/package.json
+++ b/package.json
@@ -27,13 +27,17 @@
     "telegraf": "./bin/telegraf"
   },
   "scripts": {
-    "lint": "eslint .",
-    "pretest": "tsc --build docs/examples",
+    "pretest": "npm run typecheck",
+    "prepare": "npm run build",
+    "build": "npm run build:lib && npm run build:docs",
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
@wojpawlik the conflicts are resolved.

You mentioned that checking the types of the example bots builds the library despite `noEmit` being set to true. Do you want this behaviour, and if not, can be even avoid it?

I'm fine with either option, so IMO can you merge this (unless you request changes).

wojpawlik (changes requested):
I'd move `mjs-entry` to `prepare`, it should only fail when building does, so running it checks nothing.

`tsc` does less than and should be no faster than `tsc -b docs/examples/`

```json
"prepare": "npm run build && && mjs-entry src/index.ts > lib/index.mjs",
"build:lib": "tsc --build docs/examples/",
```

package.json
```diff
@@ -27,13 +27,14 @@
     "telegraf": "./bin/telegraf"
   },
   "scripts": {
-    "lint": "eslint .",
-    "pretest": "tsc --build docs/examples",
-    "test": "ava test/*",
-    "typedoc": "typedoc",
-    "clean": "git clean -fX docs/build/ lib/ typings/",
+    "prepare": "npm run build && npm run build:docs && mjs-entry src/index.ts > lib/index.mjs",
```
wojpawlik, line 30: Docs aren't necessary to use the lib
```suggestion
    "prepare": "npm run build && mjs-entry src/index.ts > lib/index.mjs",
```

wojpawlik (changes requested):

Issue #1139, opened 2020-09-15
Add an example for Stage and Scene with a custom context


negebauer:
**Is your feature request related to a problem? Please describe.**

I'm new to typescript and I've been trying to write a bot that has uses session with a custom session object and stages.
I haven't found any examples of Stages using a custom context type.

**Describe the solution you'd like**

A complete example with a bot with a custom type and stages using that custom type.

**Additional context**

This is the code I have

```ts
import { Telegraf, Context, Stage, BaseScene } from 'telegraf'
import LocalSession from 'telegraf-session-local'

interface Session {
  bar: string
}

interface BotContext extends Context {
  session: Session
}

const bot = new Telegraf<BotContext>('foo')
const session = new LocalSession<BotContext>()
bot.use(session.middleware())

const scene = new BaseScene('scene')
scene.command('back', ctx => ctx.scene.leave())

const stage = new Stage([scene], { ttl: 10 })
bot.use(stage.middleware())

bot.command('foo', ctx => ctx.scene.enter('scene'))

bot.launch()
```

`bot.use(stage.middleware())` shows the following

```
Argument of type 'MiddlewareFn<SceneContextMessageUpdate>' is not assignable to parameter of type 'Middleware<BotContext>'.
  Type 'MiddlewareFn<SceneContextMessageUpdate>' is not assignable to type 'MiddlewareFn<BotContext>'.
    Property 'scene' is missing in type 'BotContext' but required in type 'SceneContextMessageUpdate'.
```

and `bot.command('foo', ctx => ctx.scene.enter('scene'))` shows

```
Property 'scene' does not exist on type 'BotContext'.
```

I though that using `BaseScene<BotContext>` and `Stage<BotContext>` would make it work, but I get the following instead

For `BaseScene<BotContext>`

```
Type 'BotContext' does not satisfy the constraint 'SceneContextMessageUpdate'.
  Property 'scene' is missing in type 'BotContext' but required in type 'SceneContextMessageUpdate'.
```

And for `Stage<BotContext>`

```
Type 'BotContext' does not satisfy the constraint 'SceneContextMessageUpdate'.
```

***

I did get to code that compiles by doing the following

```ts
import { Telegraf, Context, Stage, BaseScene } from 'telegraf'
import LocalSession from 'telegraf-session-local'
import { SceneContextMessageUpdate } from '../node_modules/telegraf/typings/stage.d'

interface Session {
  foo: string
}

[...]

Issue #705, opened 2019-07-24
How can I get more documentation on Scenes, Stages and stuff?


takdw:
Hi all, I was checking out the docs and noticed that it doesn't say a lot about Scenes and Stages. Where can I get more detailed docs since I'm new to this module?

dvgololobov:
You can use scenes like some microbot instance. [Examples](https://github.com/telegraf/telegraf/tree/master/docs/examples).

IMalyugin:
I ran into bagful of troubles implementing Scenes myself, the main problem is lack of working examples or even detailed issues regarding the question. To help other people having similar issues, I'll just leave this tutorial here.

**What are Scenes and SceneWizard, what are their purpose?**
Basic API of telegraf allows for implementation of simple interactions, such as:
command -> response or message -> response, trigger -> response

While that is quite powerful and easy to use, there is a set of tasks, that require to create a DIALOGUE between user and the bot. That is where Scenes and SceneWizards fill in.


**What is a Scene, how does it work?**
Scene is like a namespace, an abstract isolated room that user can be pushed into via code, when inside one, all the outside noise such as global commands or message handlers stop reacting on user actions.
Scenes can be entered, or left from, scenes can have their own command/message listeners, that only work while user is inside.


**What kind of Scenes are there, what's the difference?**
There are two types of Scenes, a more configurable `BaseScene`, or a less configurable, Step-based `SceneWizard`.


**How to create and use a `BaseScene`**
`BaseScene` is just an isolated namespace that needs manually attached hooks, here is an example:
```
import { BaseScene } from 'telegraf';

const scenarioTypeScene = new BaseScene('SCENARIO_TYPE_SCENE_ID');

scenarioTypeScene.enter((ctx) => {
  ctx.session.myData = {};
  ctx.reply('What is your drug?', Markup.inlineKeyboard([
    Markup.callbackButton('Movie', MOVIE_ACTION),
    Markup.callbackButton('Theater', THEATER_ACTION),
  ]).extra());
});

scenarioTypeScene.action(THEATER_ACTION, (ctx) => {
  ctx.reply('You choose theater');
  ctx.session.myData.preferenceType = 'Theater';
  return ctx.scene.enter('SOME_OTHER_SCENE_ID'); // switch to some other scene
});

scenarioTypeScene.action(MOVIE_ACTION, (ctx) => {
  ctx.reply('You choose movie, your loss');
  ctx.session.myData.preferenceType = 'Movie';
[...]

PR #1194, opened 2020-11-27
Types and examples for scenes and wizards

### LINKED ISSUE #894: Type defenition of ctx.scene.state
I want to use state of scene to save data, but because it originally typed as object typescript 
does not allow this.

It would be great to be able to set the type of state. For example pass an iterface to Telegraf or Stage

### LINKED ISSUE #1165: Wrong typings in 3.38.0
You have wrong typings in the latest release, js files have methods that are not declared in d.ts. For example, there is setChatStickerSet method in telegram.js and no declaration for it in telegram.d.ts

### LINKED ISSUE #1172: Typescript building error
Code 
```
import { Telegraf, Context, Stage, BaseScene, Middleware } from "telegraf"
import { MiddlewareFn } from "telegraf/typings/composer"
import { TelegrafContext } from "telegraf/typings/context"
import { BOT_TOKEN } from "./consts"

export const StartBot = async() => {
  const greeter = new BaseScene("greeter")
  greeter.enter((ctx) => {
    console.log("enter scene")
    ctx.reply("Hi")
  })

  const stage = new Stage([greeter])
  stage.register(greeter)

  const _bot = new Telegraf(BOT_TOKEN)
  _bot.use(stage.middleware())
  _bot.start((ctx) => {
    // ctx.scene.enter("greeter")
    // ctx.reply("Start input your class name and select from autocomplite")
  })
  await _bot.launch()
}
```
When trying to build throwing errors:
```
src/bot.ts(17,12): error TS2345: Argument of type 'MiddlewareFn<SceneContextMessageUpdate>' is not assignable to parameter of type 'Middleware<TelegrafContext>'.
  Type 'MiddlewareFn<SceneContextMessageUpdate>' is not assignable to type 'MiddlewareFn<TelegrafContext>'.
    Property 'scene' is missing in type 'TelegrafContext' but required in type 'SceneContextMessageUpdate'.
src/bot.ts(19,9): error TS2339: Property 'scene' does not exist on type 'TelegrafContext'.
```

KnorpelSenf:
# Description

The types of Stages, Scenes, and Wizards were mostly wrong. The examples did not typecheck.

This PR applies consistent types to this part of the library. The example bots are migrated to TS, they build now with minimal changes.

While the types should be correct everywhere now, this comes with the downside of complexity. Scenes rely on the session. As a result, everything is bloated up to having two type variables: one for the session data and one for the context object, as both types can be adjusted. This trickles down far.

Note that the type variables follow the standard convention of single character names rather than the special `telegraf` version of prepending a `T`. We can change this back if desired, I simply found it even less readable than the current version to make all names longer.

Fixes #1172 
Fixes #1165
Fixes #894

Helps with
- #705
- #1074
- #1139

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update (need an example bot that shows how to customize context and session data)

# How Has This Been Tested?

```bash
npm t
npm run build
npm run lint
```

# Checklist:


wojpawlik (commented):
I don't see anything obviously very wrong, but I understand pretty much nothing here...

docs/examples/scenes-bot.ts
```diff
@@ -19,7 +25,7 @@ echoScene.command('back', leave())
 echoScene.on('text', (ctx) => ctx.reply(ctx.message.text))
 echoScene.on('message', (ctx) => ctx.reply('Only text messages please'))
 
-const bot = new Telegraf(process.env.BOT_TOKEN)
+const bot = new Telegraf<SceneContext>(process.env.BOT_TOKEN)
```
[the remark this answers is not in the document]
KnorpelSenf: It is necessary to extend the bot context in one way or another if scenes should be used.

The default `SceneContext` interface can be used (or `extend`ed) to automatically get the correct types for `ctx.scene` on the context object. If it is omitted, only the regular `Context` properties are available, and hence
```ts
bot.command('greeter', (ctx) => ctx.scene.enter('greeter'))
```
would fail.

KnorpelSenf:
> but I understand pretty much nothing here

Then check out what is going on if #1195 … no but seriously, do you think that I should write more comments?

wojpawlik:
I don't know, better ask someone familiar with Scenes/Stages/Wizards

KnorpelSenf:
> I don't know, better ask someone familiar with Scenes/Stages/Wizards

Who would that be?

I also had no idea about these things before I started working on this PR, so I'd appreciate it if someone with more experience than me could have a look.

KnorpelSenf:
@dotcypress `git blame` told me that you an expert on this domain—would you mind reviewing this PR?

src/scenes/context.ts
```diff
@@ -7,52 +7,76 @@ const debug = d('telegraf:scenes:context')
...
+  ) {
+    this.options = { sessionName: 'session', ...options }
+  }
 
   get session() {
     const sessionName = this.options.sessionName
-    let session = (this.ctx as any)[sessionName].__scenes ?? {}
-    if (session.expires < now()) {
-      session = {}
+    let session: S = (this.ctx as any)[sessionName].__scenes ?? {}
```
wojpawlik, line 46: 3 `this.ctx as any`s in this file have to go

wojpawlik (changes requested):

src/scenes/context.ts
KnorpelSenf: There is only three. How are you going to construct an object that has exactly one property which can be parametrized? I don't think TS allows us to do this, but I'm happy if you can prove me wrong

wojpawlik:
```ts
C extends Context & { session?: { __scenes: something } }
```

KnorpelSenf: That is wrong. You have to parametrize `session` with the `sessionName` argument of the constructor. Session data is not always stored in the same property on the context object.

KnorpelSenf: We would maybe need some magic with string literal types as type variable for the property name, mapped types to dynamically construct the object, `keyof` things to make sure everything matches, … and there is zero benefit to the end user as everything is typed already.

If this was possible, we would be able to infer automatically generated context objects from the user in many cases.

wojpawlik: I got read of that misfeature to simplify typings.

KnorpelSenf: Well have a go

KnorpelSenf: I mean, everything. Holy cow. How was I not aware of #1131?

I'll have to look into removing a large part of the type parameters everywhere (not just in this module). Takes some time.

KnorpelSenf: Turns out I could not remove the parameters as I wanted to. Either way, I also removed the option to specify `sessionName` for scenes and wizards, hence removing the casts to `any`.

KnorpelSenf:
@IMalyugin @wwwat @foodornt you all seem to be interested in having TypeScript types for Stages/Scenes.

The types are fixed now. However, I have never used that feature of `telegraf`, so I have quite little experience with scenes etc. Can you please pull in the code from this branch and verify that
- the types work as expected,
- the PR does not alter the behaviour of the feature, and
- you can customise the types with your own session data and context objects in order to adjust them to your needs?

This would be very helpful. We cannot merge this without other people's approval :)

Regarding the third point: I will write some examples about how to that. This is necessary anyway, but please don't hesitate to validate the first two things already!

KnorpelSenf:
Some type variables are redundant. It has to be possible to infer certain variables across `session`, `scene.session`, and `session.__scenes`. It is annoying to type the names multiple times, and also it does not make sense to specify them differently in different positions.

This needs to be figured out before a merge.

KnorpelSenf:
I added a lot of example bots. They illustrate every possible way to use the type variables. It's mostly redundant stuff, so it should be enough to check out the most basic one and the most advanced one for reviewing this PR. There is a lot of nesting of the same objects here. This makes it very convenient to use, so I don't think we should change it, but it also makes the types a bit annoying internally. They are pretty fine on the surface, though.

KnorpelSenf:
I figured out https://github.com/telegraf/telegraf/pull/1194#issuecomment-741228992.

The problem is that `ctx.scene.session` is an alias for `ctx.session.__scenes`. It is in fact possible to add type annotations to `ctx.scene: SceneContextScene` that infer this correctly, however, as soon as we extend scenes and add wizards, this logic breaks. It is not possible to keep up this alias without larger restructurings of the module, which I'd like to avoid.

As a consequence, whenever the scene session under `ctx.scene.session` is augmented, this type has to be specified twice, once in a place that is responsible for `ctx.scene.session` and once for `ctx.session.scenes`. I added a large number of examples that illustrate this.

It is difficult to explain to new users why type inference works correctly for scenes but not for wizards in this place. Hence, it makes sense to simply advise to consequently specify both type variables for `SceneContext` and `WizardContext` whenever the scene/wizard-specific session is extended. Nothing can go wrong here, it may just be redundant in very special cases that are too rare to care about.

wojpawlik:
Please review and test #1224, after it's merged you can merge `develop` here to fix the conflict and `// Enable graceful stop` to all the new examples. I'll review this in the meantime.

wojpawlik:
I'd still be happy to get review from someone more experienced with the stuff before I approve.

docs/examples/scenes/scenes-bot.ts
```diff
@@ -41,3 +41,7 @@ bot.launch()
 // Enable graceful stop
 process.once('SIGINT', () => bot.stop('SIGINT'))
 process.once('SIGTERM', () => bot.stop('SIGTERM'))
+
+// Enable graceful stop
+process.once('SIGINT', () => bot.stop('SIGINT'))
+process.once('SIGTERM', () => bot.stop('SIGTERM'))
```
wojpawlik, line 47: Duplicated
```suggestion
```

wojpawlik (changes requested):

KnorpelSenf:
For reviewing this, it helps a lot to install `ts-node` and run the example bots. You can
```
bot.use(async (c,n) => { console.log(c.session); await next(); console.log(c.session); })
```
to inspect that the scenes actually do their job.

The types are easy to check by verifying that the `scenes-bot-custom-context-and-session-and-scene-and-scene-session.ts` compiles, which it does.

wojpawlik (approved):
It'd be great if you sorted these out:
```
 /home/runner/work/telegraf/telegraf/src/scenes/context.ts
Warning:   101:10  warning  Unexpected nullable object value in conditional. An explicit null check is required  @typescript-eslint/strict-boolean-expressions
Warning:   120:10  warning  Unexpected nullable object value in conditional. An explicit null check is required  @typescript-eslint/strict-boolean-expressions

/home/runner/work/telegraf/telegraf/src/stage.ts
Warning:   27:12  warning  Unexpected string value in conditional. An explicit empty string check is required  @typescript-eslint/strict-boolean-expressions
```

Other than that, I don't see anything wrong here, but to be honest I understand almost nothing, so I just trust TypeScript (and you).

KnorpelSenf:
If no one else shows up and pops a question until the end of tomorrow, I'll merge.

I have tested example bots for both scenes and wizards thoroughly and they worked without problems. Writing the other example bots made me confident that the types are correct.

Even if we missed something here, the existing types will be an order of magnitude better than the current state, so the rest is fixing bugs (and I can do that).

PR #1195, opened 2020-11-27
Leverage unionization

KnorpelSenf:
# Description

`typegram`allows to narrow down certain types (most notably `Update` and `Message`) to its different subtypes. However, `telegraf` does not use this feature yet ([reason](https://github.com/telegraf/telegraf/commit/5f20e61a248a1a29f775a0f179085e0695329c72#commitcomment-44566050)).
We therefore have to check if `message.text` is present even in the case where we register the middleware only for `text` events and hence know in advance that it is always truthy.

This PR adds a number of new types to the library that allow for type inference on the context object in middleware functions. As a result, a context object is correctly assumed to provide a `message` property when it is the argument of a middleware function that is being passed to `bot.on('message')`. The same is true for subtypes of messages.

Furthermore, it is possible to provide a list of update types and message subtypes. The type inference will then guarantee that all of the relevant properties are available, but optional. Thus, we get autocompletion for the properties that make sense, while still having to check explicitly which of them is actually present. Even here it is possible to narrow down the type of update, i.e. when registering for `['callback_query', 'text']` and then checking if either of the update types is present, the other option will be discarded. (Callback queries and text messages are mutually exclusive in an update.)

Registering middleware on `forward` correctly assures that `forward_date` is present, even though the property `forward` will never exist on the message. In other words, the type mapping of `telegraf` that happens under the hood is regarded.

Even though this PR is good as it is, there is still one open question that we may want to discuss before merging. `telegraf` supports switching on [`channelMode`](https://telegraf.js.org/#/?id=constructor). When activated, all channel posts are handled as messages. **What happens in this case?** It is technically feasible to make `message` optional every time and also provide `channel_post` as an optional property. In the current state, `channel_post` is said to be unavailable when registering middleware for `text` while `message` is guaranteed to be present. This happens regardless of `channelMode` (which is obviously wrong). It may be possible to do this by adding another type variable to `Composer`.

## Type of change

Only changes type annotations. Might require elaborations on how to use this project with TypeScript, depending on how we decide on the aforementioned open question.

# How Has This Been Tested?

```
npm run build
npm run lint
npm t
```

# Checklist:

- [ ] I have made corresponding changes to the documentation (perhaps we need docs about this)
- [ ] I have added tests that prove my fix is effective or that my feature works (we need more examples or example bots, confer e.g. #1139)


wojpawlik (commented):
Impressive.

I'm not sure how this typechecks without updating corresponding static methods, but I expect inference to break when you do.

Minor thing: ideally, for example in `composer.on('message', ctx => ...)`, `ctx.editedMessage` and others should be `undefined` (currently they're left as optional).

> `telegraf` supports switching on [`channelMode`](https://telegraf.js.org/#/?id=constructor). [...] It is technically feasible to make `message` optional every time and also provide `channel_post` as an optional property.

That's what I did in my failed attempt.

src/context.ts
```diff
@@ -64,11 +75,44 @@ const MessageSubTypes = [
   'poll',
   'forward_date',
 ] as const
+export type MessageProps = {
+  [key in MessageSubTypesUnionMapped]: {
+    message: {
+      [k in MapSubTypeBack<key>]: UnionToIntersection<tt.Message>[k]
+    }
+  }
+}
```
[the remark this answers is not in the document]
KnorpelSenf: Don't bother reviewing this further. First, I must work around the problem of having union types that are too large to be represented. It is possible to manually pre-compute some types in some critical inference steps. This cuts down the compile time significantly, too.

The above type is one of the magical parts that I abandoned. It is correct but too slow. I can still explain it, though, it is crucial that I'm not the only one who knows what's going on :wink:

The general idea is straightforward: instead of having middleware for `TContext extends Context`, we intersect the context subclass with a type that requires the properties that we assume to be present. Here is a concrete example of this.

```ts
bot.on('message', mw)
```

We can now infer that `mw` has to be middleware of type `Middleware<Context & { message: tt.Message }>`. This makes sure that the `message` property is given and has the correct type. (Note that `Middleware<Context & { message: object }>` would also be enough because the intersection of `Context['message'] & object` simply boils down to `Context['message']`.)

Thinking this through, we therefore have to map the `'message'` in the first argument of `on` to this type. As we always intersect with `TContext extends Context` independent of `'message'`, we only need to obtain the type `{ message: tt.Message }` from `'message'`.

There are three types alike to the one you quoted, one of them does exactly this.

Regarding subtypes of messages, we simply need to take everything one step further and also intersect with the message property in order to assure the presence of the corresponding properties during type inference. The type you quoted does exactly that: it takes a `tt.Message` and picks out the respective property, hence creating an object that requires it. This resulting object can be intersected with `TContext extends Context` such that we obtain the correct type for `mw`.

The mapping of types with `MessageSubTypesUnionMapped` and `MapSubTypeBack` solves that problem that some message subtypes have different names in the event name than in the property name. This mapping is specified in `MessageSubTypesMapping`, i.e. `forward` is used to register middleware for message forwards, but `forward_date` is actually the property that we have to require on `TContext extends Context`. We derive a reverse mapping of the `MessageSubTypesMapping` by inventing a type that swaps keys and values in an object. This can then be used during type inference.

As a said, the PR works fine and I'm halfway through annotating the method signatures, but these comprehensive union types are simply a too large thing to rely on in such advanced use cases. I'll ping you once I have something that is both correct _and_ fast enough to the pleasant to use.

KnorpelSenf:
> I'm not sure how this typechecks without updating corresponding static methods, but I expect inference to break when you do.

Updating these methods is both required and not a problem.

Note that there is one difference to a previous approach: it is not possible to infer from previous middleware arguments the types of the next. This is available for `compose`, but other methods use compositional inference rather than inferring the types from the usage. This solves the problem we had earlier and it works very well.

> Minor thing: ideally, for example in `composer.on('message', ctx => ...)`, `ctx.editedMessage` and others should be `undefined` (currently they're left as optional).

Good idea! I would not have thought about it. I added that now.

KnorpelSenf:
I do have to add that we should write more documentation about how to use `telegraf` with TypeScript. Given the size of this PR, however, it might make sense to do this in the next iteration.

@wojpawlik feel free to play around with the example bots, the types are working well there IMO. Looking forward to your hearing your opinion about it.

src/composer.ts
```diff
@@ -34,6 +42,38 @@ function getText(
...
+  [key in tt.MessageSubType]: Props['message'] &
+    tt.UpdateSubProps[key] &
+    tt.ContextSubProps[key]
+}
+
+type MatchedContext<
+  C extends Context,
+  T extends tt.UpdateType | tt.MessageSubType
+> = C &
+  // TODO: support subtypes for: callback query, channel post, maybe more?
```
KnorpelSenf, line 58: Do we want to support type inference in even more cases, not just for subtypes of `message`?

KnorpelSenf: Currently, when using a `MessageSubType`, we infer subtypes for the `message` property only, even though it could come from a different update type.

Take this code:
```ts
bot.on('text', ctx => {
  const text = ctx.message.text // works
})
```
We correctly infer `ctx.message.text: string` as well as `ctx.update.message.text: string`.

However, `mount` not only checks message texts, but also the content of all update types, including callback queries, inline queries, channel posts, etc.

Take this code:
bot.on('channel_chat_created', ctx => {
  const chat = ctx.channelPost.chat // error
})
We infer both `ctx.message.channel_chat_created` and `ctx.update.message.channel_chat_created` as `true` and `ctx.channelPost` as `undefined`, all of which is incorrect.

I'm suggesting that we should also do a similar thing for other subtypes that we provide. It may make sense to do this generically, but I believe that this will create too complex types again, so we may have to come up with something smarter.

Note: the TODO comment in the code is not quite correct about callback queries as neither `'data'` nor `'game_short_name`' are valid message subtypes, so callback queries should in fact not be the issue here. It's about channel posts, edits, and channel edits.

KnorpelSenf: Now that I think about it, it basically boils down to supporting channel mode also in this place.

wojpawlik:
Tbh., I'd suggest to remove `channelMode` and replace subtypes to be like `edited_message:text` in next PR.

KnorpelSenf:
> Tbh., I'd suggest to remove `channelMode` and replace subtypes to be like `edited_message:text` in next PR.

Sounds like a perfect fit for template literal types.

I like the idea actually. This should make it much easier to and explain. I remember well how confused I was in the beginning that I could do both `on('message')` and `on('text')`and it meant very different things.

I suggest in turn that we leave this PR as it is, get a review done and then open a third PR that does the rest.

wojpawlik:
Actually, the `edited_message:text` format I suggested doesn't allow to express more complex queries like "forwarded text message". Ideally, in 5.0 `optional` could accept a type guard, and then `optional(is.message(is.forwarded, is.text), handler)`. Not sure what's the best for now tho.

KnorpelSenf:
I see the point that reworking how to listen to update types should be done thoroughly. This opens two questions:

1. Given that we won't change the behaviour of `bot.on` for now, how are we going to go about `channelMode` then? I looked into adding types for it, but that would require a second or third type variable for all methods in `Composer`. I have no idea whether TypeScript could handle this (think of all the unions), and I won't try this before we deem it necessary.

1. Do we have a roadmap to what needs to be done before 4.0? I feel like the v4 milestone is updated whenever we feel like it, flexibly adding and removing things as we like. I'd like to somehow put on record which parts we want to include in 4.0 and which parts we postpone.

wojpawlik:
I put every breaking change that comes up into the milestone.

KnorpelSenf:
One possible solution here is to exclude `channelPost` from being inferred as `undefined`. That way, users of `channelMode` retain access to the property, while others probably won't care.

After applying this change, the types are still a bit incorrect for `channelMode = true` in the way that `message` should in fact be optional while it is inferred as required, and they are a bit incorrect for `channelMode = false` in the way that `channelPost` should be `undefined` while it is inferred as `undefined | tt.Message`. Either way, we're getting pretty close to being completely correct, while saving us a lot of work. This PR would be feature complete (after this small change).

src/composer.ts
```diff
@@ -338,114 +439,122 @@ export class Composer<TContext extends Context>
...
-    entityType: string,
-    predicate: Triggers<TContext>,
-    ...fns: NonemptyReadonlyArray<
-      Middleware<TContext & { match: RegExpExecArray }>
-    >
-  ): Middleware<TContext> {
+  static entityText<TContext extends Context, P extends Triggers<TContext>>(
+    entityType: MaybeArray<string>,
+    predicate: P,
+    ...fns: EntityMatch<TContext, P>
```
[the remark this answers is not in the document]
KnorpelSenf: You're right. So basically, parametrizing P is nonsense. `ctx.match` is always of type `RegExp`. Correct?

src/composer.ts
```diff
@@ -183,13 +244,15 @@ export class Composer<TContext extends Context>
     return this.handler
   }
 
-  static reply(...args: Parameters<Context['reply']>) {
-    return (ctx: Context) => ctx.reply(...args)
+  static reply<TContext extends Context>(
+    ...args: Parameters<Context['reply']>
+  ): Middleware.Fn<TContext> {
+    return (ctx: TContext) => ctx.reply(...args)
```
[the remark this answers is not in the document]
KnorpelSenf: Redundant yes, but not doing it is inconsistent with the rest of the methods. Is there a disadvantage of being more explicit in this case?

src/composer.ts
```diff
@@ -34,6 +34,38 @@ function getText(
...
+
+type EntityMatch<
+  C extends Context,
+  M = RegExpExecArray,
+  T extends tt.UpdateType | tt.MessageSubType = 'message' | 'channel_post'
+> = NonemptyReadonlyArray<Middleware<MatchedContext<C & { match: M }, T>>>
+
+type Prefix<S extends MaybeArray<string>, P extends string> = S extends string
+  ? `${P}${S}`
+  : `${P}${S[number]}`
```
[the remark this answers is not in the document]
KnorpelSenf: Given https://github.com/telegraf/telegraf/pull/1195#discussion_r533631269, we can remove it entirely.

src/composer.ts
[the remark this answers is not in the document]
wojpawlik: \* `RegExpExecArray`

src/composer.ts
```diff
@@ -34,6 +34,38 @@ function getText(
...
+  T extends tt.UpdateType | tt.MessageSubType
+> = C &
+  // TODO: support subtypes for: callback query, channel post, maybe more?
+  (T extends tt.UpdateType ? Props[T] : SubProps[Exclude<T, tt.UpdateType>]) &
+  tt.AbsentProps<
+    Exclude<
+      tt.UpdateType,
+      T | (T extends tt.MessageSubType ? 'message' : never)
+    >
+  >
```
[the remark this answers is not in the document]
KnorpelSenf: I have no idea where you're going with your plan, so I can't prepare it for you to use. But I can definitely make it more modular.

KnorpelSenf:
> For correctness, both `message` and `channelPost` should be left optional

That would also be wrong. Consider this:

```ts
// channel mode if off (default)
bot.on('text', ctx => {
  const t = ctx.message.text // does not compile with your suggestion, but it should
})
```

That's why I think this weird middle ground is a good compromise.

```ts
// channel mode OFF
bot.on('text', ctx => {
  const m = ctx.message // inferred as message, which is correct
  const c = ctx.channelPost // inferred as message or undefined, while it should in fact be undefined
})
// channel mode ON
bot.on('text', ctx => {
  const t = ctx.message.text // inferred as message, while it should in fact be message or undefined
  const c = ctx.channelPost // inferred as message or undefined, which is correct
})
```

Those two corner cases are of course not 100 % nice. I get that. But I'm pretty sure we can assume that very few people actually use channel mode. Not having `message.text` available when explicitly listening for message text is really odd and much more common. And it's not just text, it basically means that any bot that handles messages of any kind always has to check for every update whether the message is actually present.

src/composer.ts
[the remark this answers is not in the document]
wojpawlik: `passThru` returns `Middleware.Fn<Context>`, I'd also be fine with that here.
```suggestion
  static reply(...args: Parameters<Context['reply']>): Middleware.Fn<Context> {
    return (ctx) => ctx.reply(...args)
```

src/composer.ts
```diff
@@ -34,6 +34,35 @@ function getText(
...
+
+type EntityMatch<
```
wojpawlik, line 59: Confusing name, sounds like it's related to `tt.MessageEntity`. Best I can come up with is `MatchedMiddlewares`, but I think just inlining the type would be the clearest.

wojpawlik (changes requested):

src/composer.ts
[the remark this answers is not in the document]
wojpawlik: I'm going for something like https://github.com/telegraf/telegraf/pull/1195#issuecomment-735779847

src/composer.ts
KnorpelSenf: Well it's called like [that](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L173) [because](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L178) [it](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L184) [has](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L191) [everything](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L198) [to](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L205) [do](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L212) [with](https://github.com/telegraf/telegraf/blob/3d8d132e658e7bfb98ae8a269b76faa39fc2dac9/src/composer.ts#L219) `tt.MessageEntity`. If taking in regular expressions makes the name less fitting, we can inline it in that one case. I don't see the need for repeating the `match` property countless times, that's a bit too WET

src/composer.ts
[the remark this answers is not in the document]
KnorpelSenf: I know that you mentioned this, but that doesn't explain it. For instance, I don't understand yet where `optional` and `is` are coming from. Is your code part of an example bot, located in `composer.ts`, or in `context.ts`? Is it internal? Should it be imported? What are even the types of it? How does it solve the problem of listening to forwards of both `message`s and `channelPost`s if the update type to be specified again?

Your comment opens many more questions than it answers and I don't think it makes sense to speculate about half a line of code, so I'll just modularize it independently of that

src/composer.ts
wojpawlik: Objection! You used `EntityMatch` in `hears` as well, which has nothing to do with `tt.MessageEntity`.

KnorpelSenf: That's what I was referring to with regular expressions

KnorpelSenf: As in, we can inline it for `hears` only.

Or, of course, find a better name.

src/composer.ts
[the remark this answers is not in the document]
KnorpelSenf: Ahm. True. That's in fact what I committed :sweat_smile:

src/composer.ts
```diff
@@ -34,6 +34,14 @@ function getText(
   return undefined
 }
 
+type MatchedContext<
```
KnorpelSenf, line 37: Does this help you in any way?

If you were thinking of having it very differently, feel free to change it as you like

wojpawlik: Nope. I need a type with params `<C extends Context, U extends tt.Update>`, independent of `tt.UpdateType | tt.MessageSubType`.

> feel free to change it as you like

I can't. I don't understand your types. They're complex, and there's so many of them. I'd need to ignore or scrap them and start from scratch. Hence why I'm asking you to give me some entry point I can immediately use, and which would perhaps allow me to understand and simplify rest of your types.

src/composer.ts
wojpawlik: Anything wrong with `MatchedMiddlewares`? Except, of course, library users having no clue what it is, but that'd be a problem with every type alias.

src/composer.ts
KnorpelSenf: > I can't. I don't understand your types.

I see. Did you see https://github.com/telegraf/telegraf/pull/1195#discussion_r532059517? Don't hesitate to ask, I am aware that what I did is not straightforward at all

wojpawlik: I did. But there's so much stuff that I don't know where to even begin to look!

In my opinion, only 3 types are needed: `MountMap`, which maps `tt.UpdateType | tt.MessageSubType` into a `TUpdate`, `GuardedContexts` which intersects `TContext` with a type derieved from `TUpdate`, and one helper.

KnorpelSenf: > I don't know where to even begin to look

`MatchedContext` is where everything begins. I wrote some docs now.

It takes a context and makes those things required that should be required. It makes those things undefined that should be undefined.

It does this by doing an intersection with some objects that have very specific properties set.

KnorpelSenf: Here is an illustration of this. I hope it addresses the confusion.

```
type Context = {
  message: {
    text: string | undefined
    left_chat_member: true | undefined
    other_prop: symbol // or whatever
  }
}

type GuaranteedStuff = {
    message: {
      text: string
      // we ignore all the rest here …
  }
}
type UndefinedStuff = {
    message: {
      left_chat_member: undefined
    }
}

// what we actually want to have in the end!
type MatchedContext = Context & GuaranteedStuff & UndefinedStuff
```

wojpawlik: Thanks, this helps a lot! I wish types you added to `telegram-types.ts` received the same treatment.

KnorpelSenf: Your wish is my command.

KnorpelSenf: Looking at your suggestion for `guard`, I think I see what you were expecting from the types I wrote, and why this did not match. You needed some sort of type guarding logic for different update types (`update is TUpdate`), while I implemented a type transformation that narrows down some unions deep down in a context object. Those approached look fundamentally different to me. Maybe it's possible to somehow make one use the other, but I don't really know how.

wojpawlik (commented):
Last question, what do you have against `Parameters`? 😄

KnorpelSenf:
Nothing really. It makes things less readable and in some cases more maintainable. I try to avoid them for complex type signatures. If I do advanced stuff, I don't want things to be more implicit than necessary. Why?

wojpawlik:
DRY?

KnorpelSenf:
Yes. That's what I mean with more maintainable.

It's useful and I like it for what it can do, but it comes at a cost and sometimes I think it's better not to pay that.

KnorpelSenf:
But does the PR work? Do the voodoo magic types do their job? I'm looking forward to your `optional` thingy 😄

src/context.ts
```diff
@@ -107,12 +97,12 @@ export class Context {
 
   get message() {
     if (!('message' in this.update)) return undefined
-    return deunionize(this.update.message)
+    return this.update.message
```
septs, line 100: why not use optional chaining.

```ts
return this.update?.message
```

see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

src/composer.ts
```diff
@@ -276,60 +333,91 @@ export class Composer<TContext extends Context>
...
+  ): Middleware.Fn<TContext> {
     return Composer.branch(predicate, anoop, Composer.passThru())
   }
 
   static dispatch<
     TContext extends Context,
     Handlers extends Record<string | number | symbol, Middleware<TContext>>
-  >(routeFn: (ctx: TContext) => keyof Handlers, handlers: Handlers) {
+  >(
+    routeFn: keyof Handlers | ((ctx: TContext) => keyof Handlers),
```
wojpawlik, line 360: I wanna drop this behavior, not document it. Users can easily enough `.use(handlers[handlerKey])`.

src/composer.ts
```diff
@@ -483,26 +578,27 @@ export class Composer<TContext extends Context>
...
     if (fns.length === 0) {
       // @ts-expect-error
       return Composer.entity(['bot_command'], command)
     }
     const commands = normalizeTextArguments(command, '/')
-    return Composer.mount(
+    return Composer.mount<TContext, 'text'>(
       'text',
-      Composer.lazy<TContext>((ctx) => {
+      Composer.lazy<TContext, MatchedContext<TContext, 'text'>>((ctx) => {
```
wojpawlik, line 590: Is the 2nd type param necessary in `Composer.lazy`? It seems to work without.
```suggestion
      Composer.lazy<MatchedContext<TContext, 'text'>>((ctx) => {
```

wojpawlik (changes requested):

wojpawlik:
To avoid merge conflicts, I'll continue working on `Composer::guard` only after this is merged.

wojpawlik (commented):
Don't worry too much about other people verifying the new types, I'll try to simplify them right after this is merged :)

In the meantime, I'll work on type guard functions to PR along with `Composer::guard` after this is merged.

src/context.ts
KnorpelSenf: We have to narrow down `Update` before we can access message, as it is not always present on an update.

Otherwise, we get
```
Property 'message' does not exist on type 'Update'.
  Property 'message' does not exist on type 'CallbackQueryUpdate'.ts(2339)
```

src/composer.ts
KnorpelSenf: Should we just remove it entirely then? After all, there's no better point in time than a new major version to drop legacy code.

src/composer.ts
KnorpelSenf: Not for me.
```
Argument of type 'Fn<TContext>' is not assignable to parameter of type 'Middleware<MatchedContext<TContext, "text">>'.
  Type 'Fn<TContext>' is not assignable to type 'Fn<MatchedContext<TContext, "text">>'.
    Type 'MatchedContext<TContext, "text">' is not assignable to type 'TContext'.
      'MatchedContext<TContext, "text">' is assignable to the constraint of type 'TContext', but 'TContext' could be instantiated with a different subtype of constraint 'Context'.ts(2345)
```

src/composer.ts
```diff
@@ -53,51 +99,51 @@ export class Composer<TContext extends Context>
   /**
    * Registers middleware for handling provided update types.
    */
-  on(
-    updateTypes: MaybeArray<tt.UpdateType | tt.MessageSubTypes>,
-    ...fns: NonemptyReadonlyArray<Middleware<TContext>>
+  on<U extends tt.UpdateType | tt.MessageSubType>(
+    updateType: MaybeArray<U>,
+    ...fns: NonemptyReadonlyArray<Middleware<MatchedContext<TContext, U>>>
```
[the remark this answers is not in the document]
KnorpelSenf: I'd love that! It's not really in the scope of this PR, though, is it? `router.ts`, `types.ts`, and everything in the scenes directory are so far unaffected by this PR, so it may make sense to do a refactoring PR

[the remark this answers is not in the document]
wojpawlik: > It's not really in the scope of this PR, though, is it?

This PR already does way more than leveraging unionization, so it's fine.

`scenes/**` are already changed in #1194, the other files can be changed while doing some other changes to them, don't touch them here.

src/composer.ts
wojpawlik: I just did `npm ci` and have no issue with the change, I'll push.

src/composer.ts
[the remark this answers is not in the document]
KnorpelSenf: > This PR already does way more than leveraging unionization

What would that be?

I mean, I see how it is convenient to do the renaming here given that the code is modified anyway. But what changes are so far not required when we want to make use of unionization?

src/composer.ts
wojpawlik: Typechecking passed on CI. Now only tests of legacy `Composer.dispatch` need to be removed.

src/composer.ts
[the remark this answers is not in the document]
wojpawlik: `MatchedMiddlewaware`, removal of `Parameters`, explicit return type annotations...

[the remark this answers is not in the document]
wojpawlik: > don't touch them here

...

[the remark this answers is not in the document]
wojpawlik: I'll do a partial revert, hang on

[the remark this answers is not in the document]
wojpawlik: Sorry for not being clear enough

wojpawlik (approved):

src/composer.ts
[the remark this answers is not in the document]
KnorpelSenf: > `MatchedMiddlewaware`

necessary to do this while staying DRY

> removal of `Parameters`

necessary because the type variable now occurs in the function arguments, and we cannot do something like `typeof Composer[<C>'url']`, so we have to reintroduce the type variable to have access to it

> explicit return type annotations

necessary because it makes type inference much faster (I was basically just looking for places where we could precompute types as much as possible)

wojpawlik:
Anything else, or can I merge?

test/composer.js
```diff
@@ -356,7 +356,7 @@ test.cb('Composer.lazy should support middlewares', (t) => {
 })
 
 test.cb('Composer.dispatch should work with handlers array', (t) => {
-const bot = createBot()
+  const bot = createBot()
```
KnorpelSenf, line 359: I assume you had no _intent to indent_ (pun indented)

```suggestion
const bot = createBot()
```

KnorpelSenf:
> Anything else, or can I merge?

You can merge.

wojpawlik:
Now it's my time to shine

KnorpelSenf:
Looking forward to it! Also, you've already been shining pretty bright so far with very good reviews.

Issue #798, opened 2019-11-12
No botInfo in context if bot not started via launch()


Niklv:
## Context

In express.js example:

```js
const Telegraf = require('telegraf')
const express = require('express')
const expressApp = express()

const bot = new Telegraf(process.env.BOT_TOKEN)
expressApp.use(bot.webhookCallback('/secret-path'))
bot.telegram.setWebhook('https://server.tld:8443/secret-path')

expressApp.get('/', (req, res) => {
  res.send('Hello World!')
})

expressApp.listen(3000, () => {
  console.log('Example app listening on port 3000!')
})
```

There are not explicitly called `bot.launch(...)` in this example.
So `botInfo` in context will be `undefined`.

https://github.com/telegraf/telegraf/blob/develop/telegraf.js#L103-L105

* Telegraf.js Version: 3.33.3
* Node.js Version: 12.12.0
* Operating System: macos

## Expected Behavior

Similar behaviour in all launch configs. 

## Current Behavior

`botInfo` exists or not exists in `context` depending launch config.

dotcypress:
@Niklv this is not a bug, because `Telegraf` doesn't get bot by default. In some use cases we can't call Telegram at bot start(multi node configuration with webhook only)

PR #1200, opened 2020-11-29
Make `ctx.botInfo` required

### LINKED ISSUE #1072: Make `ctx.botInfo` required in 4.0
1. Make `Telegraf::startPolling` and `Telegraf::startWebhook` `private`, favor `Telegraf::launch`.
2. Return `Promise<(req: IncomingMessage, res: ServerResponse) => void>` from `Telegraf::webhookCallback` -- probably unpleasant to deal with without top-level await...

KnorpelSenf:
# Description

Makes sure that `ctx.botInfo` is always initialized. Makes that property required, too.

Setting `botInfo` directly allows for the feature to be disabled. [Reason](https://github.com/telegraf/telegraf/issues/798#issuecomment-554002376).

Fixes #1072
Impacts #798

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# How Has This Been Tested?

```bash
npm run build
npm test
npm run lint
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -70,7 +70,6 @@ const MessageSubTypesMapping = {
 }
 
 export class Context {
-  public botInfo?: tt.UserFromGetMe
   readonly updateType: tt.UpdateType
   readonly updateSubTypes: ReadonlyArray<typeof MessageSubTypes[number]>
   readonly state: Record<string | symbol, any> = {}
@@ -78,6 +77,7 @@ export class Context {
   constructor(
     readonly update: tt.Update,
     readonly tg: Telegram,
+    public readonly botInfo: tt.UserFromGetMe,
     private readonly options: { channelMode?: boolean; username?: string } = {}
   ) {
     this.updateType = UpdateTypes.find((key) => key in this.update)!
--- a/src/context.ts
+++ b/src/context.ts
@@ -78,7 +78,7 @@ export class Context {
     readonly update: tt.Update,
     readonly tg: Telegram,
     public readonly botInfo: tt.UserFromGetMe,
-    private readonly options: { channelMode?: boolean; username?: string } = {}
+    private readonly options: { channelMode?: boolean } = {}
   ) {
     this.updateType = UpdateTypes.find((key) => key in this.update)!
     // prettier-ignore
@@ -98,7 +98,7 @@ export class Context {
   }
 
   get me() {
-    return this.options.username
+    return this.botInfo?.username
   }
 
   get telegram() {
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -38,7 +38,6 @@ namespace Telegraf {
     handlerTimeout: number
     retryAfter: number
     telegram: Partial<ApiClient.Options>
-    username?: string
   }
 
   export interface LaunchOptions {
@@ -78,6 +77,7 @@ export class Telegraf<
 > extends Composer<TContext> {
   private readonly options: Telegraf.Options<TContext>
   private webhookServer?: http.Server | https.Server
+  private botInfo!: tt.UserFromGetMe
   public telegram: Telegram
   readonly context: Partial<TContext> = {}
   private readonly polling = {
@@ -91,7 +91,7 @@ export class Telegraf<
 
   private handleError: (err: any, ctx: TContext) => void = (err) => {
     console.error()
-    console.error((err.stack || err.toString()).replace(/^/gm, '  '))
+    console.error((err.stack ?? err.toString()).replace(/^/gm, '  '))
     console.error()
     throw err
   }
@@ -116,17 +116,26 @@ export class Telegraf<
 
   get webhookReply() {
     return this.telegram.webhookReply
-  } /* eslint brace-style: 0 */
+  }
 
   catch(handler: (err: any, ctx: TContext) => void) {
     this.handleError = handler
     return this
   }
 
-  webhookCallback(path = '/') {
+  webhookCallback(path = '/', skipBotInfoInitialization = false) {
     return generateCallback(
       path,
-      (update: tt.Update, res: any) => this.handleUpdate(update, res),
+      async (update: tt.Update, res: any) => {
+        if (!skipBotInfoInitialization) {
+          try {
+            this.botInfo ??= await this.telegram.getMe()
+          } finally {
+            skipBotInfoInitialization = true
+          }
+        }
+        return await this.handleUpdate(update, res)
+      },
       debug
     )
   }
@@ -171,10 +180,8 @@ export class Telegraf<
 
   async launch(config: Telegraf.LaunchOptions = {}) {
     debug('Connecting to Telegram')
-    const botInfo = await this.telegram.getMe()
-    debug(`Launching @${botInfo.username}`)
-    this.options.username = botInfo.username
-    this.context.botInfo = botInfo
+    this.botInfo = await this.telegram.getMe()
+    debug(`Launching @${this.botInfo.username}`)
     if (!config.webhook) {
       const { timeout, limit, allowedUpdates, stopCallback } =
         config.polling ?? {}
@@ -239,7 +246,7 @@ export class Telegraf<
     debug('Processing update', update.update_id)
     const tg = new Telegram(this.token, this.telegram.options, webhookResponse)
     const TelegrafContext = this.options.contextType
-    const ctx = new TelegrafContext(update, tg, this.options)
+    const ctx = new TelegrafContext(update, tg, this.botInfo, this.options)
     Object.assign(ctx, this.context)
     try {
       await this.middleware()(ctx, anoop)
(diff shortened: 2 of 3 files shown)


wojpawlik (changes requested):
> Introduced a new option to `webhookCallback` that allows for the feature to be disabled.

`skipBotInfoInitialization` defeats the point! It keeps `botInfo` effectively optional. If library users want to skip the `getMe` request, they should simply manually set `Telegraf::botInfo` before calling `launch` or `webhookCallback`.

One more thing I'm not sure how to address, if another request / update arrives before the `getMe` call finishes, Telegraf will call `getMe` again.

src/context.ts
```diff
@@ -70,14 +70,14 @@ const MessageSubTypesMapping = {
...
-  public botInfo?: tt.UserFromGetMe
   readonly updateType: tt.UpdateType
   readonly updateSubTypes: ReadonlyArray<typeof MessageSubTypes[number]>
   readonly state: Record<string | symbol, any> = {}
 
   constructor(
     readonly update: tt.Update,
     readonly tg: Telegram,
+    public readonly botInfo: tt.UserFromGetMe,
     private readonly options: { channelMode?: boolean; username?: string } = {}
```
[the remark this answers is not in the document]
KnorpelSenf: I know, that's why I removed it?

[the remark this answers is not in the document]
KnorpelSenf: Ah, I see that now. There, too. Got it.

src/telegraf.ts
```diff
@@ -116,17 +117,28 @@ export class Telegraf<
...
     return generateCallback(
       path,
-      (update: tt.Update, res: any) => this.handleUpdate(update, res),
+      async (update: tt.Update, res: http.ServerResponse) => {
+        if (mustFetchBotInfo) {
+          mustFetchBotInfo = false
+          try {
+            this.botInfo ??= await this.telegram.getMe()
+          } catch (err) {
+            debug('Could not initialize bot info', err)
```
wojpawlik, line 137: `handleUpdate` will run, without `botInfo`, if you don't rethrow

```suggestion
            debug('Could not initialize `botInfo`', err)
            throw err
```

src/telegraf.ts
```diff
@@ -116,17 +117,28 @@ export class Telegraf<
...
   }
 
   webhookCallback(path = '/') {
+    let mustFetchBotInfo = !this.botInfo
     return generateCallback(
       path,
-      (update: tt.Update, res: any) => this.handleUpdate(update, res),
+      async (update: tt.Update, res: http.ServerResponse) => {
+        if (mustFetchBotInfo) {
+          mustFetchBotInfo = false
```
wojpawlik, line 133: Not good as well. Now subsequent updates would be processed immiediately, without waiting for `botInfo`. They should ideally await the same promise.

src/telegraf.ts
```diff
@@ -78,6 +77,8 @@ export class Telegraf<
 > extends Composer<TContext> {
   private readonly options: Telegraf.Options<TContext>
   private webhookServer?: http.Server | https.Server
+  /** Set manually to avoid implicit `getMe` call in `launch` or `webhookCallback` */
+  public botInfo!: tt.UserFromGetMe
```
wojpawlik, line 81: How about simply `throw`ing in `handleUpdate` if `botInfo` is not set?

```suggestion
  public botInfo?: tt.UserFromGetMe
```

wojpawlik (changes requested):

src/context.ts
```diff
@@ -98,7 +98,7 @@ export class Context {
   }
 
   get me() {
-    return this.options.username
+    return this.botInfo?.username
```
wojpawlik, line 101:
```suggestion
    return this.botInfo.username
```

wojpawlik (changes requested):

src/telegraf.ts
KnorpelSenf: The whole point of this PR is to make botInfo required. Otherwise, library consumers have to check for `botInfo` being absent and request it themselves, which is inconvenient.

Also, none of our tests sets `botInfo`, so we'd have to fix all of those now …

src/context.ts
KnorpelSenf: I agree it would be correct. It is what I wrote at first, too. But it breaks the tests because we never set the `botInfo` there, and this was the easiest solution, though QND.

src/telegraf.ts
wojpawlik: `Context::botInfo` should be required, not necessarily `Telegraf::botInfo`.

src/telegraf.ts
wojpawlik: Actually, I wouldn't mind just a naive `this.botInfo ??= await this.telegram.getMe()` in this PR, it can be improved upon in the future if it ever causes any issues.

src/context.ts
wojpawlik: Actually type is correct, so whatever

src/telegraf.ts
KnorpelSenf: How are you going to initialize `Context` from `Telegraf` if `Telegraf` might not have access to `botInfo`?

src/context.ts
KnorpelSenf: Yes, exactly my thinking :smile:

src/telegraf.ts
KnorpelSenf: It's not too difficult to make all requests await the same promise at this point. Check out 6dc7745

src/telegraf.ts
KnorpelSenf: This would fail if `Telegraf::botInfo` is optional:

https://github.com/telegraf/telegraf/pull/1200/files#diff-407ce1c4de4ad023ddb3930497b90b8c0ca458d9a6096cea80abe06692d250c9R256

src/telegraf.ts
```diff
@@ -116,17 +117,32 @@ export class Telegraf<
...
 
   webhookCallback(path = '/') {
+    let mustFetchBotInfo = !this.botInfo
+    let botInfoCall: Promise<void> | undefined
     return generateCallback(
       path,
-      (update: tt.Update, res: any) => this.handleUpdate(update, res),
+      async (update: tt.Update, res: http.ServerResponse) => {
+        if (mustFetchBotInfo) {
+          mustFetchBotInfo = false
```
wojpawlik, line 134:
```suggestion
        if (this.botInfo === undefined) {
```

src/telegraf.ts
KnorpelSenf: I believe that all incoming requests would throw the same error now, it `getMe` fails for some reason. I'm not sure if this should be designed differently. I don't really like it, but I see no other way to make all waiting requests fail other than by making them throw.

wojpawlik (changes requested):

src/telegraf.ts
wojpawlik: Not if you throw in `handleUpdate` if it's not set

wojpawlik: Mind if I try?

KnorpelSenf: It would compile, that it true. All respective unit tests would fail, though. Feel free to set `botInfo` everywhere manually, I don't think its worth the effort.

Why is it even that you want to allow setting `botInfo` to `undefined` if every new request would try initialize it again anyway?

KnorpelSenf: > Mind if I try?

Of course not!

wojpawlik: > All respective unit tests would fail, though.

Oh, I see. Then it's fine as-is.

src/telegraf.ts
```diff
@@ -116,17 +117,30 @@ export class Telegraf<
...
+        if (!this.botInfo) {
+          await (botInfoCall ??= (async () => {
+            try {
+              this.botInfo ??= await this.telegram.getMe()
+            } catch (err) {
+              debug('Could not initialize `botInfo`', err)
+              throw err
+            }
+          })())
+        }
```
wojpawlik, line 141: Could be significantly simplified if you're willing to drop the `debug` call:

```suggestion
        this.botInfo ??= await (botInfoCall ??= this.telegram.getMe())
```

wojpawlik (approved):

src/telegraf.ts
KnorpelSenf: Yeah I think the error message should be enough. Thanks for the hint :+1:

wojpawlik:
Now types are wrong! If somebody follows https://github.com/telegraf/telegraf/blob/1f7dee72c702bdf05791ccba2a8a8c3b3c6f9dfc/docs/examples/koa-webhook-bot.js#L18-L23 `ctx.botInfo` will be missing (which should never happen according to types) and trying to use it will throw!

Throwing in `handleUpdate` if `botInfo` is the obvious solution, but I think we can do better. How about moving the code to `handleUpdate`? I mean, if `botInfo` is missing in `handleUpdate`, trying to call `getMe`.

KnorpelSenf:
Can reproduce it.

Yeah that makes a lot of sense, I like that. I didn't really consider that one could call `handleUpdate` directly. Do you want to submit the fix or should I?

wojpawlik:
Feel free. I have lots of stuff to review and reply to.

PR #1208, opened 2020-12-02
Initialize `botInfo` in `handleUpdate`

KnorpelSenf:
# Description

Title says it all

Fixes https://github.com/telegraf/telegraf/pull/1200#issuecomment-737233820

## Type of change

# How Has This Been Tested?

```bash
npm run test
npm run build
npm run lint
```

# Checklist:

### DIFF
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -235,7 +235,7 @@ export class Telegraf<
     return Promise.race([processAll, sleep(this.options.handlerTimeout)])
   }
 
-  private botInfoCall: Promise<tt.UserFromGetMe> | undefined
+  private botInfoCall?: Promise<tt.UserFromGetMe>
   async handleUpdate(update: tt.Update, webhookResponse?: http.ServerResponse) {
     debug('Processing update', update.update_id)
     this.botInfo ??=
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -78,7 +78,7 @@ export class Telegraf<
   private readonly options: Telegraf.Options<TContext>
   private webhookServer?: http.Server | https.Server
   /** Set manually to avoid implicit `getMe` call in `launch` or `webhookCallback` */
-  public botInfo!: tt.UserFromGetMe
+  public botInfo?: tt.UserFromGetMe
   public telegram: Telegram
   readonly context: Partial<TContext> = {}
   private readonly polling = {
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -125,13 +125,10 @@ export class Telegraf<
   }
 
   webhookCallback(path = '/') {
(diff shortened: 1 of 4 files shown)


src/telegraf.ts
```diff
@@ -238,8 +235,16 @@ export class Telegraf<
...
 
+  private botInfoCall: Promise<tt.UserFromGetMe> | undefined
   async handleUpdate(update: tt.Update, webhookResponse?: http.ServerResponse) {
     debug('Processing update', update.update_id)
+    this.botInfo ??=
+      (debug(
+        'Update',
+        update.update_id,
+        'waiting for `botInfo` to be initialized'
+      ),
```
wojpawlik, line 246: Magic, begone!
```suggestion
```

wojpawlik: I'm against comma operator. You could do

```ts
if (this.botInfo === undefined && this.botInfoCall === undefined) {
  debug("Initializing `botInfo`...)
}
```

On 2nd thought, this would introduce weird logic duplication... I'd drop the `debug` call for now.

KnorpelSenf: > I'm against comma operator.

As in, radically? I feel like logging statements is the only place for them.

wojpawlik: Yes. But on 3rd thought, this message is useful enough, I'll let it in.

wojpawlik (changes requested):
`Telegraf::botInfo` needs to be made optional. Otherwise types are wrong if you try to read it.

KnorpelSenf:
WDYM? I don't see where that can happen. Also, it compiles.

wojpawlik:
https://github.com/telegraf/telegraf/blob/1f7dee72c702bdf05791ccba2a8a8c3b3c6f9dfc/src/telegraf.ts#L81 needs to be changed to

```ts
public botInfo?: tt.UserFromGetMe
```

KnorpelSenf:
I suspect you mean when reading from outside

test/telegraf.js
```diff
@@ -1,6 +1,12 @@
 const test = require('ava')
 const { Telegraf, session } = require('../')
 
+function createBot (...args) {
+  const bot = new Telegraf(...args)
+  bot.botInfo = { id: 42, is_bot: true, username: 'bot', first_name: 'Bot' }
+  return bot
+}
+
```
KnorpelSenf, line 9: Can and should we reuse this from `composer.js`? I did not find a way to make testing modules dependent on a third factory module without ava complaining that there were no tests in it.

wojpawlik: I wouldn't bother. I'm surprised you didn't have to specify more properties in `botInfo`.

wojpawlik (approved):

PR #1211, opened 2020-12-04
add chatType to log function

ssshojaei:
# Description

With a slight change in the log function, I added the possibility of logging a specific chatType

## Type of change

- [ ] updated keyboard-bot example
- [ ] add chatType to log function


wojpawlik:
```ts
bot.use(Composer.privateChat(Composer.log()))
```

PR #1215, opened 2020-12-06
Add `Composer::guard`

wojpawlik:
# Description

`guard`, because it operates on TypeScript type guards.
No `static` variant, because `U` should always be inferred, but users may need to manually specify `C`.

Since type guards are decoupled from `Composer`,
1. they take `update` instead of `ctx` to keep them library-agnostic.
2. I decided to not add any to Telegraf until I decide how to structure them. The most flexible would be to have type guard functions for `Update` (`is.message`, `is.editedMessage`, `is.channelPost`, `is.editedChannelPost`), which take `Message` type guard as argument(s), but I'm not sure whether `Update` and `Message` type guard functions should live in one `is` namespace, or in separate namespaces, which I don't know how to name.

<hr>

While I was at it, I removed `channelMode` which was messing with my types, and drastically simplified and sped up `MatchedContext` type introduced in #1195.

## Type of change

# Checklist:


src/composer.ts
```diff
@@ -12,6 +13,34 @@ type Triggers<C> = MaybeArray<
...
+  C extends Context,
+  U extends Omit<tt.Update, 'update_id'>
+> = C &
+  {
+    [P in tt.UpdateType as SnakeToCamelCase<P>]: PropOr<U, P, undefined>
+  } & {
+    update: U
+    updateType: keyof UnionToIntersection<U>
+  }
+type MountMap = {
```
KnorpelSenf, line 32: I would like to find a better name for this, or add an explanation that highlights its usage so the name becomes clearer. The type basically maps `T in tt.UpdateType | tt.MesssageSubType` to the correct properties on an `Update` that need to be intersected with `Context` in order to narrow it down according to `T`, correct? This makes it correspond to the superseded `GuaranteedContextProps` type, so that would be my first suggestion.

Alternatively, we could add a comment that explains its usage with a terminology that points out what mounting and mapping would mean here.

src/composer.ts
```diff
@@ -12,6 +13,34 @@ type Triggers<C> = MaybeArray<
...
+} &
+  {
+    [T in tt.MessageSubType]: {
+      message: Record<
+        PropOr<typeof MessageSubTypesMapping, T, T>,
+        UnionToIntersection<tt.Message>[T]
+      >
+    }
+  }
+
```
KnorpelSenf, line 43: Should this be grouped together with `MatchedContext`, rather than being separated from it by the helper functions?

KnorpelSenf (commented):
`guard` can be implemented by means of `optional`. Is there a reason not to do this?

Having a `static guard` function would allow us to get rid of a lot of `@ts-expect-error` annotations. Also, even the instance method requires to know the subtype of `Update`. This is not possible:
```ts
bot.guard(
  (u) => 'message' in u,
  (ctx) => {
    const c = ctx.message
  }
)
```
It gives:
```
Argument of type '(u: Update) => boolean' is not assignable to parameter of type '(update: Update) => update is Update'.
  Signature '(u: Update): boolean' must be a type predicate.ts(2345)
```
Hence, the `guardFn` has to be explicitly annotated with `(u): u is Update.MessageUpdate` before it works. It would therefore not be an issue to have a static method with the same requirement.

src/composer.ts
wojpawlik: > correct?

Yes.

Name is good, it's a mapping used exclusively by `Composer.mount` and `Composer::on`, I'll add some JSDoc tho.

src/composer.ts
wojpawlik: If anything, `MatchedContext` and `MatchedMiddleware` are the only type aliases that are not grouped with the rest, and I don't wish to have `git blame` on those lines, so I'm leaving as-is :)

wojpawlik:
> `guard` can be implemented by means of `optional`. Is there a reason not to do this?

`Composer.optional` doesn't support type guards. And the problem with adding that, or static `Composer.guard`, is inconvenience when custom context is needed:

obraz

src/composer.ts
KnorpelSenf: Isn't it a part of `MatchedContext`? So it is basically becoming the central point of type inference for all `Composer` methods, together with `GuardedContext`. It's hence not exclusively used by those two methods

src/composer.ts
KnorpelSenf: I can move them for you if that's the issue. After all, they are heavily dependent on each other and belong semantically together, too.

wojpawlik:
In addition to being slightly inconvenient, static `Composer.guard` would prompt questions about what's the difference from `Composer.optional` if not documented well.

Note that type guards functions understandably must be synchronous, so they're incompatible with `AsyncPredicate<C>` already accepted by `Composer.optional`, so adding an ([inconvenient](https://github.com/telegraf/telegraf/pull/1215#issuecomment-739491471)) overload would be the simplest solution to extending `Composer.optional` to accept type guards.

<hr>

Either way, I'm very happy with the PR as-is, so if that's everything you have issues with, why don't you approve this and open a new pull request dedicated to resolving `// @ts-expect-error`s in `Composer`? There, you could address your naming and ordering concerns in any way you wish.

KnorpelSenf (approved):

KnorpelSenf:
Those are four clean commits, merging without squashing.

KnorpelSenf:
Confer #1217

PR #1217, opened 2020-12-06
Improve `Composer`

KnorpelSenf:
# Description

Introduces a `static guard` function, with a type signature that is comparable to the existing `static mount`.

Improves documentation.

Moves semantically similar things together in the same file.

Fixes linting errors and removes some `@ts-expect-error` comments by using the newly introduced `static guard`.

Boosts the performance of probably _the_ most common composer operation `on` (alias `mount`). Previous: O(n*m), now: O(n) where n is the number of supplied arguments and m is the length of supported update types.

Improves the `gameQuery` types by using the newly introduced `static guard`.

Adds unit tests for `bot.guard`.

## Type of change

- [X] Documentation (typos, code examples or any documentation update)

# How Has This Been Tested?

```bash
npm t
npm run lint
```

# Checklist:

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -414,7 +414,8 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
     )
 
     const predicate = (u: tt.Update): u is tt.Update & MountMap[T] =>
-      updateTypes.some((t) => t in u || ('message' in u && t in u.message))
+      updateTypes.some((t) => t in u) ||
+      ('message' in u && updateTypes.some((t) => t in u.message))
 
     return Composer.guard(predicate, ...fns)
   }
@@ -706,7 +707,7 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
     const types = Array.isArray(type) ? type : [type]
     return Composer.optional((ctx) => {
       const chat = ctx.chat
-      return (chat !== undefined && types.includes(chat.type)) ?? false
+      return chat !== undefined && types.includes(chat.type)
     }, ...fns)
   }
 
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -19,6 +19,23 @@ type UnionToIntersection<U> = (
   ? I
   : never
 type PropOr<T, P extends string, D> = T extends Record<P, infer V> ? V : D
+
+type MatchedMiddleware<
+  C extends Context,
+  T extends tt.UpdateType | tt.MessageSubType = 'message' | 'channel_post'
+> = NonemptyReadonlyArray<
+  Middleware<MatchedContext<C & { match: RegExpExecArray }, T>>
+>
+
+/** Takes: a context type and an update type (or message subtype).
+    Produces: a context that has some properties required, and some undefined.
+    The required ones are those that are always present when the given update (or message) arrives.
+    The undefined ones are those that are always absent when the given update (or message) arrives. */
+type MatchedContext<
+  C extends Context,
+  T extends tt.UpdateType | tt.MessageSubType
+> = GuardedContext<C, MountMap[T]>
+
 /**
  * Narrows down `Context['update']` and some derived properties.
  */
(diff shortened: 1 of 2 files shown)


src/composer.ts
```diff
@@ -48,44 +65,15 @@ type MountMap = {
     }
   }
 
+interface GameQueryUpdate extends tt.Update.CallbackQueryUpdate {
+  callback_query: tt.CallbackQuery.GameShortGameCallbackQuery
+}
```
wojpawlik, line 70: Add to `typegram`?

src/composer.ts
```diff
@@ -382,22 +372,51 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
...
+    const updateTypes = reverseRenameMessageSubtypes(
+      normalizeTextArguments(updateType)
+    )
 
-    const predicate = (ctx: C) =>
-      updateTypes.includes(ctx.updateType) ||
-      // @ts-expect-error `type` is a string and not a union of literals
-      updateTypes.some((type) => ctx.updateSubTypes.includes(type))
+    const predicate = (u: tt.Update): u is tt.Update & MountMap[T] =>
+      updateTypes.some((t) => t in u || ('message' in u && t in u.message))
```
wojpawlik, line 417: This is cryptic, this could be avoided by overloading `Composer.optional` to support type guards... that would also solve `@ts-expect-error` in static `Composer.guard`.

Also, `normalizeTextArguments` could probably be typed better.

src/composer.ts
```diff
@@ -685,9 +704,10 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
     ...fns: NonemptyReadonlyArray<Middleware<C>>
   ): Middleware.Fn<C> {
     const types = Array.isArray(type) ? type : [type]
-    // @ts-expect-error
-    // prettier-ignore
-    return Composer.optional((ctx) => ctx.chat && types.includes(ctx.chat.type), ...fns)
+    return Composer.optional((ctx) => {
+      const chat = ctx.chat
+      return (chat !== undefined && types.includes(chat.type)) ?? false
```
wojpawlik, line 709:
```suggestion
      return ctx.chat !== undefined && types.includes(chat.type)
```

wojpawlik (changes requested):

src/composer.ts
KnorpelSenf: I don't think it belongs there, or else we should also add `TextMessageUpdate` and the like. That would introduce a lot of redundancy.

`telegraf` also just needs it because we explicitly support subtypes of _message_, but nothing else. As a consequence, we need this helper interface when we want to work with other types of `Update`s that contain unions.

src/composer.ts
KnorpelSenf: I tried to overload `Composer.optional`, but did not achieve anything neat. The same is true for `normalizeTextArguments`. The best I could come up with has so many strange types to cover the corner cases (and none of my experiments worked without explicit type casts), so that I decided to drop this. After all, it's still an internal helper method.

If you feel like you can find something beautiful there that solves the problems, even if it's just a rough idea, I'd be happy to hear about it.

KnorpelSenf: I improved the performance in 3ab84b82c67b1e7ab57c4233332969758c239a46. It also makes it more readable. If you still think that this change is cryptic, which part about it is more difficult to understand than the previous implementation?

wojpawlik: This is great!

src/composer.ts
```diff
@@ -808,6 +824,35 @@ function normalizeTriggers<C extends Context>(
...
+  string,
+  string
+> = Object.fromEntries(
+  Object.entries(MessageSubTypesMapping).map(([k, v]) => [v, k])
+)
+function reverseRenameMessageSubtypes(types: string[]): string[] {
+  return types.map((t) =>
+    t in reversedMessageSubTypesMapping ? reversedMessageSubTypesMapping[t] : t
+  )
+}
```
wojpawlik, line 854: I'll try to get rid of this in my next PR

wojpawlik (approved):

PR #1220, opened 2020-12-07
Remove replicators

wojpawlik:
I could straight up remove `sendCopy`, but

> Before you completely remove the functionality in a new major release there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API. (https://semver.org/#how-should-i-handle-deprecating-functionality)


KnorpelSenf:
Does that imply a 3.39.0?

wojpawlik:
I'd rather just deprecate `sendCopy` in 4.0 and remove in 5.0.

KnorpelSenf (approved):

Issue #1104, opened 2020-08-11
Failed to process updates. Error: Updates must be an array


ikerya:
* Telegraf.js Version: 3.38.0
* Node.js Version: 13.10.1
* Operating System: Debian GNU/Linux 8.11 (jessie)

Just got the error. What to do with that? 

```
Failed to process updates. Error: Updates must be an array
     at Telegraf.handleUpdates (/var/www/project/node_modules/telegraf/telegraf.js:153:29)
     at /var/www/dictionary-bot/node_modules/telegraf/telegraf.js:190:16
     at runMicrotasks (<anonymous>)
     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

ikerya:
Any comments?

imdkbj:
Can you share the array what u are suppling to restrict the updates.

ikerya:
I do not restrict anything. This is the only thing i do with the bot's instance:

```
bot.launch({
    polling: { timeout: 3 }
});
```

ikerya:
Any comments? Still happening...

imdkbj:
Hey! I am unable to reproduce the issue. Can u share the complete code after removing sensitive info.

![image](https://user-images.githubusercontent.com/46839821/92246645-ae734c00-eee3-11ea-8f74-a6e28f869ddd.png)


The said error will occurred only if you pass **allowedUpdates**
https://github.com/telegraf/telegraf/blob/develop/src/telegraf.ts

![image](https://user-images.githubusercontent.com/46839821/92246417-56d4e080-eee3-11ea-941e-4e3b03474ff9.png)

ikerya:
![image](https://user-images.githubusercontent.com/49207045/94704849-079a9800-0349-11eb-9695-f140662efdd9.png)

I think the case is not in allowedUpdates, because as it seems, the getUpdates function is called and gets updates of a non array type, then an error is thrown. and the most important thing — the polling stops.

imdkbj:
Check console for err. Pls share the code to reproduce the error.

KnorpelSenf:
@ikerya please install the current `develop` branch and check if this is still an issue. Otherwise, please consider closing this.

KnorpelSenf:
Closing because outdated.

Issue #1202, opened 2020-11-30
4.0 Roadmap


KnorpelSenf:
# 4.0 Major Release Roadmap 

There are a number of tasks and discussions that we need to address (partially, at least), before releasing 4.0 makes sense.

As a quick recap, the main point of 4.0 is the migration to TypeScript, as tracked by #1074. Also, support for Telegram Bot API 5.0 will arrive.

This issue is meant to serve as a discussion forum for all the points below.

- [x] [ `channelMode`:](https://github.com/telegraf/telegraf/pull/1195#issue-528642015) ~~leave it there or~~ throw it out? ~~If leave: respect it in type annotations? Else, types are wrong for users of the feature.~~ There will no longer be a channel mode, but other means to achieve the same thing instead. They are yet to be developed.
- [x] `bot.on`: ~~[refactor](https://github.com/telegraf/telegraf/pull/1195#issuecomment-735421884) or~~ leave as-is. <ins>A new method can be introduced anytime</ins>.
- [x] #1194 needs to be reviewed: ~~**help wanted**. Please checkout the branch `fix-scene-types` and verify it works!~~
- [x] #1195 should be played around with by preferably at least two other people to make sure we didn't screw up: ~~**help wanted**. Please checkout the branch `leverage-unionization` and verify it works!~~
- [x] Example bots have to be updated.
- [x] New example bots should be written that illustrate TypeScript usage.
- [x] #1280
- [x] #1282

wojpawlik:
I managed to shoehorn type guard support into `Composer.optional` (using overloading or `// @ts-expect-error`). I also managed to write a standalone one that does typecheck (how do I name it?). If it lands in 4.0.0 alongside a set of suitable type guards, I don't care what happens to `channelMode`, `Composer.mount`, `Composer::on`, `Context::updateType` and `Context::updateSubTypes`.

Which examples need updating?

Docs improvements are blocked by https://telegraf.js.org/ not pointing to typedoc-generated docs.

In general, docs aren't blocking 4.0.0. Improved docs can be released even in a patch release.

KnorpelSenf:
I'm intrigued—can you share the code for `Composer.optional`?

A standalone version of what exactly?

Generally, I believe we should go through the examples once again after #1195, in case they need to be adapted. I don't have anything specific in mind yet.

I agree about the docs. I removed them from this list so that we can stay focused on the release. Doesn't mean we don't need them eventually.

KnorpelSenf:
[...]

PR #1224, opened 2020-12-08
Rewrite polling

### LINKED ISSUE #735: Long Polling drops after 1 day
* Telegraf.js Version: 3.32.0
* Node.js Version: v12.5.0
* Operating System: Ubuntu 18.04

Hi, I have a problem where the bot stops responding to any message after 1 day or so. I use long polling method because my server is in a corporate intranet with proxy connection to internet.  So webhook is no option for me. 

I use PM2 and the process is still running fine, the bot is still running. no errors from log.  I've been searching for answers for 3 days and haven't found it yet, so I'm writing here. please help..

### LINKED ISSUE #851: Processing an update twice when graceful shutdowned
## Context

Hey,

I've implemented graceful shutdown for my bot and face an issue. I've created a small demo project: https://github.com/mbelsky/tf-update-issue/blob/master/index.js

* Telegraf.js Version: 3.35.0
* Node.js Version: 12.13.0
* Operating System: MacOS 10.15.1

## Expected Behavior

An updated handled once

## Current Behavior

An updated handled twice

### Steps to Reproduce

1. `npm start`
2. Send '/long' message to a bot
3. Send SIGINT to the node process while it processing update
4. Let the process die
5. `npm start` again 

Screenshot 2020-01-03 at 01 29 35

As you may see it processes `933862816` twice.

I think it could be fixed with adding `offset` arg in `Telegraf. startPolling` method, but before making a PR I'm interested to know your opinion about the preferred way to fix this issue.

### LINKED ISSUE #1023: What is "Failed to fetch updates"?
I got this error, even I'm not using bot.
`Failed to fetch updates. Waiting: 1s request to https://api.telegram.org/bot......botToken......./getUpdates?offset=0&limit=100&timeout=30 failed, reason: read ECONNRESET`
**What is it? And how can I turn off it? In what kind of reason it can be?**
**UPDATED**: I got this error after some moment of sending message from another program side of api: 
```
const Telegraf = require('telegraf');
const bot_token = config.get('telegram.bot-token');
const bot = new Telegraf(bot_token);
```
And I send message like this -> 
```
bot.telegram.sendMessage(adminChannelId, orderMessage, {parse_mode: "HTML"});
return jsonPreProcessor.success({payment: payment, orderArrivals: orderArrivals});
```
Is it normal? Or I did something wrong? Telegram bot is starting from server.js with bot.launch();

wojpawlik:
I set out to fix #851. To keep my sanity, I extracted all polling logic into a separate class. Also fixed some other issues.

- Fixes #735, `handlerTimeout` now defaults to 5 seconds.
- Fixes #851.
- Fixes #1023, the error is now logged using `debug` and thus isn't displayed by default.
- #1104 throws a nicer error.

Also removed some barely useful configuration options.

Perhaps webhook logic should be extracted to a separate file as well? https://github.com/garronej/denoify/issues/13#issuecomment-689076929

## Type of change

## How Has This Been Tested?

By starting and stopping a bot in various ways :)

**Test Configuration**: Node.js Version: 14.0.0

# Checklist:

### DIFF
--- a/bin/telegraf
+++ b/bin/telegraf
@@ -103,5 +103,6 @@ bot.use(botHandler)
 log(`Starting module ${file}`)
 bot.launch(config)
 
+// Enable graceful stop
 process.once('SIGINT', () => bot.stop('SIGINT'))
 process.once('SIGTERM', () => bot.stop('SIGTERM'))
--- a/bin/telegraf
+++ b/bin/telegraf
@@ -103,8 +103,5 @@ bot.use(botHandler)
 log(`Starting module ${file}`)
 bot.launch(config)
 
-function stop () {
-  bot.stop()
-}
-process.once('SIGINT', stop)
-process.once('SIGTERM', stop)
+process.once('SIGINT', () => bot.stop('SIGINT'))
+process.once('SIGTERM', () => bot.stop('SIGTERM'))
--- a/bin/telegraf
+++ b/bin/telegraf
@@ -15,7 +15,6 @@ const help = () => {
   -d  Webhook domain
   -H  Webhook host [0.0.0.0]
   -p  Webhook port [$PORT or 3000]
-  -s  Stop on error
   -l  Enable logs
   -h  Show this help message`)
 }
@@ -27,7 +26,6 @@ const args = parse(process.argv, {
     H: 'host',
     h: 'help',
     l: 'logs',
-    s: 'stop',
     p: 'port'
   },
   boolean: ['h', 'l', 's'],
@@ -100,10 +98,13 @@ if (domain) {
 }
 
 const bot = new Telegraf(token)
-if (!args.stop) {
-  bot.catch(log)
-}
 bot.use(botHandler)
 
 log(`Starting module ${file}`)
 bot.launch(config)
+
+function stop () {
+  bot.stop()
+}
+process.once('SIGINT', stop)
+process.once('SIGTERM', stop)
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   },
   "types": "./typings/index.d.ts",
   "dependencies": {
+    "abort-controller": "^3.0.0",
     "debug": "^4.2.0",
     "minimist": "^1.2.5",
     "module-alias": "^2.2.2",
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -5,6 +5,7 @@ import d from 'debug'
 import { promisify } from 'util'
 const wait = promisify(setTimeout)
 const debug = d('telegraf:polling')
+const noop = () => {}
 
 export class Polling {
   private readonly abortController = new AbortController()
@@ -40,17 +41,21 @@ export class Polling {
         await wait(retryAfter * 1000)
       }
     } while (!this.abortController.signal.aborted)
-    await this.confirmUpdates()
   }
 
-  async confirmUpdates() {
+  private async confirmUpdates() {
     debug('Confirming updates...')
     await this.telegram.callApi('getUpdates', { offset: this.offset, limit: 1 })
   }
 
   async loop(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
-    for await (const updates of this) {
-      await handleUpdates(updates)
+    try {
+      for await (const updates of this) {
+        await handleUpdates(updates)
+      }
+    } finally {
+      debug('Long polling stopped.')
+      await this.confirmUpdates().catch(noop)
     }
   }
 
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -10,23 +10,27 @@ const noop = () => {}
 export class Polling {
   private readonly abortController = new AbortController()
   private stopCallback = noop
+  private offset = 0
   constructor(
     private readonly telegram: ApiClient,
     private readonly allowedUpdates: readonly tt.UpdateType[]
   ) {}
 
   async *[Symbol.asyncIterator](): AsyncGenerator<tt.Update[], void, void> {
-    let offset = 0
     do {
       try {
         const updates = await this.telegram.callApi(
           'getUpdates',
-          { timeout: 60, offset, allowed_updates: this.allowedUpdates },
+          {
+            timeout: 50,
+            offset: this.offset,
+            allowed_updates: this.allowedUpdates,
+          },
           this.abortController
         )
         const last = updates[updates.length - 1]
         if (last !== undefined) {
-          offset = last.update_id + 1
+          this.offset = last.update_id + 1
         }
         yield updates
       } catch (err) {
@@ -34,12 +38,17 @@ export class Polling {
         if (err.code === 401 || err.code === 409) throw err
         const retryAfter = err.parameters?.retry_after ?? 5
         // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-        debug(`Failed to fetch updates. Retrying after ${retryAfter}s`, err)
+        debug(`Failed to fetch updates, retrying after ${retryAfter}s.`, err)
         await wait(retryAfter * 1000)
       }
     } while (!this.abortController.signal.aborted)
-    await this.telegram.callApi('getUpdates', { timeout: 0, offset })
+    await this.confirmUpdates()
     return this.stopCallback()
+  }
+
+  async confirmUpdates() {
+    debug('Confirming updates...')
+    await this.telegram.callApi('getUpdates', { offset: this.offset })
   }
 
   async consume(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -5,11 +5,9 @@ import d from 'debug'
 import { promisify } from 'util'
 const wait = promisify(setTimeout)
 const debug = d('telegraf:polling')
-const noop = () => {}
 
 export class Polling {
   private readonly abortController = new AbortController()
-  private stopCallback = noop
   private offset = 0
   constructor(
     private readonly telegram: ApiClient,
@@ -18,6 +16,7 @@ export class Polling {
 
   async *[Symbol.asyncIterator](): AsyncGenerator<tt.Update[], void, void> {
     do {
+      debug('Starting long polling')
       try {
         const updates = await this.telegram.callApi(
           'getUpdates',
@@ -34,7 +33,7 @@ export class Polling {
         }
         yield updates
       } catch (err) {
-        if (err.name === 'AbortError') return this.stopCallback()
+        if (err.name === 'AbortError') return
         if (err.code === 401 || err.code === 409) throw err
         const retryAfter = err.parameters?.retry_after ?? 5
         // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
@@ -43,7 +42,6 @@ export class Polling {
       }
     } while (!this.abortController.signal.aborted)
     await this.confirmUpdates()
-    return this.stopCallback()
   }
 
   async confirmUpdates() {
@@ -51,7 +49,7 @@ export class Polling {
     await this.telegram.callApi('getUpdates', { offset: this.offset })
   }
 
-  async consume(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
+  async loop(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
     for await (const updates of this) {
       await handleUpdates(updates)
     }
@@ -59,9 +57,6 @@ export class Polling {
 
   stop() {
     debug('Gracefully stopping long polling...')
-    return new Promise<void>((resolve) => {
-      this.stopCallback = resolve
-      this.abortController.abort()
-    })
+    this.abortController.abort()
   }
 }
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -80,7 +80,6 @@ export class Polling {
   }
 
   stop() {
-    debug('Stopping long polling...')
     this.abortController.abort()
   }
 }
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -1,0 +1,58 @@
+import * as tt from '../../telegram-types'
+import AbortController from 'abort-controller'
+import ApiClient from './client'
+import d from 'debug'
+import { promisify } from 'util'
+const wait = promisify(setTimeout)
+const debug = d('telegraf:polling')
+const noop = () => {}
+
+export class Polling {
+  private readonly abortController = new AbortController()
+  private stopCallback = noop
+  constructor(
+    private readonly telegram: ApiClient,
+    private readonly allowedUpdates: readonly tt.UpdateType[]
+  ) {}
+
+  async *[Symbol.asyncIterator](): AsyncGenerator<tt.Update[], void, void> {
+    let offset = 0
+    do {
+      try {
+        const updates = await this.telegram.callApi(
+          'getUpdates',
+          { timeout: 60, offset, allowed_updates: this.allowedUpdates },
+          this.abortController
+        )
+        const last = updates[updates.length - 1]
+        if (last !== undefined) {
+          offset = last.update_id + 1
+        }
+        yield updates
+      } catch (err) {
+        if (err.name === 'AbortError') return this.stopCallback()
+        if (err.code === 401 || err.code === 409) throw err
+        const retryAfter = err.parameters?.retry_after ?? 5
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        debug(`Failed to fetch updates. Retrying after ${retryAfter}s`, err)
+        await wait(retryAfter * 1000)
+      }
+    } while (!this.abortController.signal.aborted)
+    await this.telegram.callApi('getUpdates', { timeout: 0, offset })
+    return this.stopCallback()
+  }
+
+  async consume(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
+    for await (const updates of this) {
+      await handleUpdates(updates)
+    }
+  }
+
+  stop() {
+    debug('Gracefully stopping long polling...')
+    return new Promise<void>((resolve) => {
+      this.stopCallback = resolve
+      this.abortController.abort()
+    })
+  }
+}
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -14,7 +14,7 @@ export class Polling {
     private readonly allowedUpdates: readonly tt.UpdateType[]
   ) {}
 
-  async *[Symbol.asyncIterator](): AsyncGenerator<tt.Update[], void, void> {
+  private async *[Symbol.asyncIterator]() {
     debug('Starting long polling')
     do {
       try {
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -45,7 +45,7 @@ export class Polling {
 
   async confirmUpdates() {
     debug('Confirming updates...')
-    await this.telegram.callApi('getUpdates', { offset: this.offset })
+    await this.telegram.callApi('getUpdates', { offset: this.offset, limit: 1 })
   }
 
   async loop(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -35,8 +35,7 @@ export class Polling {
       } catch (err) {
         if (err.name === 'AbortError') return
         if (err.code === 401 || err.code === 409) throw err
-        const retryAfter = err.parameters?.retry_after ?? 5
-        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        const retryAfter: number = err.parameters?.retry_after ?? 5
         debug(`Failed to fetch updates, retrying after ${retryAfter}s.`, err)
         await wait(retryAfter * 1000)
       }
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -4,8 +4,8 @@ import ApiClient from './client'
 import d from 'debug'
 import { promisify } from 'util'
 import { TelegramError } from './error'
-const wait = promisify(setTimeout)
 const debug = d('telegraf:polling')
+const wait = promisify(setTimeout)
 const noop = () => {}
 
 export class Polling {
@@ -51,7 +51,7 @@ export class Polling {
   }
 
   private async confirmUpdates() {
-    debug('Confirming updates...')
+    debug('Confirming received updates...')
     await this.telegram.callApi('getUpdates', { offset: this.offset, limit: 1 })
   }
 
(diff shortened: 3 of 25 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
   },
   "types": "./typings/index.d.ts",
   "dependencies": {
+    "abort-controller": "^3.0.0",
```
KnorpelSenf, line 51: Can I assume that you verified that there is no way around adding another dep if we want to process updates reliably around starting and stopping bots?

KnorpelSenf (commented):
Looks pretty good so far. I will checkout your code and play around with that tomorrow, so I can confirm that it works.

package.json
wojpawlik: Except for reimplementing it or dropping support for Node <15, there's no way.

KnorpelSenf: Yeah I see now that it's the way to go if you use `node-fetch` which we do.

src/core/network/polling.ts
```diff
@@ -0,0 +1,61 @@
...
+        if (last !== undefined) {
+          this.offset = last.update_id + 1
+        }
+        yield updates
+      } catch (err) {
+        if (err.name === 'AbortError') return
+        if (err.code === 401 || err.code === 409) throw err
+        const retryAfter: number = err.parameters?.retry_after ?? 5
+        debug('Failed to fetch updates, retrying after %ds.', retryAfter, err)
+        await wait(retryAfter * 1000)
```
KnorpelSenf, line 40: I can see that you copied this from the old code, but why is it even that we only rethrow upon 401 and 409? Wouldn't it make more sense to rather check for 429 or for `retry_after` and throw in all other cases?

src/core/network/polling.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export class Polling {
+  private readonly abortController = new AbortController()
+  private offset = 0
+  constructor(
+    private readonly telegram: ApiClient,
+    private readonly allowedUpdates: readonly tt.UpdateType[]
+  ) {}
+
+  async *[Symbol.asyncIterator](): AsyncGenerator<tt.Update[], void, void> {
```
KnorpelSenf, line 17: Fancy and I like it, but should this be `private`?

src/core/network/polling.ts
```diff
@@ -0,0 +1,61 @@
...
+        await wait(retryAfter * 1000)
+      }
+    } while (!this.abortController.signal.aborted)
+    await this.confirmUpdates()
+  }
+
+  async confirmUpdates() {
+    debug('Confirming updates...')
+    await this.telegram.callApi('getUpdates', { offset: this.offset, limit: 1 })
+  }
```
KnorpelSenf, line 49: It is not clear why we need this call. Please add a brief comment explaining why it is necessary to fetch an update and immediately discard it.

Also, it seems to be called inconsistently depending on where an error is thrown, but I cannot really pinpoint this without knowing what it actually does.

Is it relevant here to pass `allowed_updates`?

src/telegraf.ts
```diff
@@ -69,30 +59,21 @@ namespace Telegraf {
   }
 }
 
-const allowedUpdates: tt.UpdateType[] | undefined = undefined
-
 export class Telegraf<
   TContext extends Context = Context
```
KnorpelSenf, line 63: As you're touching this file anyway, it might be the right time to rename `TContext` to `C` here, as well.

src/telegraf.ts
```diff
@@ -69,30 +59,21 @@ namespace Telegraf {
...
-    stopCallback: noop,
-    timeout: 30,
-  }
 
-  private handleError: (err: unknown, ctx: TContext) => void = (err: any) => {
-    console.error()
-    console.error((err.stack ?? err.toString()).replace(/^/gm, '  '))
-    console.error()
+  private handleError = async (err: unknown, ctx: TContext): Promise<void> => {
+    process.exitCode = 1
```
KnorpelSenf, line 74: This does not belong into `telegraf` itself, it should be the bot that determines its exit codes.

However, it certainly does belong into `bin/telegraf`, so maybe we should move this line there.

src/telegram.ts
```diff
@@ -41,6 +41,9 @@ class Telegram extends ApiClient {
     return `${this.options.apiRoot}/file/bot${this.token}/${fileId.file_path}`
   }
 
+  /**
+   * @deprecated use `Telegraf::launch`
+   */
```
KnorpelSenf, line 46: Why adding `@depecated`? If a user what's to get the raw updates directly for whatever, this should be supported.

However, I do agree that we should add a comment here, explaining why this method is rarely useful. It must not become a pitfall for new users, so the method should never be called by accident.

KnorpelSenf (commented):
Great work. I just have a few minor things and questions:

package.json
wojpawlik: We could alternatively use more lightweight but less popular https://www.npmjs.com/package/node-abort-controller, I'll try it tomorrow

KnorpelSenf:
Does this introduce breaking changes or rather improve the reliability?

I'm interested in getting to the point where we can actually release 4.0, so anything that is not mission-critical may also be merged afterwards IMO. If this PR can make it before—even better. Still I think it makes sense to only add things to the 4.0 Roadmap #1202 that really need to be there (because they break stuff).

wojpawlik:
> Does this introduce breaking changes

Yes, `LaunchOptions` are revamped, so it'll be awkward if it doesn't make it into 4.0.

src/core/network/polling.ts
wojpawlik: I could link to https://github.com/telegraf/telegraf/issues/851#issuecomment-736025893 in a comment...

Or, I could do something that sounds more appealing the more I think about it: call `confirmUpdates` in `finally` clause in `loop`. Disadvantage: call would happen even if not necessary. Advantages:

+ Bulletproof. It'll always be called, always exactly once (unless the process is killed of course).
+ There's no need to call `confirmUpdates` anywhere else, so it can be made `private`.


> Is it relevant here to pass `allowed_updates`?

Not really. We don't care what we get. And it defaults to value from previous call anyway.

KnorpelSenf: > I could link to https://github.com/telegraf/telegraf/issues/851#issuecomment-736025893 in a comment...

Yes, or just explain it :)

I won't be the last one to wonder about this, and a quick sentence or two is much more convenient.

> call `confirmUpdates` in `finally` clause in `loop`

That actually sounds much clearer. I think reducing the time from dozens of seconds to a few msec justifies one superfluous call every blue moon, just to be safe.

src/telegraf.ts
wojpawlik: This tries to emulate `warn-with-error-code` [unhandled rejections mode](https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode) when `warn` mode is used (default before Node 15). It effectively changes nothing in `throw`, `strict`, and `warn-with-error-code` modes. Without this, the process will cleanly exit with status code 0, perhaps fooling process manager into not restarting the process. Any better ideas?

KnorpelSenf:
```suggestion
    // set exit code to emulate `warn-with-error-code` behavior of
    // https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode
    // to prevent a clean exit despite an error being thrown
    process.exitCode = 1
```

bin/telegraf
```diff
@@ -100,10 +98,10 @@ if (domain) {
...
-if (!args.stop) {
-  bot.catch(log)
-}
 bot.use(botHandler)
 
 log(`Starting module ${file}`)
 bot.launch(config)
+
+process.once('SIGINT', () => bot.stop('SIGINT'))
+process.once('SIGTERM', () => bot.stop('SIGTERM'))
```
wojpawlik, line 107: I'm considering adding these to all examples

src/core/network/polling.ts
wojpawlik: I'm still open to better `debug` message

src/telegraf.ts
```diff
@@ -69,34 +59,25 @@ namespace Telegraf {
...
-    console.error((err.stack ?? err.toString()).replace(/^/gm, '  '))
-    console.error()
+  readonly context: Partial<C> = {}
+
+  private handleError = async (err: unknown, ctx: C): Promise<void> => {
+    // set exit code to emulate `warn-with-error-code` behavior of
+    // https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode
+    // to prevent a clean exit despite an error being thrown
+    process.exitCode = 1
+    debug('Middleware threw', err)
```
wojpawlik, line 76: Perhaps this is redundant now, when `handleError` doesn't do much, and `confirmUpdates` cannot replace the error?

KnorpelSenf: Yes.
```suggestion
```

src/core/network/polling.ts
KnorpelSenf: I'll throw some suggestions at you. Just pick any, or none, or any combination:
- `Saving update offset`
- `Ensuring correct update offset`
- `Acknowledging updates`
- `Confirming received updates to Telegram`
- `Syncing update queue`
- `Securing update index`
I can keep going if you dislike all of them.

src/core/network/polling.ts
wojpawlik: What errors are we dealing with and how should we handle them?

- `AbortError` - `return`
- 401 Unauthorized - unrecoverable, `throw err`
- 409 Conflict - unrecoverable, `throw err`
- 429 Too Many Requests - if it ever happens, wait and retry
- 5xx - wait and retry
- `FetchError` - wait and retry, it's probably connectivity issue
- Did I forget anything?
- Unknown errors - `throw err` I guess?

Note that throwing here = crashing.

src/telegram.ts
wojpawlik: My logic was, it's not used internally anymore, and probably shouldn't be used by users either. Users can use well-typed `Telegram::callApi` if they want to proceed anyway.

src/core/network/polling.ts
wojpawlik: I wish I could use `continue` here.

wojpawlik: > `getUpdates` never returns 429.
https://t.me/tdlibchat/5368

KnorpelSenf: > I wish I could use `continue` here.

Why is that not possible? :thinking:

KnorpelSenf: Regarding the errors, we should maybe just leave it like that and try to figure out what's best over time. I cannot come up with an ad hoc solution here, especially not when we don't have any official source that tells use which errors could be returned (do we?).

wojpawlik: >> I wish I could use `continue` here.
>
> Why is that not possible? :thinking: 

Bare `if (...) continue` wouldn't `await wait(...)`, I had to get more creative than that.

I decided to retry only on `FetchError` and `>= 500` error codes, people will open issues if it's too harsh, at least we know we aren't hiding any important errors.

package.json
wojpawlik: `node-abort-controller` produces this:

```
src/core/network/polling.ts:30:11 - error TS2345: Argument of type 'AbortController' is not assignable to parameter of type 'CallApiOptions'.
  The types of 'signal.onabort' are incompatible between these types.        
    Type '((this: AbortSignal, ev: Event) => any) | null' is not assignable to type '((this: AbortSignal, event: any) => void) | null | undefined'.       
      Type '(this: AbortSignal, ev: Event) => any' is not assignable to type 
'(this: AbortSignal, event: any) => void'.
        The 'this' types of each signature are incompatible.
          Type 'import("D:/Wojciech/JS/telegraf/node_modules/@types/node-fetch/externals").AbortSignal' is not assignable to type 'AbortSignal'.
            Types of property 'onabort' are incompatible.
              Type '((this: AbortSignal, event: any) => void) | null | undefined' is not assignable to type '((this: AbortSignal, ev: Event) => any) | null'.
                Type 'undefined' is not assignable to type '((this: AbortSignal, ev: Event) => any) | null'.
```

Sticking with well-typed `abort-controller`.

src/telegram.ts
wojpawlik: I undeprecated it already, should I un-undeprecate? :)

KnorpelSenf: No, I still think that it should not be deprecated, but I understood your point now.

I suggest to leave it undeprecated

KnorpelSenf:
I tried out the changes on `echo-bot.js`. They work pretty flawlessly. There are two things I'd like to discuss:

1. When starting the bot `node echo-bot.js` and stopping it immediately again (`C-c` as fast as you can type), the process takes several seconds before it completes. In contrast to that, when the bot has been running for more than 5 seconds, it can complete immediately. Is this the intended behaviour (and why is it even happening)?
2. I started the bot and disconnected my machine from the network.
```
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 954758540, allowed_updates: [] } +50s
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 954758540, allowed_updates: [] } +5m
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 954758540, allowed_updates: [] } +50s
```
Notice how the request is pending for five minutes, even though the timeout is set to 50 s. We certainly should not throw when the connection fails, this would be a pain on unstable connections. However, maybe it is also not desired that the bot would just hang indefinitely, until the connection is restored. As an example, changes to the network configuration would require a clean reconnect of the bot. This does happen frequently on mobile, but most users are probably not running their bot inside Termux or something. Are there ways to host a bot that could require this?

wojpawlik:
> 1. When starting the bot `node echo-bot.js` and stopping it immediately again (`C-c` as fast as you can type), the process takes several seconds before it completes.

Can't reporoduce, but now I experienced something _worse_: capturing `SIGINT` _before_ `bot.polling` is set! As a result, bot keeps running and will now close dirtily on next `SIGING`. The real solution to that is top-level await...

I'll be going offline for a moment to test the no network scenario.

wojpawlik:
Perhaps Telegraf should refuse `.launch()` after `.stop()`?

> Notice how the request is pending for five minutes

Can't reproduce that as well, maybe because I'm on Windows.

> As an example, changes to the network configuration would require a clean reconnect of the bot.

Why?

KnorpelSenf:
> > Notice how the request is pending for five minutes
> 
> Can't reproduce that as well, maybe because I'm on Windows.

That might be it, I cannot figure it out. I might look at it again in a while if I have more time.

> > As an example, changes to the network configuration would require a clean reconnect of the bot.
> 
> Why?

Once you send the request and wait for its response, you cannot receive these packets from another connection. I don't know the technical details (I know who to ask if you're interested), but I assume that you can try this out by running the bot on a phone.

wojpawlik:
Always willing to learn :)

Can it be fixed by setting `timeout: 60000` or higher on `fetch`?

wojpawlik:
I added 2 minute timeout to every request, does it fix the pending requests issue? Don't forget to `npm run build` before trying :)

KnorpelSenf:
> > > Notice how the request is pending for five minutes
> > 
> > 
> > Can't reproduce that as well, maybe because I'm on Windows.
> 
> That might be it, I cannot figure it out. I might look at it again in a while if I have more time.

Added `this.abortController` to the offset syncing call defeats the whole point, is that true? As this does indeed solve the “problem”, I assume that it's actually the HTTP request that takes so long. I suspect it's some sort of backend thing that TG cannot do setup and teardown with a split second. I might be mistaken.

> I added 2 minute timeout to every request, does it fix the pending requests issue? Don't forget to `npm run build` before trying :)

Your last commits do indeed solve the problem. The `handlerTimeout` of 5000 ms is respected, so a new request is sent every five seconds. I don't know how this relates to the two minutes you set.

KnorpelSenf:
Please run `npm audit fix` now that you're updating `package-lock.json` anyway

KnorpelSenf:
> but now I experienced something _worse_: capturing `SIGINT` _before_ `bot.polling` is set!

That's something I cannot reproduce but perhaps because I'm not on Windows. Sounds like it's easy to fix this with a flag, though.

KnorpelSenf:
One more observation: as you know, it is not possible to have the same bot running twice. Terminated by other `getUpdates` request is the keyword here.

Usually, if instance A is running and you start instance B, then A's request will fail, and A exits. B continues to run, it effectively takes over control.

This PR attempts to do another `getUpdates` request upon shutdown. As a result, the following happens:
1. A calls `getUpdates`
2. B calls `getUpdates`
3. A's request is terminated
4. A shuts down
5. A tries to sync the offset, hence making a last `getUpdates` request before exiting
6. B's request is terminated
7. B shuts down
8. B syncs the offset before by calling `getUpdates`, and then exists
9. No instance is running

Generally, it is the bot developer's responsibility to make sure that there is only ever a single active instance, however, I feel like it would be better to aim for the old behaviour here.

As an example, if A's request is terminated due to B's request, it makes sense to skip step 5, i.e. syncing the offset.

src/core/network/polling.ts
```diff
@@ -45,6 +48,10 @@ export class Polling {
           await wait(retryAfter * 1000)
           continue
         }
+        if (err instanceof TelegramError && err.code === 409) {
+          this.confirmUpdates = noop
```
KnorpelSenf, line 52: This makes it impossible to restart the bot after the error!

Also, arbitrarily rewriting instance members on the fly is generally really causing mindfuck bugs. The same solution can be achieved with a simple flag and not much more effort.

KnorpelSenf (changes requested):

src/core/network/polling.ts
wojpawlik: `throw`ing here causes unhandled rejection (= process crash on Node 15, preventing any attempts at reusing `Polling`).

401 should probably also avoid confirming, since it's guaranteed to fail.

wojpawlik: `Polling` objects are meant to be recreated, not reused. To make reuse impossible, I can abort the controller in `finally`. It can't be replaced since it's `readonly`.

wojpawlik: I don't know how the member would need to be named so that the code is as clear as `this.confirmUpdates = noop`. It wouldn't be used for anything else.

wojpawlik:
> Adding `this.abortController` to the offset syncing call defeats the whole point, is that true?

Yes. If you're gonna abort that request, might as well not make it in the first place.

> Your last commits do indeed solve the problem. The `handlerTimeout` of 5000 ms is respected, so a new request is sent every five seconds. I don't know how this relates to the two minutes you set.

Weird but ok.

Any other concerns before this can be approved?

src/core/network/polling.ts
```diff
@@ -0,0 +1,86 @@
...
+    try {
+      for await (const updates of this) {
+        await handleUpdates(updates)
+      }
+    } finally {
+      debug('Long polling stopped')
+      // prevent instance reuse
+      this.abortController.abort()
+      await this.confirmUpdates().catch(noop)
+    }
```
KnorpelSenf, line 79: Just use this.
```suggestion
    try {
      for await (const updates of this) {
        await handleUpdates(updates)
      }
    } catch (err) {
      if (
        !(
          err instanceof TelegramError &&
          (err.code === 401 || err.code === 409)
        )
      ) {
        await this.confirmUpdates().catch(noop)
      }
    } finally {
      debug('Long polling stopped')
    }
```
Now you can get rid of
1) some error branching logic above
2) that unconventional method rewriting stuff ([this mistake was made before](https://github.com/telegraf/telegraf/blob/359f4f106dddccabfa6523c3bbd6ddf79a8b826a/extra.js#L45-L46))

Also, there's in fact no need for util flags and its perfectly readable.

src/core/network/polling.ts
KnorpelSenf: > causes unhandled rejection

does it really? It is awaited inside a loop, no?

src/core/network/polling.ts
wojpawlik: Pretty good, but slightly wrong: `confirmUpdates` is not called on successful graceful stop.

KnorpelSenf: > I can abort the controller in `finally`

That might still be a good idea if you still want to ensure that `Polling` objects won't be reused. Note that I skipped those lines in the suggestion above.

KnorpelSenf:
> Any other concerns before this can be approved?

Nothing apart from the method rewriting.

Good job.

src/core/network/polling.ts
KnorpelSenf: I'm missing the rethrowing part! We have to rethrow the error in the new catch block of course. Don't use the suggestion as-is :smile:

src/core/network/polling.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  stop() {
+    debug('Stopping long polling...')
```
wojpawlik, line 83: This feels a little redundant:

```
  telegraf:main Stopping bot... Reason: SIGINT +57s
  telegraf:polling Stopping long polling... +1m
  telegraf:polling Long polling stopped +6ms
```

```suggestion
```

KnorpelSenf: Good catch

src/core/network/polling.ts
wojpawlik: Yes, but `.loop` is not awaited.

src/core/network/polling.ts
KnorpelSenf: > Pretty good, but slightly wrong: confirmUpdates is not called on successful graceful stop.

Only saw that now, you're right, fixed it

KnorpelSenf: If the `catch` this is not DRY enough, you can move it to `confirmUpdates` internally.

wojpawlik: How? it has no access to `err`.

wojpawlik: And yes, your suggestion is not DRY enough, it's not clearly superior to mine.

wojpawlik: If you don't have anything better, why don't you approve this, and perhaps change it next time you'll be touching the file?

wojpawlik: Retrying logic could be extracted into a separate function and reused for `getMe` in `.launch()` and `.handleUpdate()`, or even exported to library users; then it could no longer concern itself with internals of `Polling`.

KnorpelSenf: > And yes, your suggestion is not DRY enough, it's not clearly superior to mine.

You do not understand what I was going for. However, given that you just merged before we could reach consensus, I'll clean up the mess in another PR.

wojpawlik:
Perhaps this could improve our error handling: https://www.npmjs.com/package/is-retry-allowed

KnorpelSenf:
> Perhaps this could improve our error handling: https://www.npmjs.com/package/is-retry-allowed

It is certainly a good idea to use this for error handling, but I'm not sure how strictly this is necessary. There is a myriad of small packages that do some things a little better.

As we rethrow the errors, it is probably simple for library users to leverage that package. We might add a section to the docs, though, and include a code snippet that shows how to perform retries the proper way.

wojpawlik:
Errors thrown from `Polling` cannot be caught: https://github.com/telegraf/telegraf/blob/a221c8b99db3fd005c447a9728068e15962a7cb2/src/telegraf.ts#L118-L119

KnorpelSenf:
Well that's bad. But the fix is a `catch` call and not a library.

PR #1227, opened 2020-12-09
Support custom key-value stores in session

septs:
# Description

Allow use external (asynchronous) key-value database. (e.q: level db / mongodb / etc)
Read/Write session data on external persistent database service.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Node.js Version:
* Operating System:

# Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -485,6 +485,15 @@ const session = require('telegraf/session')
 
 const bot = new Telegraf(process.env.BOT_TOKEN)
 bot.use(session())
+// customize session make key algorithm
+//   bot.use(session((ctx) => `${ctx.from.id}:${ctx.chat.id}`))
+//
+// customize session storage backend
+//   bot.use(session(null, {
+//     getItem(name) { ... },
+//     setItem(name, value) { ... },
+//     deleteItem(name) { ... }
+//   }))
 bot.on('text', (ctx) => {
   ctx.session.counter = ctx.session.counter || 0
   ctx.session.counter++
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,14 +1,15 @@
 import { Context } from './context'
 import { Middleware } from './types'
 
-export type StorageKeyFn<T extends Context> = (
-  ctx: T
-) => Promise<string | undefined>
-
 export interface Storage<T> {
   getItem: (name: string) => Promise<T | undefined>
   setItem: (name: string, value: T) => Promise<void>
   deleteItem: (name: string) => Promise<void>
+}
+
+interface SessionOptions<SessionData> {
+  makeKey?: (ctx: Context) => Promise<string | undefined>
+  storage?: Storage<SessionData>
 }
 
 /**
@@ -19,9 +20,10 @@ export interface Storage<T> {
  * see https://npm.im/search?q=telegraf-session
  */
 export function session<SessionData extends object>(
-  makeKey: StorageKeyFn<Context> = makeDefaultKey,
-  storage: Storage<SessionData> = new MemorySessionStorage<SessionData>()
+  options?: SessionOptions<SessionData>
 ): Middleware.ExtFn<Context, { session?: SessionData }> {
+  const makeKey = options?.makeKey ?? makeDefaultKey
+  const storage = options?.storage ?? new MemorySessionStorage()
   return async (ctx, next) => {
     const key = await makeKey(ctx)
     if (key == null) {
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,8 +1,11 @@
 import { Context } from './context'
 import { Middleware } from './types'
 
+export type StorageKeyFn<T extends Context> = (
+  ctx: T
+) => Promise<string | undefined>
+
 export interface Storage<T> {
-  makeKey: (ctx: Context) => Promise<string | undefined>
   getItem: (name: string) => Promise<T | undefined>
   setItem: (name: string, value: T) => Promise<void>
   deleteItem: (name: string) => Promise<void>
@@ -14,15 +17,6 @@ export class MemorySessionStorage<T> implements Storage<T> {
 
   constructor(ttl = Infinity) {
     this.ttl = ttl * 1000
-  }
-
-  async makeKey(ctx: Context): Promise<string | undefined> {
-    const fromId = ctx.callbackQuery?.from?.id ?? ctx.from?.id ?? null
-    const chatId = ctx.callbackQuery?.message?.chat.id ?? ctx.chat?.id ?? null
-    if (fromId == null || chatId == null) {
-      return undefined
-    }
-    return `${fromId}:${chatId}`
   }
 
   async getItem(name: string): Promise<T | undefined> {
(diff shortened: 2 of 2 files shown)


src/session.ts
```diff
@@ -1,29 +1,71 @@
...
 
-export function session<SessionData extends object>({
-  ttl = Infinity,
-  store = new Map<string, { expires: number; session: SessionData }>(),
-  getSessionKey = (ctx: Context) =>
-    ctx.from && ctx.chat && `${ctx.from.id}:${ctx.chat.id}`,
-} = {}): Middleware.ExtFn<Context, { session?: SessionData }> {
-  const ttlMs = ttl * 1000
+export interface Storage<T> {
+  makeKey: (ctx: Context) => Promise<string | undefined>
```
septs, line 5: e.q:
user has external user id and real business system id remapping
it need can async

src/session.ts
```diff
@@ -1,29 +1,66 @@
...
+  private readonly ttl: number
+  private readonly store = new Map<string, { session: T; expires: number }>()
 
+  constructor(ttl = Infinity) {
+    this.ttl = ttl * 1000
+  }
+
+  async makeKey(ctx: Context): Promise<string | undefined> {
+    const fromId = ctx.callbackQuery?.from?.id ?? ctx.from?.id ?? null
+    const chatId = ctx.callbackQuery?.message?.chat.id ?? ctx.chat?.id ?? null
```
wojpawlik, line 21:
```suggestion
    const fromId = ctx.from?.id
    const chatId = ctx.chat?.id
```

src/session.ts
```diff
@@ -1,29 +1,66 @@
...
 
-export function session<SessionData extends object>({
-  ttl = Infinity,
-  store = new Map<string, { expires: number; session: SessionData }>(),
-  getSessionKey = (ctx: Context) =>
-    ctx.from && ctx.chat && `${ctx.from.id}:${ctx.chat.id}`,
-} = {}): Middleware.ExtFn<Context, { session?: SessionData }> {
-  const ttlMs = ttl * 1000
+export interface Storage<T> {
+  makeKey: (ctx: Context) => Promise<string | undefined>
```
wojpawlik, line 5: I believe this should be decoupled from `Storage`

wojpawlik (dismissed):
What's wrong with https://www.npmjs.com/search?q=telegraf-session?

src/session.ts
septs: why, not consider callback button?

septs:
> What's wrong with https://www.npmjs.com/search?q=telegraf-session?

why not can built-in?

src/session.ts
wojpawlik: https://github.com/telegraf/telegraf/blob/ab49c1a6c419416bc57780d15d458364aba5d832/src/context.ts#L145 https://github.com/telegraf/telegraf/blob/ab49c1a6c419416bc57780d15d458364aba5d832/src/context.ts#L155

septs: OK

src/session.ts
```diff
@@ -1,29 +1,77 @@
...
+    this.store.set(name, { session: value, expires: now + this.ttl })
+  }
+
+  async deleteItem(name: string): Promise<void> {
+    this.store.delete(name)
+  }
+}
+
+/**
+ * session middleware
```
septs, line 44: image

src/session.ts
```diff
@@ -1,29 +1,77 @@
...
+ *
+ * more 3rd-party session middlware
+ *
+ * see https://npm.im/search?q=telegraf-session
+ */
+export function session<SessionData extends object>(
+  storage: Storage<SessionData> = new MemorySessionStorage<SessionData>(),
+  makeKey: StorageKeyFn<Context> = makeDefaultKey
```
wojpawlik, line 52: I think these should be swapped
```suggestion
  makeKey: StorageKeyFn<Context> = makeDefaultKey,
  storage: Storage<SessionData> = new MemorySessionStorage<SessionData>()
```

or even better, put into an object.

docs/README.md
```diff
@@ -485,6 +485,15 @@ const session = require('telegraf/session')
 
 const bot = new Telegraf(process.env.BOT_TOKEN)
 bot.use(session())
+// customize session make key algorithm
```
septs, line 488: @wojpawlik

src/session.ts
```diff
@@ -1,29 +1,77 @@
...
+ * session middleware
+ *
+ * more 3rd-party session middlware
+ *
+ * see https://npm.im/search?q=telegraf-session
+ */
+export function session<SessionData extends object>(
+  makeKey: StorageKeyFn<Context> = makeDefaultKey,
+  storage: Storage<SessionData> = new MemorySessionStorage<SessionData>()
+): Middleware.ExtFn<Context, { session?: SessionData }> {
```
KnorpelSenf, line 24: This ordering is definitely better than `storage, makeKey`. However, I think it would make sense to stick with the object version we've had before. That way, every user can choose what's best for them.

src/session.ts
```diff
@@ -1,29 +1,77 @@
...
+  return `${fromId}:${chatId}`
+}
+
+export class MemorySessionStorage<T> implements Storage<T> {
+  private readonly ttl: number
+  private readonly store = new Map<string, { session: T; expires: number }>()
+
+  constructor(ttl = Infinity) {
+    this.ttl = ttl * 1000
+  }
```
KnorpelSenf, line 56: Is there a reason to put this into `MemorySessionStorage` only?

If you want to connect to MongoDB instead, it could still be very useful to store a `ttl`. So why not just keep it in the argument object of `session` and enable `ttl` for all storage types? :)

KnorpelSenf (changes requested):
I was sceptical about integrating a default session implementation before I looked at your work, but this is a really great PR! :tada:

Generally, there are just so many options for storage that I could not have come up with such a helpful yet surprisingly straightforward solution. Really cool stuff.

I suggest to improve two minor things, nothing severe:

KnorpelSenf:
Btw this will conflict with #1194 but it should be trivial to resolve. I can do it if we merge this first.

src/session.ts
septs: It should be decided by storage,
e.q: redis built-in `expire` command

see https://redis.io/commands/expire

septs: Different storage has built-in TTL.

KnorpelSenf: I see, that makes sense.

src/session.ts
KnorpelSenf, line 24:
```suggestion
export function session<SessionData extends object>(opts: {
  makeKey?: StorageKeyFn<Context>,
  storage?: Storage<SessionData>
} = {}): Middleware.ExtFn<Context, { session?: SessionData }> {
  opts = {
    makeKey: makeDefaultKey,
    storage: new MemorySessionStorage<SessionData>(),
    ...opts
  }
```

KnorpelSenf (approved):

KnorpelSenf (approved):

PR #1231, opened 2020-12-10
Fix missing exports

KnorpelSenf:
Fixes missing exports introduced after #1227


src/index.ts
```diff
@@ -4,7 +4,7 @@ export { Context } from './context'
 export * as Markup from './markup'
 export { Middleware } from './types'
 export { Router } from './router'
-export { session } from './session'
+export { session, MemorySessionStorage, Storage } from './session'
```
septs, line 7: why not ?
```suggestion
export * from './session'
```

KnorpelSenf:
@septs should we rename `makeKey` to `getSessionKey` and `storage` to `store` so #1227 is less breaking?

septs:
v4 milestone, itself break changes.

septs:
or provide `src/middleware/*`, new design built-in middleware.
(e.q: general session, command parser, etc)

septs:
Counterexample:
https://github.com/telegraf/telegraf-i18n/blame/54ba79c24f314189071f27b222192c504d3df946/readme.md#L50

KnorpelSenf:
> provide new design built-in middleware

I know that this would make `telegraf` easier to use, but it is very hard to draw the line between what should be included and what not.

I like that there is a relatively easy way to install new middleware by adding new packages. This keeps everything lightweight while still offering a lot of functionality. It also keeps things more modular.

src/index.ts
```diff
@@ -4,7 +4,7 @@ export { Context } from './context'
 export * as Markup from './markup'
 export { Middleware } from './types'
 export { Router } from './router'
-export { session } from './session'
+export * from './session'
```
wojpawlik, line 7: Same as with #1194, I don't like the idea of overwhelming new users by exporting everything directly from entry point. I'd rather have `session`, `interface session.Storage` or `interface session.SessionStorage`, and `class session.MemorySessionStorage` or `class session.MemoryStorage`.

I don't love `namespace`, it's just the only tool I know of that allows to accomplish this. Sadly, it leads to https://github.com/telegraf/telegraf/pull/1134#issuecomment-707634828.

KnorpelSenf: I agree completely that we shouldn't overwhelm new users, but I don't think that a large number of exports is too much of an issue here. I assume that most users either copy their code from the examples, or use their editor to automatically resolve them. I don't think many people will type `import {     } from 'telegraf'`, open autocomplete, and see what's available.

wojpawlik (approved):
It looks like namespacing everything would take a considerable effort... so fuck it, we'll fix it in 5.0 if it gets out of hand :)

PR #1216, opened 2020-12-06
fix #814

septs:
# Description

Use `node-fetch` extension definition.
automate select HTTP(S) Agent

```ts
    // node-fetch extensions to the whatwg/fetch spec
    agent?: Agent | ((parsedUrl: URL) => Agent);
```

see  https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a29ba12c44624c16d68da6acfb267cdfcdbd700/types/node-fetch/index.d.ts#L34-L35

Fixes #814

## Type of change

Please delete options that are not relevant.

- [ ] Documentation (typos, code examples or any documentation update)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Node.js Version:
* Operating System:

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

Issue #814, opened 2019-11-29
Any way to send image from HTTP (not HTTPS)?


y1n:
Hello! Sorry for a newbie question, but I have been stuck with it for at least a few hours now. 

Is there any shortcut provided by Telegraf to be able to send pictures from HTTP remote host? In particular, I would like to embed them into **sendMediaGroup** : )

y1n:
```
                for await (image_page_url of gallery_image_array) {
                    let image_page_body = await axios.get(image_page_url).then(res1 => res1.data);
                    let image_url = image_page_body.match(/<img id="img" src="(.*?)" style="(.*?)" onerror="(.*?)">/s)[0].replace(/<img id="img" src="/, '').replace(/" style="(.*?)" onerror="(.*?)">/, '').trim()
                    image_array.push({'type': 'photo', 'media': {url: image_url, filename: '1.jpg'}});
                }
await api.sendMediaGroup(-1001366696396, image_array)
                .then(data => {
                    console.log(data)
                })
                .catch(error => {
                    console.error(error)
                })
```
`TypeError [ERR_INVALID_PROTOCOL]: Protocol "http:" not supported. Expected "https:"`
![image](https://user-images.githubusercontent.com/13171326/69841510-24b59480-1268-11ea-8c0f-3e0fc835aa41.png)
I think its caused by the fact that the array is a mix of http and https URLs. However, maybe I just did something else wrong.

twoperkg:
Hi! I had the same issue.  Telefgraf.js is using node-fetch module to fetch data and by default node-fetch  not allowing requests by http protocol. 
Just for testing purposes swap telegraf module with my  fork git@github.com:twoperkg/telegraf.git  and see if it will work for you. 

Or reference this file:
https://github.com/twoperkg/telegraf/blob/develop/core/network/client.js
rows 12-27 and 179 where I pass options to  agent property.

septs:
workaround
```ts
const bot = new Telegraf(process.env.BOT_TOKEN!, {
  telegram: {
    // ...
    // @ts-ignore
    agent(parsedURL: URL) {
      if (parsedURL.protocol !== 'https:') {
        return http.globalAgent;
      }
      return new https.Agent({
        keepAlive: true,
        keepAliveMsecs: 10000,
      });
    },
    // ...
  },
});
```

PR #1232, opened 2020-12-10
Introduce `attachmentAgent` option

### LINKED ISSUE #1222: Cannot specify agent when attaching media via URL
Whenever `telegraf` fetches media before sending it, it is impossible to set an HTTP agent. The implementation for supporting this exists in #1232, but it is not going to be merged ([discussion](https://github.com/telegraf/telegraf/pull/1232#discussion_r540372299)).

Confer https://github.com/telegraf/telegraf/commit/863e8d9ba0e733e87091e831f2c0df191824cbb1#commitcomment-44914763 on the commit 863e8d9ba0e733e87091e831f2c0df191824cbb1.

This issue exists to inform about this.

We're open to suggestions on how to implement it well. Continue using a version prior to 4.0 in the meantime.

May be relevant in use cases like the one described in #1201.

KnorpelSenf:
# Description

Lets users supply a custom `agent` that is only used for fetching files.

Fixes #1222

## Type of change

- [X] Documentation (typos, code examples or any documentation update)
- [X] This change requires a documentation update

# How Has This Been Tested?

Untested, but compiles and lints.

# Checklist:


wojpawlik:
This doesn't allow to specify both http and https agent, so it's #814 all over again...

What I do like, is that it leaves `agent` options as-is, which means this can be fixed after 4.0, after getting feedback from people. Removing from milestone.

KnorpelSenf:
> This doesn't allow to specify both http and https agent, so it's #814 all over again...

The current default is `attachmentAgent: undefined`, which allows both protocols. Hence, #814 remains resolved the same way it was done by 863e8d9ba0e733e87091e831f2c0df191824cbb1.

#814 is only an issue again if one specifies a custom agent, but we cannot do much about this then. If a user installs an agent that blocks some kind of requests, and we use that agent, the user has to expect that those very requests ain't gonna work.

wojpawlik:
I'd rather not give people footguns, let's wait.

src/core/network/client.ts
```diff
@@ -4,10 +4,10 @@ import * as fs from 'fs'
 import * as http from 'http'
 import * as https from 'https'
 import * as path from 'path'
-import { compactOptions } from '../helpers/compact'
 import fetch, { RequestInfo, RequestInit } from 'node-fetch'
 import { hasProp, hasPropType } from '../helpers/check'
 import { Opts, Telegram } from '../../telegram-types'
+import { compactOptions } from '../helpers/compact'
```
wojpawlik, line 10: Doesn't moving it add a warning?

src/core/network/client.ts
```diff
@@ -31,8 +31,10 @@ const WEBHOOK_BLACKLIST = [
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 namespace ApiClient {
+  export type Agent = RequestInit['agent']
   export interface Options {
-    agent?: https.Agent | http.Agent
+    agent?: Agent
+    attachmentAgent?: Agent
```
wojpawlik, line 37: On second thought, we could force people to confront the issue head-on, as @septs suggested in https://github.com/telegraf/telegraf/pull/1216/files#diff-730d79a9ad7b5cb013c7df3ec05108beb986853d97a329c213700e8f2aab83c6R40:
```suggestion
    attachmentAgent?: (parsedURL: URL) => http.Agent
```

wojpawlik (changes requested):

src/core/network/client.ts
wojpawlik: But then, some people might naively create new agent per request, which defeats the point...

KnorpelSenf: I don't like using a factory here, that's not too elegant.

Currently, it is possible to set the correct agent by assigning it to `bot.telegram.options.attachmentAgent` before making a corresponding call. One way would be to document this explicitly in the docs, and point users to the issue of different protocols when introducing it.

Another way would be to add the good old router, but that's probably overdoing it and almost as bad as regenerating the agent per request:
```ts
<A extends Record<string, Agent>>setAttachmentAgents(routeFn: (parsed: URL) => keyof A, agents: A): void
```

We could also provide two options, one for HTTP and one for HTTPS, but that's also really weird, because there's some many more protocol that we would then somehow ignore.

I think I actually favor a good explanation in the docs here. If someone messes around with proxies, they have to know what they're doing. I'm

wojpawlik: > Currently, it is possible to set the correct agent by assigning it to `bot.telegram.options.attachmentAgent`

That's bad. Instead, we could introduce `attachmentAgent` option to `callApi` next to `signal`... but this still wouldn't solve the mixed protocol media group issue...

Let's just wait until it becomes an issue and someone more experienced offers a good solution.

septs: what is good solution?

septs: @wojpawlik if you not accept, you should provide "good solution"

You reject it without solving it

septs:
This problem may not be solved

src/core/network/client.ts
KnorpelSenf: @wojpawlik how about we remove `attachmentAgent` from the options one may supply, but leave it supported in `ApiClient`? We leave it undocumented, but users who have to have a proxy can still rely on `telegraf`.

I would be uncomfortable with not providing any solution whatsoever for this in 4.0, hence effectively excluding every user from sending files that has to use a proxy for some reason.

wojpawlik: > excluding every user from sending files that has to use a proxy for some reason

They're free to suggest a solution or continue using 3.38.0.

KnorpelSenf: So then what's the answer to the suggestion I just gave you?

wojpawlik: The answer is, types are documentation, so I fail to see how you could leave it in, really undocumented.

KnorpelSenf: By only allowing a direct assignment to `bot.telegram.options.attachmentAgent` but not supplying this option during setup. This is not 100 % undocumented ofc, but I assume that no one really modifies the internal options at runtime anyway.

And yes, if someone sets arbitrary values without outside of the public API that are four levels deep into the library, then I find it OK that this can break things :)

src/core/network/client.ts
KnorpelSenf: Quite the contrary, it removes a warning.

`Expected 'multiple' syntax before 'single' syntax. eslint(sort-imports)`

src/core/network/client.ts
wojpawlik: \* `ctx.telegram.options.attachmentAgent`

`Telegram::options` is `public`, so it'd still be documented and impossible to remove before 5.0.

KnorpelSenf: Yeah but that's exactly what I'm suggesting … we can literally just `Omit` that one property in the place where the options can be passed. Then no one can do it, only afterwards by explicit assignments (cause that's where we don't `Omit` it).

I don't really see where you're going tbh. We either have to explain to people that modify this option accidentally (if that's even possible) why their bot broke, or we have to reintroduce this feature at a later point anyway and then ship a bugfix release so 4.0 is usable behind the GFW.

If all you're concerned about is just dealing with the issues opened by devs who accidentally write
`ctx.telegram.options.attachmentAgent = new https.Agent({ keepAlive: true, keepAliveMsecs: 10000 })` and don't know what this does, then I can offer to take care of them.

septs: try fetching check error name is `ERR_INVALID_PROTOCOL`, wrap a new error (provide an faq documentation link)
`attachmentAgent` need is factory function. (may be)
https://github.com/telegraf/telegraf/blob/28a893e6e4157731b2e2081fca191790090e3ea4/src/core/network/client.ts#L213-L221

wojpawlik: @KnorpelSenf I'm concerned about inventing a better solution later and still having to support the poor workaround.

> we have to reintroduce this feature at a later point anyway and then ship a bugfix release so 4.0 is usable behind the GFW

Yes, I'm fine with this.

wojpawlik: Dealing with errors is messy (= not well-typed) @septs, I'd rather not.

wojpawlik:
On 2nd thought, I think the best solution is to pass-thru `agent` as-is, and document the 2 gotchas (http(s)-specific agents, and recreating agent every request), I'm taking over.

wojpawlik (approved):

PR #1233, opened 2020-12-11
Improve `Polling`

KnorpelSenf:
Fixes the bad practices introduced by #1224

Why is it even that instances of `Polling` may not be reused?

### DIFF
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -14,7 +14,7 @@ const noop = always(Promise.resolve())
 export class Polling {
   private readonly abortController = new AbortController()
   private offset = 0
-  public skipOffsetSync = false
+  private skipOffsetSync = false
   constructor(
     private readonly telegram: ApiClient,
     private readonly allowedUpdates: readonly tt.UpdateType[]
@@ -49,6 +49,14 @@ export class Polling {
           await wait(retryAfter * 1000)
           continue
         }
+        if (
+          err instanceof TelegramError &&
+          // Unauthorized      Conflict
+          (err.code === 401 || err.code === 409)
+        ) {
+          this.skipOffsetSync = true
+          throw err
+        }
         throw err
       }
     } while (!this.abortController.signal.aborted)
@@ -57,9 +65,7 @@ export class Polling {
   private async syncUpdateOffset() {
     if (this.skipOffsetSync) return
     debug('Syncing update offset...')
-    await this.telegram
-      .callApi('getUpdates', { offset: this.offset, limit: 1 })
-      .catch(noop)
+    await this.telegram.callApi('getUpdates', { offset: this.offset, limit: 1 })
   }
 
   async loop(handleUpdates: (updates: tt.Update[]) => Promise<void>) {
@@ -70,20 +76,11 @@ export class Polling {
       for await (const updates of this) {
         await handleUpdates(updates)
       }
-      await this.syncUpdateOffset()
-    } catch (err: unknown) {
-      if (
-        !(err instanceof TelegramError) ||
-        // Unauthorized      Conflict
-        (err.code !== 401 && err.code !== 409)
-      ) {
-        await this.syncUpdateOffset()
-      }
-      throw err
     } finally {
       debug('Long polling stopped')
       // prevent instance reuse
       this.stop()
+      await this.syncUpdateOffset().catch(noop)
     }
   }
 
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -17,7 +17,7 @@ export class Polling {
   constructor(
     private readonly telegram: ApiClient,
     private readonly allowedUpdates: readonly tt.UpdateType[],
-    private readonly skipOffsetSync: boolean
+    private readonly skipOffsetSync = false
   ) {}
 
   private async *[Symbol.asyncIterator]() {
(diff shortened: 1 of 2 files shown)


src/core/network/polling.ts
```diff
@@ -48,34 +48,40 @@ export class Polling {
...
+      if (
+        !(err instanceof TelegramError) ||
+        // Unauthorized    Conflict
+        (err.code !== 401 && err.code !== 409)
+      ) {
+        await this.syncUpdateOffset()
+      }
+      throw err
     } finally {
       debug('Long polling stopped')
```
wojpawlik, line 82: Perhaps a nitpick, but in my version this intentionally runs before `await this.syncUpdateOffset()`, whereas here it runs afterwards. Not asking to change it, I believe that would require adding another layer of `try finally`, just wondering if the message isn't inaccurate.

wojpawlik (changes requested):
I still prefer simpler `try finally`, but I actually wouldn't oppose introducing a boolean property (`sync`/`skipSync`?) if you insist.

> Why is it even that instances of `Polling` may not be reused?

For simplicity, I don't need to consider `abortController` being possibly reassigned.

src/core/network/polling.ts
KnorpelSenf: It briefly crossed my mind, but I didn't give too much thought to it. Now that I think about it, I can't really come up with any improvement …

KnorpelSenf:
> I still prefer simpler `try finally`, but I actually wouldn't oppose introducing a boolean property (`sync`/`skipSync`?) if you insist.

I'm not sure what this would do.

The main point why I was unhappy with #1224 was that the object rewrites its own behaviour during runtime. This may not be an issue as of right now, but it's basically preprogramming bugs for whenever anyone else but us touches the code.

I'm completely fine with the rest.

wojpawlik:
```ts
  private async syncUpdateOffset() {
    if (this.skipSync) return
    // or
    if (!this.sync) return
    debug('Syncing update offset...')
```

Names are only a suggestion; naming things is hard so I'm trying to help.

KnorpelSenf:
> ```ts
>   private async syncUpdateOffset() {
>     if (this.skipSync) return
>     // or
>     if (!this.sync) return
>     debug('Syncing update offset...')
> ```
> 
> Names are only a suggestion; naming things is hard so I'm trying to help.

Ahh thanks, but now I get what you meant by optionally skipping the sync. Yes, such an option would be nice. Does this belong into this PR, or shall we open a new one?

wojpawlik:
Do it here, let's not split the conversation further.

src/core/network/polling.ts
```diff
@@ -54,10 +55,12 @@ export class Polling {
   }
 
   private async syncUpdateOffset() {
-    debug('Syncing update offset...')
-    await this.telegram
-      .callApi('getUpdates', { offset: this.offset, limit: 1 })
-      .catch(noop)
+    if (!this.skipOffsetSync) {
```
wojpawlik, line 58: Minor thing, I won't insist, but I'd prefer avoiding nesting by doing:
```suggestion
    if (this.skipOffsetSync) return
```

src/core/network/polling.ts
```diff
@@ -16,7 +16,8 @@ export class Polling {
   private offset = 0
   constructor(
     private readonly telegram: ApiClient,
-    private readonly allowedUpdates: readonly tt.UpdateType[]
+    private readonly allowedUpdates: readonly tt.UpdateType[],
+    private readonly skipOffsetSync: boolean
```
wojpawlik, line 20: No no. Declare it outside constructor and default it to false, and change it at runtime if needed!

wojpawlik (changes requested):

src/core/network/polling.ts
KnorpelSenf: I added a default, but there is literally no difference between
```ts
class A {
  constructor(private x = 0) {}
}
```
and
```ts
class B {
  private y;
  constructor(y = 0) {
    this.y = y
  }
}
```
and
```ts
class C {
  private z = 0;
  constructor(z?: number) {
    if (z !== undefined) {
      this.z = z;
    }
  }
}
```
The first two generate JS that is identical on the byte level and the third one behaves the same way as the first two.

Any style preferences I should be aware of?

wojpawlik: I mean, it should not be possible to construct `Polling` with `skipOffsetSync` set to false, it should only be adjusted during the life of the object.

KnorpelSenf: This basically means that the feature cannot be used at all, unless we
1. either introduce a new setter method to `telegraf.ts` that allows changing the value, or
2. change `polling` in `telegraf.ts` from private to public.

wojpawlik: Yes, I think this feature should be only used internally, to prevent syncing which will make things worse or definitely fail.

wojpawlik:
Lemme take over for a moment

src/core/network/polling.ts
```diff
@@ -14,6 +14,7 @@ const noop = always(Promise.resolve())
 export class Polling {
   private readonly abortController = new AbortController()
   private offset = 0
+  public skipOffsetSync = false
```
KnorpelSenf, line 17: It is fine that this is `public` because it still cannot be reached from the outside, as `Polling` is never exposed.

It cannot be made `private` becuase `Member 'skipOffsetSync' is never reassigned; mark it as `readonly`. (eslint@typescript-eslint/prefer-readonly)` which makes sense, but making it `readonly` would defeat the point.

wojpawlik (approved):

PR #1237, opened 2020-12-12
Local session - Avoid overriding the middleware's result with an empty object

schallerala:
# Description

1. Summary of the change: small change of the async call of the next middleware in the local session middleware. The next/inner middleware might want to return some data, therefore, forward a potential resolution data.
2. Motivation: In the previous release, it was worse, as it returned the result of `Map#set` call. Anyway, it's maybe the best, but when I tried testing my code, in some cases, it seemed good to test the promise resolution data.

3. Fixes issue: _none created, none related found_

## Type of change

  - _Kind of_

# How Has This Been Tested?

Did run the AVA tests already present in the project. Can't imagine a world where I broke anything with this change.

**Test Configuration**:
* Node.js Version: _v15.2.1_
* Operating System: _macOS Big Sur, version 11.0.1_

# Checklist:

  - Not sure which warnings you are referring to.
  - Should I add a test for this pretty special use case?

PS. Can't wait to get my hands on the next release. Nice work! 👍🏼

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -31,12 +31,13 @@ export function session<SessionData extends object>(
     }
     const entry = await storage.getItem(key)
     const ctx2 = Object.assign(ctx, { session: entry })
-    await next(ctx2)
+    const nextResult = await next(ctx2)
     if (ctx2.session == null) {
       await storage.deleteItem(key)
     } else {
       await storage.setItem(key, ctx2.session)
     }
+    return nextResult
   }
 }
 


KnorpelSenf (approved):
If there are any bots out there that rely on this, then the developers clearly did not understand how `telegraf`'s middleware system _should_ be used.

However, I do agree that this change makes sense, as it lets the session middleware act as if it wasn't present at all (apart from providing session data), which is pleasant.

wojpawlik:
Middleware is not designed to return useful value (https://github.com/telegraf/telegraf/pull/1156#pullrequestreview-508746149), and it will lead to tight coupling when abused.

Have you actually tested this on `develop`? `Composer.compose` now swallows the value anyway to return `Promise<void>`: https://github.com/telegraf/telegraf/blob/fa13854a356f8d2faf617d2372ed32e9394e840a/src/composer.ts#L780-L782

KnorpelSenf:
> but on `develop`, `Composer.compose` swallows it to return `Promise<void>`

Good point, I did not consider this.

As said above, it's clearly the wrong way to use `telegraf`, no matter whether using version 3.x or 4.x. But given that `await next()` will __always__ return `undefined` starting in 4.0 (`next` returns `void`), this PR does not make as much sense as I thought it would. (It also would not break anything, though.)

Remember that passing values from one middleware to the next should always happen via `ctx`. New properties can be declared on the context object, and they will be available to subsequent middleware.

@schallerala I saw that this is your first contribution to `telegraf`, thank you for this PR. Don't let me closing this discourage you, I'm looking forward to more ideas :)

PR #1239, opened 2020-12-13
Enable `noUncheckedIndexedAccess`

wojpawlik:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Node.js Version:
* Operating System:

# Checklist:

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -760,7 +760,7 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
       return Composer.passThru()
     }
     if (middlewares.length === 1) {
-      return Composer.unwrap(middlewares[0])
+      return Composer.unwrap(middlewares[0]!)
     }
     return (ctx, next) => {
       let index = -1
@@ -773,10 +773,7 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
           throw new Error('next() called multiple times')
         }
         index = i
-        const handler = middlewares[i] ? Composer.unwrap(middlewares[i]) : next
-        if (!handler) {
-          return
-        }
+        const handler = Composer.unwrap(middlewares[i] ?? next)
         await handler(context, async (ctx = context) => {
           await execute(i + 1, ctx)
         })
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "pretest": "npm run build",
     "test": "ava test/*",
     "lint": "eslint --cache .",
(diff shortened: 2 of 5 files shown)


src/composer.ts
```diff
@@ -773,10 +773,7 @@ export class Composer<C extends Context> implements Middleware.Obj<C> {
           throw new Error('next() called multiple times')
         }
         index = i
-        const handler = middlewares[i] ? Composer.unwrap(middlewares[i]) : next
-        if (!handler) {
-          return
-        }
+        const handler = Composer.unwrap(middlewares[i] ?? next)
```
KnorpelSenf, line 776: Why would `next` need to be passed to `Composer.unwrap`? It won't be `Middleware.Obj`, will it?

wojpawlik: It won't be `Middleware.Obj`, but it actually happens to be assignable to `Middleware.Fn`.

It's passed to `Composer.unwrap` because `middlewares[i]` (now typed as possibly `undefined`) is.

KnorpelSenf: So what's wrong with
```suggestion
        const handler = middlewares[i] === undefined ? next : Composer.unwrap(middlewares[i])
```
?

wojpawlik: TypeScript is not smart enough to be able to narrow one element of array, I'd need to introduce a `const` for `middleares[i]`

KnorpelSenf: I find two lines of code still better than doing that function call for `next`. That's confusing because it's not necessary.

I know that this is a nitpick ;)

wojpawlik:
I've just ran `npm ci`, and have no typing / linting issues on my machine... interesting...

KnorpelSenf (approved):
I'd still be happy to have https://github.com/telegraf/telegraf/pull/1239#discussion_r541912115 inserted before merging, but I won't let this block the PR.

src/composer.ts
wojpawlik: I found a better way to write this:

```ts
        const handler = middlewares[i]
        if (handler === undefined) return next()
        await Composer.unwrap(handler)(context, async (ctx = context) => {
          await execute(i + 1, ctx)
        })
```

PR #1240, opened 2020-12-14
Improve DX

KnorpelSenf:
# Description

`husky` slows down commits too much. This is nasty. We already have
- CI
- PRs
- dev branch

so running tests and linting yet another time does not help in any way.

Note that people who like it can still
```
echo -e '#!/bin/bash\n\nnpm t\nnpm run lint' > .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
```
if they really want to.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,11 +38,6 @@
     "refresh": "npm run clean && npm ci",
     "clean": "git clean -fX .eslintcache docs/build/ lib/ typings/"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm test && npm run lint"
-    }
-  },
   "type": "commonjs",
   "engines": {
     "node": ">=12"
@@ -74,7 +69,6 @@
     "eslint-plugin-prettier": "^3.1.4",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
-    "husky": "^4.3.0",
     "mjs-entry": "gist:de6257751f54b3c66319bae8d2a8aea0",
     "prettier": "2.2.0",
     "typedoc": "^0.19.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,11 +38,6 @@
     "refresh": "npm run clean && npm ci",
     "clean": "git clean -fX .eslintcache docs/build/ lib/ typings/"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm test && npm run lint"
```
wojpawlik, line 43: I'll change it to just `npm run build`, and if that turns out to be not enough, I'll merge this, how about that?

wojpawlik (commented):
Would you be fine if I introduced `npm run checks` or something, performing `npm test && npm run lint`?

package.json
wojpawlik: Or perhaps `tsc --build docs/examples`, to avoid the extra `npm` process

KnorpelSenf: It's certainly going to improve it for other contributors, so that's a good idea.

I won't reinstall the hooks locally for me, though.

KnorpelSenf:
> Would you be fine if I introduced `npm run checks` or something, performing `npm test && npm run lint`?

What is the benefit?

package.json
wojpawlik: `tsc --build docs/examples` is faster by almost .8s for me, so I'll go with that

wojpawlik:
To have an easy way to gain confidence in my changes whenever needed.

package.json
wojpawlik: [Linting proved unreliable](https://github.com/telegraf/telegraf/pull/1239#issuecomment-743999803), and testing requires building first anyway.

KnorpelSenf:
> To have an easy way to gain confidence in my changes whenever needed.

I see. I like that idea

wojpawlik:
I now agree that Husky is useless. For PRs, we have CI, for pushing to `develop` we have `npm run checks`. I'll remove it.

Issue #1241, opened 2020-12-15
Switching from one WizardScene to another causes infinite loop.


brookmg:
## Context

If one tries to switch to another wizard from `wizard.leave( ctx => { ... } )`, it will cause an infinite loop that causes `RangeError: Maximum call stack size exceeded`. 

* Telegraf.js Version: 3.36.0
* Node.js Version: v14.15.1
* Operating System: Windows 10 x64 20H2

## Expected Behavior

The `ctx.scene.enter( sceneId )` function shouldn't assume the action of the user and call `ctx.scene.leave()`. Atleast there should be an option to block this trigger to avoid infinite loop.

## Current Behavior

If one tries to switch to another wizard from `wizard.leave( ctx => { ... } )`, it will cause an infinite loop that causes `RangeError: Maximum call stack size exceeded`. 

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Enter a specific WizardScene
2. Simulate moving to another WizardScene when the user leaves the current one ( using the `leave` callback method )
3. This should crash the module

### Failure Logs

RangeError: Maximum call stack size exceeded

KnorpelSenf:
Could you retry from source (`npm i telegraf/telegraf`) or at least with the most recent version 3.38.0? I'm not familiar with the code base from back then, and there have been a lot of changes already. It is possible that your problem is already solved.

brookmg:
Sure I will do that. But I feel like this was a design problem. Give me a sec

brookmg:
https://github.com/brookmg/TelegrafBotSample Check out this sample repo and see the instruction I placed on index.js. The problem is still present on 3.38.0 but like I said it's more of a design decision than a bug. Just wish if there was a way to disable the default operation

KnorpelSenf:
I see, thank you! That helped a lot.

Did you try passing `silent = true` to the call in [this line](https://github.com/brookmg/TelegrafBotSample/blob/b82e914435436bcc9e8f369f4705b001b2d1fa24/index.js#L30)?

```ts
  ctx.scene.enter('scene_2', {}, true)
```

brookmg:
I just did and it doesn't call the `Scene.enter` callback for the newly entered scene.

KnorpelSenf:
Damn it, I missed that, you're right. Then I see now way around, this needs to be fixed (preferably automatically and without requiring settings the correct flags).

Thanks for reporting this.

KnorpelSenf:
[...]

PR #1243, opened 2020-12-17
Remove `silent` flag from `SceneContextScene::enter`

wojpawlik:
# Description

There's no point to have enter and leave hooks if caller can skip calling them anytime.

See https://github.com/telegraf/telegraf/issues/1241#issuecomment-747503292

## Type of change

# Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2456,7 +2456,7 @@ Scenes related context props and functions:
 ```js
 bot.on('message', (ctx) => {
   ctx.scene.state                                    // Current scene state (persistent)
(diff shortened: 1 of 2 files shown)


KnorpelSenf:
This is nonsense. Those handlers are very useful for providing a default enter message to the user, which can be skipped and replaced by something else depending on the previous stage. Removing the `silent` flag without offering an adequate substitute is missing the point.

If you're worried about breaking expectations, then we shouldn't build them up in the first place by writing zero bytes of introduction about stages.

PR #1164, opened 2020-10-25
Add Ethiopian community chat

miki-tebe:
# Description

Added Ethiopian chat community. 

## Type of change

- [x] Documentation (typos, code examples or any documentation update)

Issue #1223, opened 2020-12-08
Strange behaviour of bot after some days, send messages twice


Spianopolese:
## Context

* Telegraf.js Version: 3.38.0
* Node.js Version: 12.0.0
* Operating System: Ubuntu 16.04.6 LTS (Xenial Xerus)  "Glitch server"

## Expected Behavior

Hi,
I'm not sure if this is a bug, but I am not finding anything and i asking you for help, I tank you anyway.

I developed a bot for a group that basically reads user commands and gives a response based on the subject of the command.
The bot remains off for the whole day and it activates himself through a chron job at 19:00. Furthermore , if  certain conditions have occurred , through a chron job at  21:00 and at 00:00 it sends other messages. 
The bot works.

## Current Behavior

**After a few days**, when the bot uses the telegram method messages.sendMessage() it sends the message twice; while when using the ctx.reply  method of telegraf  it works fine.

When the bot starts to duplicate messages, the other chron jobs from 21:00 and 00:00 even if the check of the variable that should prevent sending messages is **FALSE** , it still sends messages (always and only those that use messages.sendMessage()). 

The strange thing is this: due to many console logs used in the code i understand that the bot behaves well because it does not actually go through the function that uses messages.sendMessage (the control is **FALSE**) but still sends the message anyway. 

Also when duplicating messages (always via the console log) in the function it passes only once.
The bot does not return any errors and continues to work by responding correctly via the ctx.reply.

It is not a server problem as it continues to have this behavior even locally. The only way I've found to make this work is to change the Telegram token. If I restart the process the problem does not fix until I change the Token.

I apologize in advance for my bad english and if I have not explained myself well. This is the first time something so strange happens to me and before bothering you I searched the web but I didn't find anything.
It looks like there are some messages in the "cache" (probably wrong). 
Furthermore, I specify that the **two messages that the bot sends have two different update_id; as if the function were called twice (which I exclude through the console logs).**

I thank everyone for your help and for your time

## Failure Information (for bugs)

No errors from bot or from telegram; the flow of program is correct . The console logs are printed i correct way

## Code

```js
cron.schedule("00 19 * * *", function() {
    console.log("Start Chron");
    getWord();
});

function getWord() {
    try {
        console.log("getParole function base"); // one console log is printed
        firebase.getLastWordFirebase().then(async value => {
            var messaggio = "";
            for (let i = 0; i < value[0]["parole"].length; i++) {
[...]

Issue #1182, opened 2020-11-15
Bug with bot.hears after ctx.sendMessage twice in one session


Ksedline:
## Bug with bot.hears after ctx.sendMessage

* Telegraf.js Version: 3.38.0
* Node.js Version: 14.8.0
* Operating System: Mac OS Big Sur 11.0.1

## Expected Behavior

bot.hears would listen any text sent via ctx.sendMessage with reply_markup: [{ 'keyboard': ['Key'] }]

## Current Behavior

bot.hears would listen and answer any text sent via ctx.sendMessage with reply_markup: [{ 'keyboard': ['Key'] }] only at once

## Failure Information (for bugs)

After hours of debugging I found that Composer.compose in composer.js method actually find this text anytime in context or ctx, but do nothing with that after second user's answer.

Reloading the bot helps to get an answer by bot if it stucks after second time.

### Steps to Reproduce

1. Send message to a user with 
```
await ctx.telegram.sendMessage(userToSendId, '<b>Test</b>', { parse_mode: 'HTML', reply_markup: JSON.stringify({ 'keyboard': [['Accept']] })});
```
also subscribe bot for 'Accept':
```
bot.hears('Accept', async (ctx) => await ctx.reply('Accepted'));
```
2. If you will send keyboard second time - bot.hears don't registrate 'Accept' and don't reply 'Accepted' to user

### Failure Logs
No specific logs or Error :(

KnorpelSenf:
I'm not sure if I understand this correctly … are you saying that `bot.hears` only works once, and then the bot needs to be restarted?

Ksedline:
Yes. 
You can reproduce it when trying to send message with keyboard dirrectly from bot ctx.telegram.sendMessage

KnorpelSenf:
I did not manage to do reproduce that, and I have several bots running in production that are not suffering from this.

Could you maybe share an [SSCCE](http://sscce.org/)?

Ksedline:
Sure, I do it later in this issue

Ksedline:
I talk with my co-developer and we can't reproduce that bug. Sorry for wasted time

KnorpelSenf:
No problem, I'm glad it's solved :)

Hisahito:
I have a similar problem with the bot.hears method: 
bot.hears('Hello', (ctx) => { 
    ctx.message("Alola and Welcome, its good to see you here!" + ctx.from.first_name)
   })

[...]

PR #1244, opened 2020-12-17
Improve issue templates

wojpawlik:
Closes #1096
Closes #1164

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -6,32 +6,42 @@ about: Create a report to help us improve
 
 ## Context
 
+<!--
+Check if the issue wasn't already fixed: `npm install --no-save telegraf/telegraf`
+
 Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.
+-->
 
 * Telegraf.js Version:
 * Node.js Version:
 * Operating System:
 
+## Minimal Example Code Reproducing the Issue
+
+```ts
+// your code here. Please omit everything not related to the issue,
+// to make it easier to pinpoint the cause.
+```
+
+<!--
+How to trigger the issue?
+Please provide detailed step-by-step guide.
+-->
+
 ## Expected Behavior
 
+<!--
 Please describe the behavior you are expecting
+-->
 
 ## Current Behavior
 
+<!--
 What is the current behavior?
+-->
 
-## Failure Information (for bugs)
+## Error Message and Logs (`export DEBUG='telegraf:*'`)
 
-Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.
-
-### Steps to Reproduce
-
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -6,32 +6,42 @@ about: Create a report to help us improve
...
+```ts
+// your code here. Please omit everything not related to the issue,
+// to make it easier to pinpoint the cause.
+```
+
+<!--
+How to trigger the issue?
+Please provide detailed step-by-step guide.
+-->
+
```
KnorpelSenf, line 30: Unsure about this, you're asking for a lot here. Providing a code example is nice if we actually have a definite bug or a reproducible crash, but it is not always required. In fact, if you look at the closed issue of the past weeks, most of them could easily be closed without any code, e.g. questions.

Instead of asking less of the authors, we could just as well add another template for questions/other/general/confusion. That is far more common than true bugs, and it could well have its own place. Then we can keep this as focused as it it.

KnorpelSenf (changes requested):
Most of the changes look very good. I request one change about the bug issue template.

I am aware that the new discussions feature may be a better place to ask questions. However, just because we'd maybe like that, it is not the reality yet. We will keep on having questions in issues, so we should prepare for that.

.github/ISSUE_TEMPLATE/Bug_report.md
wojpawlik: I'm thinking about introducing a question template too, but I'm not yet sure how it should differ from the bug report template, so I was planning to do it in next pull request. Advice or help welcome.

I heard there's a way to link issue template to open a discussion instead, but I haven't looked into it yet.

> if you look at the closed issue of the past weeks, most of them could easily be closed without any code

True. On the flip side, we have several open (#1182, #1223) and a bunch of closed issues that are / were difficult to move forward because they have too little or way too much code.

I added `needs more info` label for those, btw. And we could have a bot to auto-close them if inactive.

We could also add "delete anything that's not applicable" comment, and "laziness will make it more difficult for us to help".

KnorpelSenf: > Advice or help welcome.

I'll figure something out.

> I added `needs more info` label for those

Nice :+1:

> we could have a bot to auto-close them if inactive

I don't think that's necessary for the scale of our project (unless you want to play around with GH bots, then feel free).

> We could also add "delete anything that's not applicable" comment

That sounds good!

> "laziness will make it more difficult for us to help".

That's rude, don't insult people before you talk to them. A friendly reminder could certainly help.

KnorpelSenf: > I heard there's a way to link issue template to open a discussion instead, but I haven't looked into it yet.

Confer https://github.com/eamodio/vscode-gitlens/issues/new/choose

KnorpelSenf (approved):

Issue #1234, opened 2020-12-11
Long polling silently hangs or swallows errors


wojpawlik:
1. If we don't await the middleware before fetching further updates, we break
   a) any semblance of sensible update confirming,
   b) [error handling](https://github.com/telegraf/telegraf/blob/bc18691c5c80a27796ef318d65ec431b7280d713/src/core/network/polling.ts#L74-L79) (every error turns into a rejection not handled by Telegraf),
   c) backpressure (bot will fetch updates even when overloaded),
   d) others, see https://github.com/telegraf/telegraf/issues/892#issuecomment-583765906.

2. If we do naively `await` (default behavior before #1224), bot will silently wait forever for handlers that break by never resolving (#735). This behavior is still available when `handlerTimeout === Infinity`, accessing it should probably [emit a warning](https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor).

3. In #1224, I changed the default behavior to resolve the promise if processing takes longer than 5 seconds. When phrased this way, the problem is obvious: errors thrown after 5 seconds are swallowed!

I have a suggestion, but it ain't pretty either: **throw** when processing takes longer than `handlerTimeout` (which'd need to be extended to say 30 seconds). Such a drastic change obviously needs to be decided upon before 4.0.

The real solution here would be to fetch updates via WebSocket instead, which could allow confirming on per-update basis.

In the meantime, please use webhook in production if you can (with `webhookReply` disabled: #1082).

yeldiRium:
Why resolve the promise after a timeout at all?

If a handler does not resolve, that is a bug on the side of the bot developer. I don't think telegraf should try to recover from bugs made by a bot developer.

If the handler resolves (or errors) after a long time, the timeout swallows its results.

I'm sorry to interject since I don't know the history of some linked issues. But the swallowing of errors seems to me to be the result of trying to fix other's mistakes which results in more room for error. Or am I missing something?

KnorpelSenf:
I'd say you're right about this. When the middleware simply does not return, we cannot do much about it. It's upon the user to handle such a scenario, and that's what we should make possible.

The thing I have in mind is to let the user provide a `timeoutHandler`, i.e. write something like `bot.catchTimeout`. This handler will be supplied with the respective context object that lead to the timeout, and maybe the unresolved `Promise` (not sure about that one yet).

KnorpelSenf:
Here we go.

---

[...]

PR #1247, opened 2020-12-19
Rewrite update handling

### LINKED ISSUE #735: Long Polling drops after 1 day

KnorpelSenf:
# Description

This PR implements the decaying deque described in #1234. If you are not familiar with that discussion, you may as well skip reading it and check out the comments in the code first.

If you did read through the main post, I will outline the differences here.

## No timeout pointer

In the original proposal, we stored a reference to the queue head, and we followed the queue upon timeout from the head up to the timeout pointer. However, we already know that the `date` property on the drifts are identical for nodes in a batch, so we can simply use them instead.

## No bulk

We could drop the bulk references the same way as the timeout pointer.

## Different error handling

The error handling is performed almost as described in https://github.com/telegraf/telegraf/discussions/1234#discussioncomment-217223, however, it was simpler to remove the elements right away instead of leaving them in the deque. This also has a better guarantee for when the timeout handler does not make the middleware complete—otherwise, we'd have a lot of elements in the queue that would break the timeout system. However, it should not make a difference to the outside.

## Storing queue containment

The `date` property is set to `-1` if an element is removed. This is to find out whether we should still catch errors, or rethrow them (because the drift had been removed due to timing out).

## Integration of queue and timer

Separating timeout handling and queue implementation is non-trivial. Given that the current file has 80 lines (130 when counting comments and types), and given that the queue implementation only accounts for 20-ish of them, I won't invest the time to do this. If you, however, want to do this, I'm not stopping you.

# Checklist:

I also ran the example bots in order to make sure the updates are processed correctly.

Fixes #735.
Closes #1252.

### DIFF
--- a/src/core/processing/queue.ts
+++ b/src/core/processing/queue.ts
@@ -1,0 +1,130 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+interface Drift<E> {
+  prev: Drift<E> | null
+  next: Drift<E> | null
+
+  task: Promise<void>
+
+  date: number
+  elem: E
+}
+
+export class DecayingDeque<E> {
+  private length: number = 0
+  private head: Drift<E> | null = null
+  private tail: Drift<E> | null = null
+
+  private readonly concurrency: number
+  private timer: NodeJS.Timeout | undefined
+  private subscribers: Array<(capacity: number) => void> = []
+
+  /**
+   * Creates a new decaying queue, confer
+   * @param handlerTimeout max period of time for a task
+   * @param worker task generator
+   * @param concurrency `add` will return only after the number of pending tasks fell below `concurrency`. `false` means `1`, `true` means `Infinity`, numbers below `1` mean `1`
+   * @param catchError error handler
+   * @param catchTimeout timeout handler
+   */
+  constructor(
+    private readonly handlerTimeout: number,
+    private readonly worker: (t: E) => Promise<void>,
+    concurrency: boolean | number,
+    private readonly catchError: (err: unknown, elem: E) => Promise<void>,
+    private readonly catchTimeout: (t: E, task: Promise<void>) => void
+  ) {
+    if (concurrency === false) this.concurrency = 1
+    else if (concurrency === true) this.concurrency = Infinity
+    else this.concurrency = concurrency < 1 ? 1 : concurrency
+  }
+
+  /**
+   * Adds the provided elements to the queue and starts tasks for all of them
+   * immediately. Returns a `Promise` that resolves with `concurrency - length`
+   * once this value becomes positive.
+   * @param elems elements to be added
+   * @returns `capacity - length` inside a `Promise`
+   */
+  add(elems: E[]): Promise<number> {
+    const len = elems.length
+    const capacity = this.concurrency - (this.length += len)
+
+    if (len > 0) {
+      let i = 0
+      const now = Date.now()
+      if (this.head === null) {
+        this.head = this.tail = this.toDrift(elems[i++]!, now)
+        this.startTimer()
+      }
+
+      let prev = this.tail!
+      while (i < len) {
+        const node = this.toDrift(elems[i++]!, now)
+        prev.next = node
+        node.prev = prev
+        prev = node
+      }
+      this.tail = prev
+    }
+
+    return capacity > 0
+      ? Promise.resolve(capacity)
+      : new Promise((resolve) => this.subscribers.push(resolve))
+  }
+
+  private decay(node: Drift<E>): void {
+    if (node.date !== node.next?.date)
+      if (this.head === node) {
+        if (this.timer !== undefined) clearTimeout(this.timer)
+        if (node.next === null) this.timer = undefined
+        else this.startTimer(node.next.date + this.handlerTimeout - Date.now())
+      }
+    this.remove(node)
+  }
+
+  private remove(node: Drift<E>): void {
+    if (this.head === node) this.head = node.next
+    else node.prev!.next = node.next
+    if (this.tail === node) this.tail = node.prev
+    else node.next!.prev = node.prev
+
+    node.date = -1
+    const capacity = this.concurrency - --this.length
+    if (capacity > 0) {
+      this.subscribers.forEach((resolve) => resolve(capacity))
+      this.subscribers = []
+    }
+  }
+
+  private toDrift(elem: E, date: number): Drift<E> {
+    const node: Drift<E> = {
+      prev: null,
+      task: this.worker(elem)
+        .catch(async (err) => {
+          if (node.date > 0) await this.catchError(err, elem)
+          else throw err
+        })
+        .finally(() => {
+          if (node.date > 0) this.decay(node)
+        }),
+      next: null,
+      date,
+      elem,
+    }
+    return node
+  }
+
+  private startTimer(ms = this.handlerTimeout): void {
+    if (ms < 1) setImmediate(() => this.timeout())
+    else this.timer = setTimeout(() => this.timeout(), ms).unref()
+  }
+
+  private timeout(): void {
+    while (this.head!.date === this.head!.next?.date) {
+      this.catchTimeout(this.head!.elem, this.head!.task)
+      this.remove(this.head!)
+    }
+    this.catchTimeout(this.head!.elem, this.head!.task)
+    this.decay(this.head!)
+  }
+}
--- a/src/core/processing/queue.ts
+++ b/src/core/processing/queue.ts
@@ -1,30 +1,111 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
+
+/**
+ * A drift is an element in a doubly linked list, and it stores a task. A task
+ * is represented as a Promise. Drifts remove themselves from the queue (they
+ * _decay_) after their task completes or exceeds the timeout. The timeout is
+ * the maximum time that drifts are allowed to be contained in the queue.
+ *
+ * Drifts are appended to the tail of the queue and _drift_ towards the head as
+ * older elements are removed, hence the name.
+ *
+ * The task of every drift is created by a worker function that operates based
+ * on a source element. A drift keeps a reference to that source element. In
+ * case of a timeout, the timeout handlers will be supplied with the source
+ * element because it is interesting to know for which source elements the
+ * worker function produced a promise that timed out.
+ *
+ * A drift also stores the date at which it was added to the queue.
+ *
+ * In the context of `telegraf`, each middleware invocation corresponds to a
+ * task in a drift. Drifts are used to manage concurrent middleware execution.
+ */
 interface Drift<E> {
+  /** Previous drift in the queue. `null` iff this drift is the head element. */
   prev: Drift<E> | null
+  /** Next drift in the queue. `null` iff this drift is the tail element. */
   next: Drift<E> | null
 
+  /**
+   * Task of the drift. Contains logic that removes this drift from the queue as
+   * soon as the task completes by itself (either resolves or rejects).
+   */
   task: Promise<void>
 
+  /**
+   * Timestamp (milliseconds since The Epoch) when this drift was added. This
+   * may be inspected when starting a new timer that might purge this drift upon
+   * timeout.
+   *
+   * The timestamp will be set to `-1` when the drift is removed from the queue,
+   * in other words, checking `date > 0` serves as a containment test.
+   */
   date: number
+  /** Reference to the source element that was used to start the task */
   elem: E
 }
 
+/**
+ * A _decaying deque_ is a special kind of doubly linked list that serves as a
+ * queue for a special kind of nodes, called _drifts_.
+ *
+ * A decaying deque has a worker function that spawns a task for each element
+ * that is added to the queue. This task then gets wrapped into a drift. The
+ * drifts are then the actual elements in the queue.
+ *
+ * In addition, the decaying deque runs a timer that purges old elements from
+ * the queue. This period of time is determined by the `handlerTimeout`.
+ *
+ * When a task completes or exceeds its timeout, the corresponding drift is
+ * removed from the queue. As a result, only drifts with pending tasks are
+ * contained in the queue at all times.
+ *
+ * When a tasks completes with failure (`reject`s or exceeds the timeout), the
+ * respective handler (`catchError` or `catchTimeout`) is called.
+ *
+ * The decaying deque has its name from the observation that new elements are
+ * appended to the tail, and the old elements are removed at arbitrary positions
+ * in the queue whenever a task completes, hence, the queue seems to _decay_.
+ */
 export class DecayingDeque<E> {
+  /**
+   * Number of drifts in the queue. Equivalent to the number of currently
+   * pending tasks.
+   */
   private length: number = 0
+  /** Head element (oldest), `null` iff the queue is empty */
   private head: Drift<E> | null = null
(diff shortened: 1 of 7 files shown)


src/telegraf.ts
```diff
@@ -38,9 +38,39 @@ namespace Telegraf {
...
+     * until enough of your middleware completed.
+     *
+     * `false` is an alias for `1`. `true` is an alias for `Infinity`. Values
+     * below `1` will be interpreted as `1`.
+     */
+    concurrency?: boolean | number
+    /** Configuration options for when the bot is run via long polling */
+    polling?: {
+      /** Determines the timout of the `getUpdates` calls */
+      timeout?: number
```
wojpawlik, line 72: Why is this needed?

src/telegraf.ts
```diff
@@ -142,7 +200,19 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
     return this
   }
 
+  initializeQueue(concurrency: boolean | number) {
```
wojpawlik, line 203:
```suggestion
  private initializeQueue(concurrency: boolean | number) {
```

src/telegraf.ts
```diff
@@ -185,40 +257,44 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
...
-      throw new TypeError(util.format('Updates must be an array, got', updates))
-    }
-    // prettier-ignore
-    const processAll = Promise.all(updates.map((update) => this.handleUpdate(update)))
-    if (this.options.handlerTimeout === Infinity) {
-      return processAll
-    }
-    return Promise.race([processAll, wait(this.options.handlerTimeout)])
+  async handleUpdate(update: tt.Update, webhookResponse?: http.ServerResponse) {
+    await this.handleUpdates([update], webhookResponse)
```
wojpawlik, line 261: Does this mean that each update delivered via webhook will spawn a new timer?

wojpawlik (commented):
My first pass thru your code.

I'd still be more confident if you used https://github.com/isaacs/yallist -- it's popular and it has tests.

src/telegraf.ts
```diff
@@ -38,9 +38,39 @@ namespace Telegraf {
...
+     * **In webhook mode,** it is possible to receive old updates again while
+     * new ones have already been processed, so this is not an issue. The flag
+     * behaves exactly the same way, but its default value is `true`. If you set
+     * it to a lower value, `telegraf` will simply not answer the HTTP requests
+     * until enough of your middleware completed.
+     *
+     * `false` is an alias for `1`. `true` is an alias for `Infinity`. Values
+     * below `1` will be interpreted as `1`.
+     */
+    concurrency?: boolean | number
```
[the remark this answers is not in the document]
KnorpelSenf: It was convenient when I developed the queue module, but as `telegraf.ts` replicates this now, we may as well drop it again.

I found it nice to say `concurrency: false` instead of having to think about which number to pick when avoiding `getUpdates` calls before all middleware has finished.

src/telegraf.ts
KnorpelSenf: The problem here is that we need to initialize the queue for running the tests. Given that we know webhook/long polling only upon `bot.launch`, we cannot construct the object when constructing the bot. As a result, it needs to be constructed manually by the test suite (which does not `launch` the bot).

Any suggestions how to improve this? We could look into making the concurrency value adjustable during runtime, but I'm not sure what this requires in turn.

src/telegraf.ts
KnorpelSenf: No. There is only ever one timer at most per bot, no matter how many updates are processed.

Instead, what this implies is that a timer will time out only one update in the same iteration, never batches of updates (unless they arrive in the same millisecond, that is).

src/telegraf.ts
KnorpelSenf: This determines the timeout of the `getUpdates` calls.

We [wanted](https://github.com/telegraf/telegraf/discussions/1234#discussioncomment-208597) to make all three timeouts configurable:
- `getUpdates` calls
- middleware
- fetching

Middleware timeouts can be configured by `handlerTimeout`. If I'm not mistaken, we do not support setting a timeout for fetching (yet).

KnorpelSenf:
> I'd still be more confident if you used [isaacs/yallist](https://github.com/isaacs/yallist).

I did consider using this and other existing packages, including batteries-included promise queues. I can explain in more depth why I discarded all of them as inferior to an integrated implementation of a basic linked list.

### Timers and nodes are heavily dependent on each other

Note how managing the timer requires inspecting and modifying next and previous nodes, e.g.

https://github.com/telegraf/telegraf/blob/bcc37b6d3cfc4e547ec0f459e5710ee59ef3bfde/src/core/processing/queue.ts#L80

Also note that nodes and their data are closely integrated.

https://github.com/telegraf/telegraf/blob/bcc37b6d3cfc4e547ec0f459e5710ee59ef3bfde/src/core/processing/queue.ts#L108

I did not find a single promise queue out there that does this. No matter what kind of existing list implementation we use, this would be made more difficult. As an example, we could not create nodes independently of the queue. (E.g. the list you suggested would require something like this:
- create data (context and promise)
- push it to queue
- inspect tail node (we know that this is the new node)
- modify data again to link it back to the retrieved node
- repeat for next node

Having a `toDrift` function is much simpler, and it prevents us from allowing `task` to be `undefined`.)

### Identical interfaces

Both adding and removing requires additional custom logic that is specific for decaying queues. We'd end up with the same interface as other packages provide, in other words, we fail to abstract away anything of the interface of the data structure. While it would save us from typing 15 lines or so, it does not significantly help simplifying the queue.

### Only a fraction of the implementation is required

95 % of any third-party implementation would be useless crap for us because we only need very little of it. I disagree with the trend in JavaScript development that as little as 15 lines of code should be pulled in as a dependency, especially when the purpose is debatable. (Other disciplines don't even need dependencies for graph diffing because it's just simpler to do it yourself, and we're talking about a primitive version of a linked list here …)

>  It's less reviewing now and less debugging later; it's popular and it has tests.

Less reviewing: only assuming that everyone is familiar with the exact behaviour of the code in the dependency. If you factor in that readers of our code have to study the library, it's _much_ more effort to understand what's going on.

Less debugging: only assuming that integrating with third-party code in an evolving ecosystem is cheap.

Popular: that's unrelated, `lodash` is also popular, and relying on it is a major disadvantage.

Has tests: We need tests anyway (another reason for draft state) and we probably can't help it but write tests that also cover testing the linked list.

src/telegraf.ts
wojpawlik: Not a problem, tests aren't typechecked :) Don't worry about it now, it can always be made `public` if needed.

src/telegraf.ts
wojpawlik: Then, does this mean that if 5 updates arrive, 5th one has 5 times as long timeout?

KnorpelSenf: No. Let's assume the middleware does not ever complete for this example:

Polling:
- five updates arrive
- one timer is set
- all updates will be purged together
- no timer is running

Webhooks:
- first update arrives
- one timer is set
- other updates arrive (timer is not touched)
- first update will be purged
- timer is restarted for the remaining time until the second update times out
- second update is purged
- timer is restarted for the remaining time until the third update times out
- etc
- last update is purged
- no timer is running

wojpawlik:
This is overwhelming, I'll try writing a simpler implementation in separate PR.

KnorpelSenf:
> This is overwhelming, I'll try writing a simpler implementation in separate PR.

It may make sense to wait until I add comments. I know it's not readable yet, and there's reason I haven't been requesting any reviews yet.

Also, I'm not sure if it's actually possible to simplify it much without also cutting down features.

KnorpelSenf:
Closed because of inactivity

src/core/processing/queue.ts
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  /**
+   * Adds the provided elements to the queue and starts tasks for all of them
+   * immediately. Returns a `Promise` that resolves with `concurrency - length`
+   * once this value becomes positive.
+   * @param elems elements to be added
+   * @returns `capacity - length` inside a `Promise`
+   */
+  add(elems: E[]): Promise<number> {
```
wojpawlik, line 129: 1. Wouldn't this be simpler if it accepted one element at a time?
2. It would be way simpler if you dropped all the list stuff, Node seems to use a list internally: https://stackoverflow.com/a/13616530

src/telegraf.ts
```diff
@@ -77,6 +113,20 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
     throw err
   }
 
+  private handleTimeout = (ctx: C, middleware: Promise<void>): void => {
```
wojpawlik, line 116: How about just moving `handleTimeout` and `handleError` onto `DecayingQueue` to make `initializeQueue` unnecessary?

src/telegraf.ts
```diff
@@ -77,6 +113,20 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
...
+    middleware.finally(() =>
+      debug(
+        'Middleware completed after timeout for update',
+        ctx.update.update_id
+      )
+    )
```
wojpawlik, line 122: This is a good thing #1262 can't easily achieve.

src/core/processing/queue.ts
```diff
@@ -0,0 +1,265 @@
...
+   */
+  constructor(
+    private readonly handlerTimeout: number,
+    private readonly worker: (t: E) => Promise<void>,
+    concurrency: boolean | number,
+    private readonly catchError: (err: unknown, elem: E) => Promise<void>,
+    private readonly catchTimeout: (t: E, task: Promise<void>) => void
+  ) {
+    if (concurrency === false) this.concurrency = 1
+    else if (concurrency === true) this.concurrency = Infinity
```
wojpawlik, line 118: Very unsafe, will confirm updates prematurely and fetch more even when bot is overloaded.

```suggestion
    else if (concurrency === true) this.concurrency = 100
```

src/core/processing/queue.ts
```diff
@@ -0,0 +1,265 @@
...
+  private timeout(): void {
+    // Rare cases of the event ordering might fire a timeout even though the
+    // head element has just decayed.
+    if (this.head === null) return
+    while (this.head.date === this.head.next?.date) {
+      this.catchTimeout(this.head.elem, this.head.task)
+      // No need to restart timer here, we'll modify head again anyway
+      this.remove(this.head)
+    }
+    this.catchTimeout(this.head.elem, this.head.task)
```
wojpawlik, line 262: If this throws, process crashes synchronously, no cleanup, no syncing update offset.

src/core/processing/queue.ts
wojpawlik: On 2nd thought, you cannot drop list stuff, it's needed for the custom backpressure.

Am I correct in saying that backpressure is the crucial difference from #1262? The face that `add` settles when bot is "bored", rather than when when middleware settles.

KnorpelSenf: Yes, that and the fact that the number of polled updates dynamically adjusts to the load. The higher the load, the updates are requested.

Also, errors are handled after timeout instead of being swallowed.

I do have more improvements in mind, e.g. providing a way to give updates more time upon certain kinds or errors. Might help with one bot I have that has comparably long-running middleware due to expensive operations, and would have been trivial to implement with this PR

src/telegraf.ts
KnorpelSenf: That would've been good, I think

src/core/processing/queue.ts
KnorpelSenf: My initial iteration had 100 as default, but this is a problem when the load goes beyond that value for webhooks. Then suddenly with too much load the `add` calls would be slowed down until Telegram re-sends updates, thus making deduplication necessary again. `Infinity` is the only choice where this does not happen, even though it looks odd.

src/core/processing/queue.ts
KnorpelSenf: True, there should've been another `try`-`catch` around

src/core/processing/queue.ts
KnorpelSenf: Also, we don't use timers, but `unref`ed timers. I know that timers are fast, but the docs explicitly state that one should avoid `unref`fing too many. We'd `unref` up to 1000/s or so at peak loads.

I tried to figure out in the Node.js source why this is supposed to be slow and if we can ignore this advice, but I could not find out about this. (It is not a coincidence that I used a comparable data structure than the internal one in Node, updates are handled similar to timeouts.) It may be possible to rewrite the rest without using an explicit queue (effectively leveraging the event loop as queue in a way), but as I figured that we need it anyway, I didn't investigate it at the time.

src/core/processing/queue.ts
wojpawlik: What would you do in the `catch` clause?

src/core/processing/queue.ts
wojpawlik: So, `100` is incorrect for webhooks, while `Infinity` is incorrect for polling?

src/core/processing/queue.ts
wojpawlik: On 3rd thought, all we need for "improved" backpressure is number of updates currently being processed, no need for list.

> providing a way to give updates more time upon certain kinds or errors.

Interesting, but again detrimental for polling correctness.

src/core/processing/queue.ts
KnorpelSenf: `1` is correct for polling, larger values may lead to calls to `getUpdates` before all middleware has completed. (One may choose to increase the value and decrease reliability to improve performance.)

`Infinity` is correct for webhooks, lower values may lead to duplicate updates being received. (I don't think it makes much sense to decrease the value, but perhaps someone else knows a use-case.)

src/core/processing/queue.ts
KnorpelSenf: Logging the error, or maybe something else if you have a better idea

wojpawlik: The best I can come up with is stopping polling and crashing. Inferior to #1262.

KnorpelSenf: The case that a timeout handler throws is completely independent of the implementation we choose. #1262 always calls the regular bot error handler instead, which does not provide any safety net at all. We can of course resemble that behaviour here if we wanted to, but #1262 is not even capable of distinguishing these very different errors from others. However, given that none of the two PRs have a chance of being merged in the current state, we may as well skip comparing them further. (This PR has a large diff but solves the problems, and the other one doesn't have a large diff but ignores some problems.)

src/core/processing/queue.ts
KnorpelSenf: > no need for list

That's what I meant with my comment above and with https://github.com/telegraf/telegraf/pull/1262#issuecomment-752932119, I'll investigate this approach in the coming days.

> detrimental for polling correctness

This may be a nitpick, but there's a difference between correctness and reliability. If unreliability is desired for other benefits, it may be the correct thing to do (example: UDP). (We agree that this should be opt-in and not the default.)

PR #1248, opened 2020-12-19
Prevent infinite recursion

### LINKED ISSUE #1241: Switching from one WizardScene to another causes infinite loop.

wojpawlik:
# Description

Please verify that this fixes #1241 @brookmg.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/scenes/context.ts
+++ b/src/scenes/context.ts
@@ -91,7 +91,7 @@ class SceneContextScene<C extends SceneContext> {
     if (!silent) {
       await this.leave()
     }
-    debug('Enter scene', sceneId, initialState, silent)
+    debug('Entering scene', sceneId, initialState, silent)
     this.session.current = sceneId
     this.state = initialState
     const ttl = this.current?.ttl ?? this.options.ttl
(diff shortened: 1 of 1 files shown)


brookmg:
I just tested this branch and seems like the issue is fixed 💯 
https://github.com/brookmg/TelegrafBotSample/commit/19fab1ebcf22b0d9c59e7b1759df014f06644691

wojpawlik:
Feel free to use the branch for production, at least until Telegraf 4.0.0 release.

KnorpelSenf (approved):

Issue #1167, opened 2020-10-28
Async/Await in Webhook updates


Geebrox:
Hello. I am trying to send a chain of messages using `async/await` but the order of messages coming wrongly. Messages' queue is lost.

Code:
```
bot.command('start', async (ctx) => {
  await ctx.reply('Message #1')
  await ctx.reply('Message #2')
  await ctx.reply('Message #3')
  await ctx.reply('Message #4')
  return ctx.reply('Message Last')
})
```
Result:
![image](https://user-images.githubusercontent.com/15965924/97405826-15c4df80-191a-11eb-815e-e026020dd1a9.png)

Disabling `webhookReply` solves this problem, but what does it mean, could someone explain what is `webhookReply` and what will happen if I disable it? Will it cause on bots performance?

**More info about my configs**:
Telegraf version: `^3.38.0`
Bot updates type: `Webhook`

Thank you!

Geebrox:
More testings:

Code:
```
bot.command('start', async (ctx, next) => {
  await ctx.reply('Message -> 1')
  await ctx.reply('Message -> 2')
  return next()
})

bot.command('start', async (ctx) => {
  await ctx.reply('Message -> 3')
  await ctx.reply('Message -> 4')
  return ctx.reply('Message Last')
})
```

Result:
![image](https://user-images.githubusercontent.com/15965924/97408069-6ab62500-191d-11eb-8230-03252d53b464.png)

Geebrox:
## Performance check:

### **Code:**
```
const { BotController } = require('./controllers')

const { bot } = (botController = new BotController())

bot.command('start', async (ctx, next) => {
  console.time('/start response took')
  await ctx.reply('Message -> 1')
  await ctx.reply('Message -> 2')
  return next()
})

bot.command('start', async (ctx) => {
  await ctx.reply('Message -> 3')
  await ctx.reply('Message -> 4')
  console.timeEnd('/start response took')
  return ctx.reply('Message Last')
})

botController
  .start()
  .then(() => console.log('Bot started'))
  .catch(console.error)
```

### **Results:**

**With** `webhookReply: false`:
```
[nodemon] starting `node src/index.js`
Bot started
/start response took: 497.622ms
/start response took: 508.447ms
/start response took: 483.859ms
```

Messages are received in the right order
<details>
  <summary>Show image</summary>
  
[...]

Issue #1086, opened 2020-07-16
messages with attachments in wrong order


artyom-vasich:
Hello! I was trying to send an album and then a bunch of callback buttons and noticed, that they arrive before the album. It appears that the first message with attachment in a row arrives later than expected. For example

```js
bot.command("foo", async ctx => {
    await ctx.replyWithPhoto("file_id 1");
    await ctx.reply("photo 1");
    await ctx.replyWithPhoto("file_id 2");
    await ctx.reply("photo 2");
    await ctx.replyWithPhoto("file_id 3");
    await ctx.reply("photo 3");
});
```
always results in sequence "photo 1" -> [img1] -> [img2] -> "photo 2" -> [img3] -> "photo 3"

If I log returned message objects like this
```js
bot.command("foo", async ctx => {
    const m1 = await ctx.replyWithPhoto("file_id 1");
    const m2 = await ctx.replyWithPhoto("file_id 2");
    const m3 = await ctx.replyWithPhoto("file_id 3");
    console.log(m1);
    console.log(m2);
    console.log(m3);
});
```
I see regular expected messages m2 and m3, but m1 is like that instead
```js
{
    webhook: true,
    details: 'https://core.telegram.org/bots/api#making-requests-when-getting-updates'
}
```

Am I doing something wrong? 

telegraf 3.38, and I am using webhook callbacks with fastify if it matters

**EDIT:** it turned out that attachments have nothing to do with this: first two messages always sent in reverse order regardless of content
```js
    await ctx.reply("1");
    await ctx.reply("2");
    await ctx.reply("3");
    await ctx.reply("4");
```
2 -> 1 -> 3 -> 4

wojpawlik:
I have no experience with Fastify, please show me the code you use to integrate Telegraf with it.

artyom-vasich:
Like with attachments, mention of fastify was misleading :( Just tried following minimal code

```js
const { Telegraf } = require("telegraf");

const bot = new Telegraf(process.env.BOT_TOKEN);

bot.command("start", async ctx => {
    await ctx.reply("1");
    await ctx.reply("2");
    await ctx.reply("3");
    await ctx.reply("4");
});

bot.telegram.setWebhook(`${process.env.HOST}/${process.env.BOT_TOKEN}`);
bot.startWebhook(`/${process.env.BOT_TOKEN}`, null, process.env.PORT);
```
and result remains the same:
[...]

PR #1250, opened 2020-12-20
Disable `webhookReply` by default

wojpawlik:
# Description

`webhookReply` is an [endless source of issues](https://github.com/telegraf/telegraf/issues?q=webhookReply). To increase reliability (and reduce the number of "bug" reports), I suggest disabling it by default. Instead, I'd encourage people to enable it on per-call basis and document the gotchas:

1. Errors in the `webhookReply` calls cannot be reported.
2. `webhookReply` calls break ordering by resolving before bot API has a chance to process them (#1167, #1086).
3. `webhookReply` call confirms the update, which means it won't be processed again if bot crashes.

I'm keeping this a draft until I hear @KnorpelSenf opinion on some deprecations I wish to make here:

To discourage enabling `webhookReply` globally, I could deprecate https://github.com/telegraf/telegraf/blob/424f89552509f5490f096800a4843a8e255c5305/src/core/network/client.ts#L37

To avoid duplicating the docs, I could also deprecate https://github.com/telegraf/telegraf/blob/424f89552509f5490f096800a4843a8e255c5305/src/context.ts#L175-L181

# Checklist:

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -54,7 +54,7 @@ const DEFAULT_EXTENSIONS: Record<string, string | undefined> = {
 
 const DEFAULT_OPTIONS = {
   apiRoot: 'https://api.telegram.org',
-  webhookReply: true,
+  webhookReply: false,
   agent: new https.Agent({
     keepAlive: true,
     keepAliveMsecs: 10000,
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -34,6 +34,7 @@ namespace ApiClient {
   export interface Options {
     agent?: https.Agent | http.Agent
     apiRoot: string
+    /** @deprecated use `ctx.telegram.webhookReply` */
     webhookReply: boolean
   }
 
@@ -320,7 +321,7 @@ class ApiClient {
   /**
    * If set to `true`, first _eligible_ call will avoid performing a POST request.
    * Note that such a call:
-   * 1. cannot report errors,
+   * 1. cannot report errors or return meaningful value,
    * 2. resolves before bot API has a chance to process it,
    * 3. prematurely confirms the update as processed.
    *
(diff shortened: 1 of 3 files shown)


KnorpelSenf:
I agree, this is a good idea. It breaks more than it helps, the performance benefit is irrelevant for small bots.

We should, however, also provide a proper way to enable it, even though it should maybe not in the most obvious place of the config. Any suggestions for how you would do this per request?

wojpawlik:
I could leave `ctx.telegram.webhookReply` intact. Or I could deprecate it and introduce a `webhookReply` option to `callApi`, next to `signal`.

What do you think about the 2 deprecations I suggested?

KnorpelSenf:
> What do you think about the 2 deprecations I suggested?

Ah, I was referring to those when I said

> I agree, this is a good idea.

I also like the rest, that should be implied.

Are you planning on doing another breaking release soon? If so, we can indeed postpone the decision about `ctx.telegram.webhookReply`. Otherwise, I think it makes sense to come to a conclusion about whether it actually should be removed or not, and then introduce the breaking changes with 4.0.

wojpawlik:
I don't wish to delay 4.0 further.

wojpawlik:
Even tho I don't plan 5.0 anytime soon; my plans currently end at 4.1.

KnorpelSenf:
Alright, I see that. I'm fine with just deprecating them in 4.0

src/core/network/client.ts
```diff
@@ -320,7 +321,7 @@ class ApiClient {
   /**
    * If set to `true`, first _eligible_ call will avoid performing a POST request.
    * Note that such a call:
-   * 1. cannot report errors,
+   * 1. cannot report errors or return meaningful value,
```
KnorpelSenf, line 324: Sounds like we're talking about human values, when we actually mean a piece of memory

```suggestion
   * 1. cannot report errors or return a meaningful value,
```

</nitpick>

KnorpelSenf (approved):
Nice work

PR #1253, opened 2020-12-23
Don't bind `Context` methods

wojpawlik:
# Description

Not useful with `await`.

# Checklist:

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -61,15 +61,8 @@ export class Context {
   constructor(
     readonly update: tt.Update,
     readonly tg: Telegram,
-    public readonly botInfo: tt.UserFromGetMe
-  ) {
-    Object.getOwnPropertyNames(Context.prototype)
-      .filter(
-        (key) =>
-          key !== 'constructor' && typeof (this as any)[key] === 'function'
-      )
-      .forEach((key) => ((this as any)[key] = (this as any)[key].bind(this)))
-  }
+    readonly botInfo: tt.UserFromGetMe
+  ) {}
 
   get updateType() {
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
This makes it impossible to use `Context.reply` outside of any middleware function body, i.e.

```js
const { Context } = require('telegraf')

// ...

module.exports = function() {
  Context.reply('yay') // this operated on the last received context object
}

// ...
```

As a result, if some part of the user bot code wants to `reply`, the context object from the middleware has to be passed along the dependency chain deep down into the module that needs it.

```js
// ...

module.exports = function(ctx) {
  ctx.reply('yay') // we have to pass in the context object now
}

// ...
```

It is not the cleanest thing to use this feature (especially given that messages could be processed in parallel), but we have to be aware that this will cause a lot of trouble in the code bases of legacy bots!

Is there a benefit to removing it, other than improved performance?

wojpawlik:
Performance and nicer `console.log(ctx)` are the only benefits.

It's not adding any methods to `Context`: `TypeError: Context.reply is not a function`. It was binding methods on the instance, so that you could do `.then(ctx.reply)`. I don't think it's very useful nowadays.

KnorpelSenf:
> Performance and nicer `console.log(ctx)` are the only benefits.
> 
> It's not adding any methods to `Context`: `TypeError: Context.reply is not a function`. It was binding methods on the instance, so that you could do `.then(ctx.reply)`. I don't think it's very useful nowadays.

Ah, then I missed something. Thanks for the clarification.

KnorpelSenf (approved):

PR #1255, opened 2020-12-23
Only include exported members in docs

KnorpelSenf:
# Description

Updates `typedoc` to a beta version in order to remove clutter from the generated docs by reducing the displayed entities to those that `telegraf` actually `export`s.

Also, switching to the `minimal` theme gives the docs a much better overview.

## Type of change

Config update

# How Has This Been Tested?

`npm run build:docs`

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "scripts": {
     "prepare": "npm run build && mjs-entry src/index.ts > lib/index.mjs",
     "build": "tsc --build docs/examples",
-    "build:docs": "typedoc src/index.ts",
+    "build:docs": "typedoc src/index.ts --name telegraf.js",
     "pretest": "npm run build",
     "test": "ava test/*",
     "lint": "eslint --cache .",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "scripts": {
     "prepare": "npm run build && mjs-entry src/index.ts > lib/index.mjs",
     "build": "tsc --build docs/examples",
-    "build:docs": "typedoc src/index.ts --theme minimal",
+    "build:docs": "typedoc src/index.ts",
     "pretest": "npm run build",
     "test": "ava test/*",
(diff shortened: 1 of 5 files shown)


typedoc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "telegraf.js"
```
wojpawlik, line 2: Why is this needed? Typedoc seems to infer `"telegraf"`

package.json
```diff
@@ -31,7 +31,7 @@
   "scripts": {
     "prepare": "npm run build && mjs-entry src/index.ts > lib/index.mjs",
     "build": "tsc --build docs/examples",
-    "build:docs": "typedoc",
+    "build:docs": "typedoc src/index.ts --theme minimal",
```
wojpawlik, line 34: Let's discuss the theme later, once we can actually use the docs in prod.
```suggestion
    "build:docs": "typedoc src/index.ts",
```

wojpawlik (changes requested):

typedoc.json
KnorpelSenf: It's already pretty inconsistent with Telegraf vs. telegraf.js. Let's not introduce a third name

wojpawlik: Fair enough, just move this to `tsconfig`

wojpawlik: `tsconfig.json` not `package.json`

KnorpelSenf: Oops

wojpawlik (approved):
Why did you change `bash` to `shellscript` and `plaintext` to `text`?

KnorpelSenf:
> Why did you change `bash` to `shellscript` and `plaintext` to `text`?

`typedoc` does not support `bash` or `plaintext`. Run `typedoc --help` to find out what kind of markdown syntax highlighting is supported.

The change removes five warnings.

PR #1257, opened 2020-12-24
Add `ctx.telegram2`

wojpawlik:
# Description

Does everything* `ctx.telegram` does, but in 980 lines less, and automatically updates with `typegram`.

\* Notable omissions: `getFileLink`, writable `callApi`.

## Type of change

# Checklist:

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,4 +1,5 @@
 import * as tt from './telegram-types'
+import { Telegram2, telegram2 } from './telegram2'
 import ApiClient from './core/network/client'
 import Telegram from './telegram'
 
@@ -58,12 +59,16 @@ export const MessageSubTypes = [
 
 export class Context {
   readonly state: Record<string | symbol, any> = {}
+  readonly telegram2: Telegram2
 
   constructor(
     readonly update: tt.Update,
+    /** @deprecated use `ctx.telegram2` */
     readonly tg: Telegram,
     readonly botInfo: tt.UserFromGetMe
-  ) {}
+  ) {
+    this.telegram2 = telegram2(tg)
+  }
 
   get updateType() {
     return UpdateTypes.find((key) => key in this.update)
@@ -73,6 +78,7 @@ export class Context {
     return this.botInfo?.username
   }
 
+  /** @deprecated use `ctx.telegram2` */
   get telegram() {
     return this.tg
   }
@@ -542,10 +548,12 @@ export class Context {
     return this.telegram.addStickerToSet(this.from.id, ...args)
   }
 
+  /** @deprecated use `this.telegram2.getMyCommands` */
   getMyCommands() {
     return this.telegram.getMyCommands()
   }
 
+  /** @deprecated use `this.telegram2.setMyCommands` */
   setMyCommands(commands: readonly tt.BotCommand[]) {
     return this.telegram.setMyCommands(commands)
   }
--- a/src/context.ts
+++ b/src/context.ts
@@ -58,8 +58,10 @@ export const MessageSubTypes = [
 ] as const
 
 export class Context {
-  readonly state: Record<string | symbol, any> = {}
+  /** Allows to make bot API requests not covered by the shorthands */
   readonly telegram2: Telegram2
+  /** Weakly-typed namespace for sharing data between middlewares */
+  readonly state: Record<string | symbol, any> = {}
 
   constructor(
     readonly update: tt.Update,
--- a/src/context.ts
+++ b/src/context.ts
@@ -65,7 +65,6 @@ export class Context {
 
   constructor(
     readonly update: tt.Update,
-    /** @deprecated use `ctx.telegram2` */
     readonly tg: Telegram,
     readonly botInfo: tt.UserFromGetMe
   ) {
@@ -80,7 +79,6 @@ export class Context {
     return this.botInfo?.username
   }
 
-  /** @deprecated use `ctx.telegram2` */
   get telegram() {
     return this.tg
   }
--- a/src/telegram-types.ts
+++ b/src/telegram-types.ts
@@ -36,6 +36,7 @@ export type InputFile =
 type TelegrafTypegram = Typegram<InputFile>
 
 export type Telegram = TelegrafTypegram['Telegram']
+export type TelegramP = TelegrafTypegram['TelegramP']
 export type Opts<M extends keyof Telegram> = TelegrafTypegram['Opts'][M]
 export type InputMedia = TelegrafTypegram['InputMedia']
 export type InputMediaPhoto = TelegrafTypegram['InputMediaPhoto']
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -2,7 +2,6 @@ import * as tt from './telegram-types'
 import ApiClient from './core/network/client'
 import { isAbsolute } from 'path'
 
-/** @deprecated use `telegram2` */
 class Telegram extends ApiClient {
   /**
    * Get basic information about the bot
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -2,6 +2,7 @@ import * as tt from './telegram-types'
 import ApiClient from './core/network/client'
 import { isAbsolute } from 'path'
 
+/** @deprecated use `telegram2` */
 class Telegram extends ApiClient {
   /**
    * Get basic information about the bot
--- a/src/telegram2.ts
+++ b/src/telegram2.ts
@@ -1,0 +1,24 @@
+import ApiClient from './core/network/client'
+import { TelegramP } from './telegram-types'
+
+type BasicApiClient = Readonly<Pick<ApiClient, 'callApi'>>
+export type Telegram2 = Readonly<BasicApiClient & TelegramP>
+
+const proxyHandler: ProxyHandler<BasicApiClient> = {
+  get(target, p: keyof Telegram2) {
+    if (p === 'callApi') {
+      return target[p]
+    }
+    return target.callApi.bind(target, p)
+  },
+  set() {
+    return false
+  },
+  ownKeys() {
+    return []
+  },
+}
+
+export function telegram2(client: BasicApiClient) {
+  return new Proxy(client, proxyHandler) as Telegram2
+}
--- a/src/telegram2.ts
+++ b/src/telegram2.ts
@@ -14,6 +14,12 @@ const proxyHandler: ProxyHandler<BasicApiClient> = {
   set() {
     return false
   },
+  defineProperty() {
+    return false
+  },
+  deleteProperty() {
+    return false
+  },
   ownKeys() {
     return []
   },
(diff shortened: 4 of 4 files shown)


KnorpelSenf:
If I'm not mistaken, the point of  `ctx.telegram` has always been to provide a natural way to interact with the API. The most important values are in the function signature, and the rest of them are captured via the `Extra*` values.

You're dropping this convenience here. `Telegram` and `Telegram2` are incompatible. As you're deprecating `Telegram`, it would be consequent not to use it internally anymore. A lot of `context.ts` would have to rewritten.

How does using `Telegram2` impact the debugging experience for JS users? If Telegraf no longer knows which methods exist, I could imagine that this makes using the lib more difficult.

wojpawlik:
TypeScript should still autocomplete for JS users. If they mistype anyway, API will simply return 404, just checked.

> The most important values are in the function signature, and the rest of them are captured via the `Extra*` values

Convenient, but messy, see `Telegram::editMessage*`.

> A lot of `context.ts` would have to rewritten.

I'll do it for 5.0, when/if we decide to actually remove `Telegram`. I'm not removing anything here; we can undeprecate `Telegram` if we decide that the small bit of convenience is worth the inconsistencies and 900+ LoC.

KnorpelSenf:
What is messy about it?

wojpawlik:
Having to explicitly pass `undefined` to `editMessage*` methods. They would also be difficult to type if `typegram` splits the types into `message_id` and `inline_message_id` overloads.

KnorpelSenf:
What's wrong with making it two methods?

wojpawlik:
Splitting them into 2 distinct methods with different names would deviate from bot API.
Splitting into 2 overloads means that the `Telegram` methods won't typecheck, and need to be overloaded as well.

Back to the topic, it's just 30 lines of surprisingly simple code that can be deprecated if it turns out inferior. What concerns prevent you from approving this?

KnorpelSenf:
I don't really see the problem it solves, so I cannot evaluate how well it performs in doing so

wojpawlik:
It's not a bugfix, it's enhancement.

1. It's smaller (meaning faster startup hopefully).
2. It's perfectly consistent.
3. It's simple.
4. It requires little maintenance (it can auto update, meaning users get new API versions faster hopefully).

KnorpelSenf:
Do you believe it's possible to resemble the old behaviour with your approach? I agree that the code base is improved by this suggestion, I just think that as a consumer, I wouldn't understand why those Telegraf people (us) made me ditch
```ts
ctx.telegram.sendMessage(42, 'Hi')
```
and then forcing me to write
```ts
ctx.telegram.sendMessage({
  chat_id: 42,
  text: 'Hi'
})
```
instead. The same goes for all other methods. Looks a bit more like NTBA now, but that's only a benefit for those who are migrating.

You may see that as “a small bit of convenience”, but from the consumers point of view, nothing has improved at all. Everything just got a little bit more annoying. Most of them won't even notice the difference in startup time, so from their perspective, it has virtually no benefit.

If it's possible to do those tiny transformations without much effort, we could get the best of both worlds. We could then even look into doing something similar for the context, it could be feasible to do `s/send/replyWith/` or something.

wojpawlik:
This would require runtime mapping: `[methodName, argPos] -> argName`. Very Telegraf-specific, I wouldn't want Typegram to ship it. So it wouldn't auto-update. I don't wanna take `telegram2` this route.

Perhaps `Telegram` could be changed to that, but it'd be slower, and I doubt it'd reduce size significantly.

KnorpelSenf:
> So it wouldn't auto-update.

Not 100 %, that's true. Existing methods would update if new parameters are added. Newly added methods would still appear automatically if we wanted to. However, those would not automatically get the right convenience args, so it could be a good idea to _not_ let them appear right away, as a later change would be breaking. As a result, it would not update for new methods.

About code size reduction: that would still be pretty good. We'd need a mapping table that specifies a string or two for each method, something like
```
const mainArguments = {
  sendMessage: ['chat_id', 'text'],
  // ...
} as const
```
plus a type transformation. Still pretty good.

wojpawlik:
Fully agree. `Telegram` changes are obviously out of scope of this PR. I still wanna go for `telegram2` because of full auto-updates and being absolutely unopinionated: I wanted to update signature of `Telegram::setGameScore`, but I have no idea what it should be.

KnorpelSenf:
I see your point. Then let's do the following:
- merge `telegram2` as-is
- leave `telegram` without deprecation
- in a different PR, reduce the code size of `telegram` as much as possible
- investigate if we can do something similar for the context object

I'm wondering if we can find a better name for `telegram2` that makes it clear what the difference to `telegram` is. Something like `telegramRaw`? Maybe you have a better suggestion here.

Please add a quick sentence about the difference to the JSdoc to help users decide what they want to use.

`Context` replicates parts of the API. Hence, we could think about adding something similar to your suggestion here to context, perhaps under `ctx.raw` (or whatever name we choose).

After these changes, both versions of calling the API are extremely simple to maintain 👍🏻

wojpawlik:
I like the name `ctx.telegram2`. It suggests that it's newer and therefore preferred.

I'd only document and undeprecate `Telegram` in the "compressing" PR. It should stay deprecated if we fail to compress it.

I'm intrigued by the idea of trying to compress `Context`.

KnorpelSenf:
Am I correct that none of the changes we talked about here are breaking? Everything here happens purely under the hood, right?

KnorpelSenf:
I'd like to try to revamp `telegram` before we deprecate it, just to undeprecate it right after. As this is non-breaking, it does not have to be merged before 4.0.

wojpawlik:
While this is indeed non-breaking, it's not a good reason not to merge; it risks unnecessary merge conflicts and prevents me from documenting this more.

KnorpelSenf:
Would you be fine with undeprecating `Telegram` then until we know whether we can reduce the size of it or not? I assume that it will not take more than a few days until we know, so most likely even before the next release

wojpawlik:
No. I'm worried we might fail to compress and forget to deprecate until too late. I can be more transparent tho:

```ts
/*
 * @deprecated until we can compress `Telegram` and `Context`. Prefer `ctx.telegram2`.
 */
```

Compressing `Telegram` alone might not be enough of a reason to keep it, but I'm more than ok with keeping it if we can reuse the data to compress `Context` as well.

wojpawlik:
I changed my mind. We're very close to 4.0 release. Let's merge this right after that. Then consumers won't even notice the deprecation if we compress `Telegram` and `Composer` before 4.1.

wojpawlik:
Thinking again, neither `Telegram` nor `Telegram2` are exported now, and they'd be constructed mainly by library creators, not consumers, so how about moving both of them to `src/core/network`?

KnorpelSenf:
Why did you you want to remove `Telegram` again? Your new thing does not hurt anyone, but I don't see why you want to break all the bots out there. There's no point

KnorpelSenf:
We btw cannot apply the same transformation and do it non-breaking. Here's why:

We can indeed define a mapping table for the most important args per method, e.g.

```ts
const MainArgs = {
  sendMessage: ['chat_id', 'text'],
  // etc
} as const
```

And then we could easily derive a new version of the API from it:

```ts
export type Api = {
  [M in keyof Telegram]: M extends keyof typeof MainArgs
    ? (...args: [...Main<M>, Rest<M>]) => ReturnType<TelegramP[M]>
    : TelegramP[M]
}
type Main<M extends keyof typeof MainArgs> = Lookup<Opts<M>, typeof MainArgs[M]>
type Rest<M extends keyof typeof MainArgs> = Omit<Opts<M>, Main<M>[number]>
type Lookup<O, L> = L extends readonly [infer H, ...infer T]
  ? [H extends keyof O ? O[H] : H, ...Lookup<O, T>]
  : []
```

But when we call a method, it would only have `args_0`, `args_1` etc as argument names. Labelled tuple types don't fix the issue unless we can dynamically generate the labels, which does not seem possible yet.

The same would be true for the context object, so I won't even begin looking into touching that.

wojpawlik:
Yeah, losing argument names is unacceptable. https://github.com/microsoft/TypeScript/issues/41594#issuecomment-733277488

> Why did you you want to remove `Telegram` again?

Mostly because it's big. But it's also high-maintenance, and its method signatures are more subjective than I'd like.

wojpawlik:
`src/telegram.ts` is not just big, it's _the biggest_, so it _needs_ to be compressed or removed if we're after small install size:

```
~/JS/telegraf (develop $%=)
$ wc -l src/*.ts | sort -r
 3215 total
 1010 src/telegram.ts
  842 src/composer.ts
  608 src/context.ts
  227 src/telegraf.ts
  143 src/telegram-types.ts
  125 src/markup.ts
   95 src/button.ts
   79 src/session.ts
   56 src/router.ts
   19 src/middleware.ts
   11 src/index.ts
```

Any gains here are fourfold: `src/telegram.ts`, `lib/telegram.js`, `typings/telegram.d.ts` and `typings/telegram.d.ts.map`.

Other ideas for decreasing install size: remove `minimist` and `module-alias` dependencies in 5.0, consider [`removeComments`](https://www.staging-typescript.org/tsconfig#removeComments).

I didn't realize `scenes/` are this tiny, I don't mind them in the library and always loaded anymore:

```
~/JS/telegraf (develop $%=)
$ wc -l src/scenes/*.ts | sort -r
 278 total
 140 src/scenes/context.ts
  71 src/scenes/stage.ts
  52 src/scenes/base.ts
  15 src/scenes/index.ts
```

For completeness:

```
~/JS/telegraf (develop $%=)
$ wc -l src/core/**/*.ts | sort -r
  645 total
  353 src/core/network/client.ts
   90 src/core/network/polling.ts
   71 src/core/helpers/check.ts
   45 src/core/network/multipart-stream.ts
   43 src/core/network/webhook.ts
   26 src/core/network/error.ts
   12 src/core/helpers/compact.ts
    5 src/core/helpers/string.ts
```

wojpawlik:
Naively removing `src/telegram.ts` decreased publish size by 27% (from 400.7 kB to 291.6 kB). `v3.38.0` (which doesn't have declaration maps) was 178 kB.

wojpawlik:
`"removeComments": true` shaves off 53.4 kB, but it also strips JSDoc from `.d.ts` files. Bad.

wojpawlik:
`Telegram2` uses JSDoc straight from `typegram`. Current `Telegram` needs its own; I doubt proxy `Telegram` can have JSDoc at all.

Still, proxy `Telegram`, even missing arg names and JSDoc, might be better than removing `Telegram`.

wojpawlik:
I changed my mind.

I decided against moving `telegram.ts` and `telegram2.ts` in this PR.

And I can retract the deprecations. I suspect we will soon hit limitations of `telegram.ts`, we may deprecate then.

wojpawlik:
To avoid breaking, flavoring should be added to `typegram` before this is merged.

wojpawlik:
#1320 wouldn't have happened if `Telegram` used `typegram` types directly.

wojpawlik:
On 2nd thought, this is not the right place for this.

PR #1258, opened 2020-12-25
Misc

wojpawlik:
# Description

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "test": "ava test/*",
     "lint": "eslint --cache .",
     "checks": "npm test && npm run lint",
-    "refresh": "npm run clean && npm ci",
     "clean": "git clean -fX .eslintcache docs/build/ lib/ typings/"
   },
   "husky": {
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "test": "ava test/*",
     "lint": "eslint --cache .",
     "checks": "npm test && npm run lint",
+    "refresh": "npm run clean && npm ci",
     "clean": "git clean -fX .eslintcache docs/build/ lib/ typings/"
   },
   "husky": {
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -55,7 +55,7 @@ type GuardedContext<
  * that narrows down `tt.Update` when intersected with it.
  */
 type MountMap = {
-  [T in tt.UpdateType]: Record<T, object>
+  [T in tt.UpdateType]: Record<T, unknown>
 } &
   {
     [T in tt.MessageSubType]: {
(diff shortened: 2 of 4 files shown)


KnorpelSenf:
Looks good, but clashes with the work around handler timeouts etc, so let's merge this after we have proper error and timeout handling

package.json
```diff
@@ -36,7 +36,6 @@
     "test": "ava test/*",
     "lint": "eslint --cache .",
     "checks": "npm test && npm run lint",
-    "refresh": "npm run clean && npm ci",
```
KnorpelSenf, line 39: Rationale?

src/composer.ts
```diff
@@ -55,7 +55,7 @@ type GuardedContext<
  * that narrows down `tt.Update` when intersected with it.
  */
 type MountMap = {
-  [T in tt.UpdateType]: Record<T, object>
+  [T in tt.UpdateType]: Record<T, unknown>
```
KnorpelSenf, line 58: Less accurate and less readable. Any reason to do this?

src/telegraf.ts
```diff
@@ -199,12 +205,10 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
 
   private botInfoCall?: Promise<tt.UserFromGetMe>
   async handleUpdate(update: tt.Update, webhookResponse?: http.ServerResponse) {
-    debug('Processing update', update.update_id)
```
KnorpelSenf, line 202: This has always been the most useful debug line for me. What's the rationale behind dropping it?

package.json
wojpawlik: `npm ci` from within script fails on my machine, not willing to troubleshoot.

src/telegraf.ts
wojpawlik: No reason, I'll leave it in then. I think it should be moved until after `getMe` call tho, do you agree?

package.json
KnorpelSenf: And `npm i` does not?

src/composer.ts
wojpawlik: Aesthetics. `T & unknown` is diplayed as `T`, leaving more space for all the union variants.

package.json
KnorpelSenf: It works both on my machine and in CI. We can just as well leave it in. You don't need to use it.

src/telegraf.ts
KnorpelSenf: Yes

package.json
wojpawlik: I mean, `npm ci` works, but `npm ci` invoked by `npm run refresh` does not. Does that work for you?

KnorpelSenf: Sure, just tried it. I also don't know why there would be a difference. What Linux version does your installation of WSL rely on?

wojpawlik: I'm on Windows Windows, no WSL. And Node 14.0.0, I should probably upgrade. Re-adding then.

KnorpelSenf (approved):

PR #1262, opened 2020-12-27
`p-timeout` update handling

wojpawlik:
# Description

Truthfully, anything more complex is "a premature performance optimization that is not backed by runtime analyses or performance tests". See https://stackoverflow.com/a/13616530. Closes #1247. Closes #1252.

Would you be fine if I changed `MiddlewareFn` to return `Promise<unknown>`? It's the simplest solution to the type error.

# How Has This Been Tested?

Purposefully timing out. Also, I ensured that graceful stop continues to work.

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2926,6 +2926,17 @@
       "dev": true,
       "requires": {
         "p-timeout": "^3.1.0"
+      },
+      "dependencies": {
+        "p-timeout": {
+          "version": "3.2.0",
+          "resolved": "https://registry.npmjs.org/p-timeout/-/p-timeout-3.2.0.tgz",
+          "integrity": "sha512-rhIwUycgwwKcP9yTOOFK/AKsAopjjCakVqLHePO3CC6Mir1Z99xT+R63jZxAT5lFZLa2inS5h+ZS2GvR99/FBg==",
+          "dev": true,
+          "requires": {
+            "p-finally": "^1.0.0"
+          }
+        }
       }
     },
     "p-finally": {
@@ -2962,13 +2973,9 @@
       }
     },
     "p-timeout": {
-      "version": "3.2.0",
-      "resolved": "https://registry.npmjs.org/p-timeout/-/p-timeout-3.2.0.tgz",
-      "integrity": "sha512-rhIwUycgwwKcP9yTOOFK/AKsAopjjCakVqLHePO3CC6Mir1Z99xT+R63jZxAT5lFZLa2inS5h+ZS2GvR99/FBg==",
-      "dev": true,
-      "requires": {
-        "p-finally": "^1.0.0"
-      }
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/p-timeout/-/p-timeout-4.1.0.tgz",
+      "integrity": "sha512-+/wmHtzJuWii1sXn3HCuH/FTwGhrp4tmJTxSKJbfS+vkipci6osxXM5mY0jUiRzWKMTgUT8l7HFbeSwZAynqHw=="
     },
     "p-try": {
       "version": "2.2.0",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "minimist": "^1.2.5",
     "module-alias": "^2.2.2",
     "node-fetch": "^2.6.1",
+    "p-timeout": "^4.1.0",
     "sandwich-stream": "^2.0.2",
     "typegram": "^3.0.0"
   },
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -387,7 +387,7 @@ class ApiClient {
     const apiUrl = `${options.apiRoot}/bot${token}/${method}`
     config.agent = options.agent
     config.signal = signal
-    config.timeout = 120000 // 2 minutes
+    config.timeout = 60_000 // 60s in ms
     const res = await fetch(apiUrl, config)
     const data = await res.json()
     if (!data.ok) {
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -10,6 +10,7 @@ import Context from './context'
 import d from 'debug'
 import generateCallback from './core/network/webhook'
 import { Polling } from './core/network/polling'
+import pTimeout from 'p-timeout'
 import Telegram from './telegram'
 import { TlsOptions } from 'tls'
 import { URL } from 'url'
@@ -17,7 +18,7 @@ const debug = d('telegraf:main')
 
 const DEFAULT_OPTIONS: Telegraf.Options<Context> = {
   telegram: {},
-  handlerTimeout: 5000,
+  handlerTimeout: 90_000, // 90s in ms
   contextType: Context,
 }
 
@@ -25,7 +26,6 @@ function always<T>(x: T) {
   return () => x
 }
 const anoop = always(Promise.resolve())
-const wait = util.promisify(setTimeout)
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 namespace Telegraf {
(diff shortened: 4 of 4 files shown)


KnorpelSenf:
Basically the only change here is that middleware that exceeds the timeout does not resolve but reject. All other things are left unimproved.

It does not solve the problem that some errors are swallowed.

It still blocks the bot while middleware is running.

It does not solve the `unref` problem, i.e. the process can't complete naturally upon `bot.stop` until timeout (minor issue, most people force-stop).

I see not providing a separate handler for timeouts as a disadvantage because I cannot imagine a realistic scenario where it would make sense to handle timeouts like a failing operation (or vice-versa), but this decision only impacts usability, it's not a bug. After all, users can apply a workaround by adding their own `instanceof` checks to later split up the control flow again.

The extra dependency was in fact on the list of packages I considered (and discarded) before starting the work on #1247, among roughly a dozen other existing solutions. By the way, it may be possible to combine several new dependencies to somehow resemble #1247 but I figured that this would be even more complicated than writing a solution that is truly tailored to our use case.

Given that you refuse to look into my PR, though, we have to change the way we mitigate the issues with update handling. Here are two suggestions to begin with:

It is a feasible solution to merge something primitive at first and then deal with the resulting issues in subsequent PRs (maybe copied from #1247) one by one again. I'm too frustrated by your baseless blocking that I won't spend any more time here, so you'd have to fix that yourself. At least in the short term, this PR is improving on the problem, so it could be a first step.

Alternatively, we can just open a “known issues” section in the new docs where we list the remaining issues. Not doing the work at all is always by far the easiest option. The main advantage is that we can simply ship half-baked code and we don't need to worry about it. It's the option I prefer because then I don't have to do anything.

KnorpelSenf:
About the compiling issue, you can use `instanceof Promise` (would be one extra line but one edge case and a lot of broken example bots less)

wojpawlik:
I'm fine with fixing the examples, was just waiting for approval before doing the work. While I could do `instanceof Promise` or pass the return value thru [`aVoid`](https://github.com/telegraf/telegraf/blob/e4049eb405bff40691aa9e797f67798354520965/src/telegraf.ts#L29-L31), it feels like working around a problem that shouldn't exist.

> It does not solve the problem that some errors are swallowed.

There is no error to swallow if we crash before the error is thrown \**taps head*\*

Seriously tho, I could open a PR to `p-timeout` to add `promise` field to `TimeoutError` so that error after timeout can still be handled.

> It still blocks the bot while middleware is running.

When polling. I don't believe this is solvable without confirming updates prematurely.

> It does not solve the `unref` problem, i.e. the process can't complete naturally upon `bot.stop` until timeout (minor issue, most people force-stop).

Wrong. `p-timeout` cancels the timer when promise settles.

> Given that you refuse to look into my PR

I'm sorry. Since I discovered `p-timeout`, I stopped thinking about #1247 and #1252, and I was waiting on your opinion on this PR.

One of my main concerns about #1247 is that it has to implement its own error handling, presumably because it breaks control flow, making the library difficult to reason about: https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/

wojpawlik:
A few of the examples would actually be improved by fixing them to return `Promise<unknown>`.

KnorpelSenf:
> it feels like working around a problem that shouldn't exist

Well there are synchronous and asynchronous methods in JavaScript … I don't really know how we could “solve” that, apart from interdicting the usage of one of two kinds for middleware. Having synchronous middleware has never been an issue for `telegraf` (and still isn't), so why make people avoid it?
```ts
      await pTimeout(
        Promise.resolve(this.middleware()(ctx, anoop)),
        this.options.handlerTimeout
      )
```

> There is no error to swallow if we crash before the error is thrown *_taps head_*

To which I reply, **if and only if**. We have custom error handlers _\*taps head\*_

> I don't believe this is solvable without confirming updates prematurely.

We should never do that by default, I agree. However, sometimes you do want to be very fast and best-effort rather than being slow and very reliable. We still have error handlers that let you retry processing the update if you want. In that case, only abnormal crashes of the Node process could lead to missed updates (if there is no chance to store them somewhere), and those are extremely rare, so it is a trade-off that one may take.

Note that not even allowing opt-in `concurrency` means that it may take up to `handlerTimeout` for your bot to respond to any message in any chat, no matter what you try. I find five seconds response time already too slow.

> Wrong. `p-timeout` cancels the timer when promise settles.

Exactly, but only if it actually does. It won't complete when some middleware refuses to resolve for whatever reason, so we agreed at some point that the timer should be `unref`ed because it should not be the _middleware timeout_ that determines when the process may exit. It should be the user code.

> has to implement its own error handling

How would you not implement your own error handling in `telegraf`? It's done that ever since, this PR just adjusts it.

> presumably because it breaks control flow

So how would you implement concurrency otherwise, if not with promises? Are you suggesting to remove promises from Node, and move to a different way to handle concurrency, as the article you quoted advocates?

wojpawlik:
`Promise.resolve` is not bad, how could I forget about it 🤦‍♂️
I still don't like `| void`, but I won't delay 4.0; I'll keep discussing and bring it up for 5.0 if I somehow don't forget.

> It won't complete when some middleware refuses to resolve for whatever reason

Then the process cannot exit gracefully anyway. Not only because of the timer, but because of the hanging middleware. Aborting it couldn't hurt.

> How would you not implement your own error handling in `telegraf`?

Valid point, `Telegraf::catch` is own error handling. But Telegraf internals still use standard `try catch finally`, because they can. #1247 breaks that.

> Are you suggesting to remove promises from Node, and move to a different way to handle concurrency, as the article you quoted advocates?

No. Promises are good enough. But to not break control flow, they must reject on error, which #1247 prevents.

KnorpelSenf:
> Valid point, `Telegraf::catch` is own error handling. But Telegraf internals still use standard `try catch finally`, because they can. #1247 breaks that.

I can easily change it to use try-catch syntax if you prefer it, i.e. turn the `p.catch().finally()` into a try-catch-finally block

> they must reject on error

Not sure what you mean—they still reject on error, no?

KnorpelSenf:
> Then the process cannot exit gracefully anyway.

Yes that's possible (`ctx => new Promise(() => {})` neither resolves nor keeps the process alive), but this would probably only be accidental. In that case, it may even make sense not to let the process die in the first place, so I agree that we should maybe reconsider this requirement. If we simply take the freedom to keep the process alive (and document this clearly ofc), this would in fact save a lot of hassle. I could try to redo #1247, keeping all the benefits and dropping `unref`.

I believe that we could then drop implementing our own queue, but I'll have to investigate this before I can say that for sure. Would that attempt make sense to you?

wojpawlik:
> If we simply take the freedom to keep the process alive (and document this clearly ofc), this would in fact save a lot of hassle.

You mean, to disallow `handleError` to stop the process from crashing? I was about to suggest that. It deosn't even need extra documentation. All that's needed is changing its return type to `MaybePromise<never>`.

KnorpelSenf:
> > If we simply take the freedom to keep the process alive (and document this clearly ofc), this would in fact save a lot of hassle.
> 
> You mean, to disallow `handleError` to stop the process from crashing? I was about to suggest that. It deosn't even need extra documentation. All that's needed is changing its return type to `MaybePromise<never>`.

No, I mean to not call `unref` on the timers.

wojpawlik:
I know it's not a guarentee, but when consumer ignores the types, they enter undefined behavior territory.

KnorpelSenf:
… or they don't use TypeScript yet.

Back to topic: do you agree that we only document that we keep the process alive, instead of trying to not do it?

wojpawlik:
Thinking again, restricting `handleError` to return `MaybePromise<never>` is probably too harsh, it would disallow globally ignoring known harmless errors, such as `AbortError`.

Documenting sounds good. It would also allow us to change internals later, as long as invariants are maintained. I'm thinking:

> ### Error handling
>
> If middleware throws an error, Telegraf will call `bot.handleError`. If it rethrows, update source will be closed, and then process will [hopefully](https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode) terminate. You can overwrite `bot.handleError` if you need to.
> ⚠️ **Careful!** After an unexpected error, process might be left in invalid state and thus should be allowed to terminate!
> ℹ️ In production, `systemd` or [`pm2`](https://www.npmjs.com/package/pm2) can restart your bot if it exits for any reason.

I'd put it outside `## Getting started`, in `## Intermediate topics` or something.

Mind if I deprecate `Telegraf::catch` in favior of reassigning `bot.handleError`? @KnorpelSenf @dotcypress 
If it's fine with you, I will open a pull request after #1271.

wojpawlik:
Marking as draft until https://github.com/sindresorhus/p-timeout/pull/19 and until we finish discussing #1156.

PR #1263, opened 2020-12-28
Adjust and re-add tests for custom session storage

KnorpelSenf:
# Description

Title says it all

## Type of change

Tests for `telegraf.ts`

# How Has This Been Tested?

`npm t`

# Checklist:

### DIFF
--- a/test/telegraf.js
+++ b/test/telegraf.js
@@ -1,7 +1,7 @@
 const test = require('ava')
 const { Telegraf, session } = require('../')
 
-function createBot(...args) {
+function createBot (...args) {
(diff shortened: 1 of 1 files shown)


wojpawlik:
The tests don't check that `session` _actually_ uses the passed storage, I'll fix this.

wojpawlik (approved):
I think it'd be best to align naming with 3.38 to avoid unnecessary breakage, but I don't care enough to prevent merging.

KnorpelSenf:
I agree, but this is not the responsibility of this PR. It only touches tests. We should open a second one that renames the properties.

PR #1266, opened 2020-12-31
Enable `strict` mode for examples

KnorpelSenf:
# Description

Reworks scene types once again. This makes it possible to enable `strict` compilation for examples.

## Type of change

Types for scenes and wizards are reworked, and the respective examples are changed accordingly.

A few other examples were updated in a way that lets them be compiled with `strict: true`.

Config update to `docs/examples/tsconfig.json`.

Some weird compilation errors still need to be fixed, but I'm not sure if that's actually to be done in the examples or rather in `composer.ts`.

# How Has This Been Tested?

`npm run checks`

# Checklist:

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ (does not apply)


docs/examples/echo-bot-module.js
```diff
@@ -31,8 +31,8 @@ bot.help(async (ctx) => {
   const info = commands.reduce((acc, val) => `${acc}/${val.command} - ${val.description}\n`, '')
   return ctx.reply(info)
 })
-bot.action('delete', ({ deleteMessage }) => deleteMessage())
+bot.action('delete', (ctx) => ctx.deleteMessage())
 bot.on('dice', (ctx) => ctx.reply(`Value: ${ctx.message.dice.value}`))
-bot.on('message', (ctx) => ctx.telegram.sendCopy(ctx.chat.id, ctx.message, keyboard))
+bot.on('message', (ctx) => ctx.telegram.sendCopy(ctx.message.chat.id, ctx.message, keyboard))
```
KnorpelSenf, line 36: Should we switch over to `copyMessage` here and at all other places? It does require one extra argument, but `sendCopy` is deprecated.

wojpawlik: Probably, but I can't be bothered.

KnorpelSenf: Not exactly in scope of this PR anyway. We can do it while revamping docs, or even later. Not required before 4.0 either.

KnorpelSenf:
LGTM

src/scenes/base.ts
```diff
@@ -12,9 +11,7 @@ export interface SceneOptions<C extends Context> {
   leaveHandlers: ReadonlyArray<MiddlewareFn<C>>
 }
 
-export class BaseScene<
-  C extends SceneContext = SceneContext
-> extends Composer<C> {
+export class BaseScene<C extends Context = Context> extends Composer<C> {
```
wojpawlik, line 14: Shouldn't `C` extend a more specific subtype here?

wojpawlik (commented):
Why did you decide to convert `docs/examples/express-webhook-bot` to `.ts`?

src/scenes/base.ts
wojpawlik: I mean, some of the scenes types (I don't know which) require `ctx.session` to be present, types should express that, no?

KnorpelSenf: In theory yes, but this has proven not to be practical. `ctx.session` should contain `_session` which in turn contains a custom type, so we'd immediately need another type parameter here that has to match with `C`. In addition, there is mutual recursion (`ctx` depends on `ctx.scene` depends on `ctx`), hence making extending any of the two extremely challenging.

I can write a more verbose explanation why `Context` is the best constraint here, for now I'll just say that I tried to apply a few different more advanced types. I had to revert all of the changes because it became incomprehensible and messy.

KnorpelSenf:
> Why did you decide to convert `docs/examples/express-webhook-bot` to `.ts`?

Because we had implicit `any`s in https://github.com/telegraf/telegraf/pull/1266/files#diff-4e1ceff248494e2881da3986927f558155db1fd31f8c17e8812fb933ac4a6ba0R21 which are not allowed with `strict` mode, and they cannot be annotated otherwise.

docs/examples/express-webhook-bot.ts
```diff
@@ -11,7 +18,7 @@ bot.on('text', ({ replyWithHTML }) => replyWithHTML('<b>Hello</b>'))
 bot.telegram.setWebhook('https://----.localtunnel.me/secret-path')
 
 const app = express()
-app.get('/', (req, res) => res.send('Hello World!'))
+app.get('/', (req: any, res: any) => res.send('Hello World!'))
```
KnorpelSenf, line 21: Would be cool if you found a better solution here, I didn't bother.

wojpawlik (approved):

PR #1268, opened 2020-12-31
Rewrite `generateCallback`

smaznet:
# Description

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/core/network/webhook.ts
+++ b/src/core/network/webhook.ts
@@ -21,9 +21,9 @@ export default function (
       res.statusCode = 403
       return res.end()
     }
-     let body = ''
-    for await (let chunk of req) {
-      body += chunk.toString()
+    let body = ''
+    for await (const chunk of req) {
+      body += (chunk as Buffer).toString()
     }
     let update: Update
     try {
--- a/src/core/network/webhook.ts
+++ b/src/core/network/webhook.ts
@@ -8,7 +8,7 @@ export default function (
   updateHandler: (update: Update, res: http.ServerResponse) => Promise<void>,
   errorHandler: (err: SyntaxError) => void
 ) {
-  return (
+  return async (
     req: http.IncomingMessage,
     res: http.ServerResponse,
     next?: () => void
@@ -34,7 +34,7 @@ export default function (
       return errorHandler(error)
     }
     try {
-      await updateHandler(update, res);
+      await updateHandler(update, res)
       if (!res.writableEnded || !res.finished) {
         res.end()
       }
--- a/src/core/network/webhook.ts
+++ b/src/core/network/webhook.ts
@@ -5,8 +5,7 @@ const debug = d('telegraf:webhook')
 
 export default function (
   hookPath: string,
-  updateHandler: (update: Update, res: http.ServerResponse) => Promise<void>,
-  errorHandler: (err: SyntaxError) => void
+  updateHandler: (update: Update, res: http.ServerResponse) => Promise<void>
 ) {
   return async (
     req: http.IncomingMessage,
@@ -27,10 +26,11 @@ export default function (
     let update: Update
     try {
       update = JSON.parse(body)
-    } catch (error) {
+    } catch (error: unknown) {
       res.writeHead(415)
       res.end()
-      return errorHandler(error)
+      debug('Failed to parse request body:', error)
+      return
     }
     try {
       await updateHandler(update, res)
--- a/src/core/network/webhook.ts
+++ b/src/core/network/webhook.ts
@@ -12,7 +12,7 @@ export default function (
     req: http.IncomingMessage,
     res: http.ServerResponse,
     next?: () => void
-  ): void => {
+  ): Promise<void> => {
     debug('Incoming request', req.method, req.url)
     if (req.method !== 'POST' || req.url !== hookPath) {
       if (typeof next === 'function') {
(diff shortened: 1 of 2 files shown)


src/core/network/webhook.ts
```diff
@@ -8,43 +8,39 @@ export default function (
...
-      body += chunk.toString()
-    })
-    req.on('end', () => {
-      let update: Update
-      try {
-        update = JSON.parse(body)
-      } catch (error) {
-        res.writeHead(415)
+    for await (const chunk of req) {
+      body += (chunk as Buffer).toString()
```
wojpawlik, line 25: I hate this cast, is `chunk` being `any` a bug of `@types/node`?

wojpawlik (changes requested):

src/core/network/webhook.ts
smaznet: i do this cast because i got an error about + operator only is usable if both are string

wojpawlik:
```suggestion
    for await (const chunk of req) {
      body += String(chunk)
```

KnorpelSenf: I'd rather perform the cast consistently with
https://github.com/telegraf/telegraf/blob/056dca05c53053d656de538365ec057dc677d806/src/core/network/client.ts#L154

wojpawlik: I'd argue `String` is clearer.

smaznet: i used as Buffer because its a typescript code and doesn't change the js output

but using `${body}` or String(body) changes js code

src/core/network/webhook.ts
```diff
@@ -8,43 +8,39 @@ export default function (
...
+      res.writeHead(415)
+      res.end()
+      return errorHandler(error)
+    }
+    try {
+      await updateHandler(update, res)
+      if (!res.writableEnded) {
         res.end()
-        return errorHandler(error)
       }
```
wojpawlik, line 39: Probably unnecessary, but too afraid to remove, not tested.

src/core/network/webhook.ts
```diff
@@ -8,43 +8,39 @@ export default function (
...
-            res.end()
-          }
-        })
-        .catch((err: unknown) => {
-          debug('Webhook error', err)
-          res.writeHead(500)
-          res.end()
-        })
-    })
+    } catch (err) {
```
wojpawlik, line 40:
```suggestion
    } catch (err: unknown) {
```

wojpawlik (approved):

src/core/network/webhook.ts
```diff
@@ -8,43 +8,39 @@ export default function (
...
+      update = JSON.parse(body)
+    } catch (error) {
+      res.writeHead(415)
+      res.end()
+      return errorHandler(error)
+    }
+    try {
+      await updateHandler(update, res)
+      if (!res.writableEnded) {
         res.end()
```
wojpawlik, line 38: Probably unnecessary, but too afraid to remove, not tested.

wojpawlik: It's actually covered by `webhookReply` tests I redid recently, and they still pass without it; removing.

src/core/network/webhook.ts
```diff
@@ -8,43 +8,36 @@ export default function (
...
+    for await (const chunk of req) {
+      body += String(chunk)
+    }
+    let update: Update
+    try {
+      update = JSON.parse(body)
+    } catch (error) {
+      res.writeHead(415)
+      res.end()
+      return errorHandler(error)
```
wojpawlik, line 33: One more thing, `Telegraf` passes its `debug` instance here, which is weird. I'll use This module's own one.

KnorpelSenf:
Will review tomorrow.

src/core/network/webhook.ts
```diff
@@ -5,46 +5,39 @@ const debug = d('telegraf:webhook')
...
+      res.end()
+      debug('Failed to parse request body:', error)
+      return
+    }
+    try {
+      await updateHandler(update, res)
+    } catch (err: unknown) {
+      debug('Webhook error', err)
+      res.writeHead(500)
+      res.end()
```
wojpawlik, line 40: And another one, this error should probably be rethrown:

```suggestion
      res.writeHead(500)
      res.end()
      throw err
```

KnorpelSenf: This makes sense to me. @smaznet what do you think?

src/core/network/webhook.ts
```diff
@@ -5,46 +5,39 @@ const debug = d('telegraf:webhook')
...
-        res.writeHead(415)
-        res.end()
-        return errorHandler(error)
-      }
-      updateHandler(update, res)
-        .then(() => {
-          if (!res.finished) {
-            res.end()
-          }
-        })
```
KnorpelSenf, line 42: Previously, if `updateHandler` completed successfully, `res.end()` was called if necessary. Is there a reason why you chose not to do this anymore?

KnorpelSenf (commented):
Great work. I have one question before I approve this:

src/core/network/webhook.ts
smaznet: yourself removed in the remove duplicate res.end commit

https://github.com/telegraf/telegraf/pull/1268/commits/226cbfbba14ad43cf2646875c425174ce40e1661

smaznet: @KnorpelSenf 
i forgot to mention you in previous message
this line removed by @wojpawlik in this commit https://github.com/telegraf/telegraf/commit/226cbfbba14ad43cf2646875c425174ce40e1661

wojpawlik: https://github.com/telegraf/telegraf/pull/1268#discussion_r550492121

src/core/network/webhook.ts
KnorpelSenf: Tests can never show the absence of bugs, only their presence.

Can you explain why we don't have to close the connection? :thinking:

KnorpelSenf: If I see this correctly, then `end` is not called if the call to `getMe` fails. (This is not covered by tests.)

KnorpelSenf: Reintroduce the finally block you removed and this will be fixed.

KnorpelSenf: Ah no, the `catch` should do it, then I guess it's fine … this is confusing

KnorpelSenf (approved):

Issue #1205, opened 2020-12-01
Move to `typedoc`


wojpawlik:
- [x] Set up a GitHub action to publish docs on GitHub Pages every release.
- [ ] Try to build docs anyway at every commit to spot breakage early.
- [x] Remove (now redundant) handwritten API reference.
- [x] Try to make `docs/README.md` shorter. Move information into JSDoc. Remove junk. Deduplicate.
- [ ] Avoid putting example code directly into readme, since it's not typechecked there and can easily get outdated.
  Prefer putting it in `docs/examples/` and linking to it.
  Or perhaps embed it with `<iframe>`, Github escapes it, but `marked` used by `typedoc` [doesn't](https://www.npmjs.com/package/marked#usage) 😈
- [x] Make sure `docs/README.md` contains everything from `readme.md`.
- [x] Remove `readme.md`, now inferior to `docs/README.md`.
- [x] Move `docs/README.md` back to root and update https://github.com/telegraf/telegraf/blob/1f7dee72c702bdf05791ccba2a8a8c3b3c6f9dfc/tsconfig.json#L19

KnorpelSenf:
Playing around with a 0.20 beta version of `typedoc`, I found that we can specify a number of every points. Only type definitions of actually exposed modules will be included in the docs. Right now, the website is extremely cluttered. As a newcomer, this is overwhelming.

I agree that we should use typedoc, but it's only half of what we need. The other half are guides and tutorials on how to use it. An API reference is not helpful if you don't know the purpose of it.

KnorpelSenf:
@wojpawlik where would we write a quickstart guide if we wanted to begin with that?

wojpawlik:
In README I think, after cleaning it up.

KnorpelSenf:
In other words, `typedoc` does not allow writing separate pages in addition to the generated ones? I did not manage to find something like that

PR #1271, opened 2021-01-01
Remove outdated, handwritten API reference

### LINKED ISSUE #898: Context.command does not support regex matching
The documentation at https://telegraf.js.org/#/?id=context mentions that the ctx.match propery is available only for hears, command, action and inline queries, but it does not seem to be available in command queries.

### LINKED ISSUE #916: replyWithContact is missing in docs and not works properly.
## Context

* Telegraf.js Version: 3.36.0
* Node.js Version: 12.7.0

## Expected Behavior

"replyWithContact" is not listed in the documentation and doesn't work properly.

## Current Behavior

It returns sometimes "Error: 429: Too Many Requests: retry after 84754".

### LINKED ISSUE #1264: Some examples are no longer valid
## Context

As I started working more with the library, I realised that much of the documentations is no longer valid / outdated. For instance, `inline-bot` example, where

```
bot.on('chosen_inline_result'...
```
does not trigger anything, and by looking in the codebase, I only see typings, but I guess the implementation has changed on either side. 
Same goes for `keyboard-bot` where it is suggested to simply include `Markup.keyboard` in the `reply` extras, but what actually needs to be done is to pass them as `{reply_markup: <keyboard_obj>}`. 
I guess there is more that I haven't come across yet.

* Telegraf.js Version: master branch, release 3.38
* Node.js Version: 10 and 12
* Operating System: Ubuntu 18, Mac OS

## Expected Behavior

Examples are valid and can be copy-pasted and ran.

## Current Behavior

Examples only hint towards the expected code, but require user to dive deep into the library to find the actual usage

wojpawlik:
I'll move `README.md` to root directory right after this is merged. I didn't do it here to preserve `git blame`.

Structure is still a little messy, I'll work on that in another pull request, before or after 4.0.

- Fixes #898
- Fixes #916
- Fixes #1264
- Advances #1205


KnorpelSenf (approved):
This is good and should be merged.

I think we agree that there's still a lot more work to do about the docs after this.

PR #1272, opened 2021-01-04
Code cleanup

dimankiev:
# Description
I've noticed, that I can improve code of telegraf, that I use in my own projects.
I've reviewed some files and have done a code cleanup by using methods and improvements of new versions of JavaScript engine

## Type of change
Code cleanup

# How Has This Been Tested?

**Test Configuration**:
* Node.js Version: 15.5

After code cleanup I've done performance tests to be sure, that this changes are not affecting efficiency
CI Tests can be failed because some properties are not defined explicitly in Update interface

# Checklist:

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -77,60 +77,17 @@ export class Context {
     return this.tg
   }
 
-  get message() {
-    if (!('message' in this.update)) return undefined
-    return this.update.message
-  }
-
-  get editedMessage() {
-    if (!('edited_message' in this.update)) return undefined
-    return this.update.edited_message
-  }
-
-  get inlineQuery() {
-    if (!('inline_query' in this.update)) return undefined
-    return this.update.inline_query
-  }
-
-  get shippingQuery() {
-    if (!('shipping_query' in this.update)) return undefined
-    return this.update.shipping_query
-  }
-
-  get preCheckoutQuery() {
-    if (!('pre_checkout_query' in this.update)) return undefined
-    return this.update.pre_checkout_query
-  }
-
-  get chosenInlineResult() {
-    if (!('chosen_inline_result' in this.update)) return undefined
-    return this.update.chosen_inline_result
-  }
-
-  get channelPost() {
-    if (!('channel_post' in this.update)) return undefined
-    return this.update.channel_post
-  }
-
-  get editedChannelPost() {
-    if (!('edited_channel_post' in this.update)) return undefined
-    return this.update.edited_channel_post
-  }
-
-  get callbackQuery() {
-    if (!('callback_query' in this.update)) return undefined
-    return this.update.callback_query
-  }
-
-  get poll() {
-    if (!('poll' in this.update)) return undefined
-    return this.update.poll
-  }
-
-  get pollAnswer() {
-    if (!('poll_answer' in this.update)) return undefined
-    return this.update.poll_answer
-  }
+  get message() { return this.update?.message }
+  get editedMessage() { return this.update?.edited_message }
+  get inlineQuery() { return this.update?.inline_query }
+  get shippingQuery() { return this.update?.shipping_query }
+  get preCheckoutQuery() { return this.update?.pre_checkout_query }
+  get chosenInlineResult() { return this.update?.chosen_inline_result }
+  get channelPost() { return this.update?.channel_post }
+  get editedChannelPost() { return this.update?.edited_channel_post }
+  get callbackQuery() { return this.update?.callback_query }
+  get poll() { return this.update?.poll }
+  get pollAnswer() { return this.update?.poll_answer }
 
   get chat() {
     return (
@@ -160,11 +117,7 @@ export class Context {
     return (this.callbackQuery ?? this.chosenInlineResult)?.inline_message_id
   }
 
-  get passportData() {
-    if (this.message == null) return undefined
-    if (!('passport_data' in this.message)) return undefined
-    return this.message?.passport_data
-  }
+  get passportData() { return this?.message?.passport_data }
 
   /** @deprecated use `ctx.telegram.webhookReply` */
   get webhookReply(): boolean {
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -55,7 +55,6 @@ export class Polling {
           (err.code === 401 || err.code === 409)
         ) {
           this.skipOffsetSync = true
-          throw err
         }
         throw err
       }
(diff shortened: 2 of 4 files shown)


src/context.ts
```diff
@@ -77,60 +77,17 @@ export class Context {
...
+  get editedMessage() { return this.update?.edited_message }
+  get inlineQuery() { return this.update?.inline_query }
+  get shippingQuery() { return this.update?.shipping_query }
+  get preCheckoutQuery() { return this.update?.pre_checkout_query }
+  get chosenInlineResult() { return this.update?.chosen_inline_result }
+  get channelPost() { return this.update?.channel_post }
+  get editedChannelPost() { return this.update?.edited_channel_post }
+  get callbackQuery() { return this.update?.callback_query }
+  get poll() { return this.update?.poll }
+  get pollAnswer() { return this.update?.poll_answer }
```
dimankiev, line 90: Use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead of checking if property is in the object

src/context.ts
```diff
@@ -77,60 +77,17 @@ export class Context {
...
-
-  get poll() {
-    if (!('poll' in this.update)) return undefined
-    return this.update.poll
-  }
-
-  get pollAnswer() {
-    if (!('poll_answer' in this.update)) return undefined
-    return this.update.poll_answer
-  }
```
dimankiev, line 133: Redundant expressions to check if property exists in the object:
`if (!('something' in this.update)) return undefined`
We can use:
`return this.update?.something`
Which will return undefined if there is no `something`

src/context.ts
```diff
@@ -160,11 +117,7 @@ export class Context {
     return (this.callbackQuery ?? this.chosenInlineResult)?.inline_message_id
   }
 
-  get passportData() {
-    if (this.message == null) return undefined
-    if (!('passport_data' in this.message)) return undefined
-    return this.message?.passport_data
-  }
```
dimankiev, line 167: Redundant expressions to check if `message` exists in the object and there is `passport_data` in the message:
`if (this.message == null) return undefined`
`if (!('passport_data' in this.message)) return undefined`
We can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here:
`return this?.message?.passport_data`

> The ?. operator functions similarly to the . chaining operator, except that instead of causing an error if a reference is nullish (null or undefined), the expression short-circuits with a return value of undefined

So this will return undefined if there is no `passport_data` defined or no `message` at all

src/core/network/polling.ts
```diff
@@ -55,7 +55,6 @@ export class Polling {
           (err.code === 401 || err.code === 409)
         ) {
           this.skipOffsetSync = true
-          throw err
```
dimankiev, line 58: Redundant `throw`, we already throwing error at the end of `catch` block

src/markup.ts
```diff
@@ -92,34 +92,34 @@ interface KeyboardBuildingOptions<B extends HideableKBtn | HideableIKBtn> {
...
-  }
+  if (!Array.isArray(buttons)) return result
   if (is2D(buttons)) {
     return buttons.map((row) => row.filter((button) => !button.hide))
   }
-  const wrapFn =
-    options.wrap !== undefined
-      ? options.wrap
-      : (_btn: B, _index: number, currentRow: B[]) =>
-          currentRow.length >= options.columns
```
dimankiev, line 110: Redundant creation a function on every parent function call
It cause performance issues (even if they unnoticeable)

src/markup.ts
```diff
@@ -92,34 +92,34 @@ interface KeyboardBuildingOptions<B extends HideableKBtn | HideableIKBtn> {
...
   let currentRow = []
   let index = 0
   for (const btn of buttons.filter((button) => !button.hide)) {
-    if (wrapFn(btn, index, currentRow) && currentRow.length > 0) {
```
dimankiev, line 114: Unused parameters in `wrapFn`

src/markup.ts
```diff
@@ -92,34 +92,34 @@ interface KeyboardBuildingOptions<B extends HideableKBtn | HideableIKBtn> {
...
+    if (wrapFn(options, currentRow) && currentRow.length > 0) {
       result.push(currentRow)
       currentRow = []
     }
     currentRow.push(btn)
     index++
   }
-  if (currentRow.length > 0) {
-    result.push(currentRow)
-  }
```
dimankiev, line 123: Can be done in single line

src/session.ts
```diff
@@ -23,9 +23,7 @@ export function session<S extends object>(
   const storage = options?.storage ?? new MemorySessionStorage()
   return async (ctx, next) => {
     const key = await makeKey(ctx)
-    if (key == null) {
-      return await next()
-    }
```
dimankiev, line 28: Can be done in single line

src/session.ts
```diff
@@ -55,9 +51,8 @@ export class MemorySessionStorage<T> implements Storage<T> {
 
   async getItem(name: string): Promise<T | undefined> {
     const entry = this.store.get(name)
-    if (entry == null) {
-      return undefined
-    } else if (entry.expires < Date.now()) {
```
dimankiev, line 60: Redundant `if else`, execution will stop after return in cause of `entry == null`

src/session.ts
```diff
@@ -39,9 +37,7 @@ export function session<S extends object>(
 async function makeDefaultKey(ctx: Context): Promise<string | undefined> {
   const fromId = ctx.from?.id
   const chatId = ctx.chat?.id
-  if (fromId == null || chatId == null) {
-    return undefined
-  }
```
dimankiev, line 44: Can be done in single line

src/markup.ts
```diff
@@ -92,34 +92,34 @@ interface KeyboardBuildingOptions<B extends HideableKBtn | HideableIKBtn> {
   columns: number
 }
 
+function wrapFn<B extends HideableKBtn | HideableIKBtn>(
+    options: KeyboardBuildingOptions<B>,
+    currentRow: B[]
+  ): any
```
wojpawlik, line 98: Unacceptable `any`.

src/context.ts
wojpawlik: Optional chaining doesn't help, these are not null checks, they're narrowing an union type, hence CI fails.

src/markup.ts
wojpawlik: Might be used by `options.wrap`.

wojpawlik (changes requested):
Anything I didn't reply specifically reply to is just a style difference, thanks but no thanks.

src/context.ts
wojpawlik: Optional chaining doesn't help, these are not null checks, they're narrowing an union type, hence CI fails.

Issue #892, opened 2020-02-08
fetchUpdates() blocks with long running update Handlers


Sheco:
## Context

I am writing a bot that sleeps for some seconds before sending a reply, when using the polling mechanism, I noticed that the bot would stop fetching updates while there was an update handler that hadn't finished.

* Telegraf.js Version: 3.36.0
* Node.js Version: v10.15.2
* Operating System: Debian 10

## Expected Behavior

When executing an update handler that takes some time to finish, the bot should continue being able to get new updates using the polling mechanism.

## Current Behavior

The bot blocks and doesn't process anything up until the update handlers finish.

## Failure Information (for bugs)

The fetchUpdates() method in telegraf.js blocks on line 190 and doesn't pass the execution to the next block which calls fetchUpdates again.

### Steps to Reproduce

use this test.js file:
```
function sleep (ms) {
  return new Promise((resolve) => {
    setTimeout(resolve, ms)
  })
}   
      
const Telegraf = require('telegraf')
const bot = new Telegraf(process.env.BOT_TOKEN)
bot.start(async (ctx) => {
  ctx.reply('Sleeping')
  await sleep(5000)
  ctx.reply('Woke up')
})        
bot.launch()
```

1. launch the bot
```
BOT_TOKEN=1234:aaaaaaaaaaaaaaaaaa DEBUG=* node test.js
```
2. send two /start messages to the bot
3. see how the bot responds

### Failure Logs

```
--- current state, not working, two parallel messages
  telegraf:core Connecting to Telegram +0ms
  telegraf:client HTTP call getMe {} +0ms
  telegraf:core Launching @FabiCarbot +868ms
  telegraf:client HTTP call deleteWebhook {} +867ms
  telegraf:client HTTP call getUpdates?offset=0&limit=100&timeout=30 {} +321ms
  telegraf:core Bot started with long-polling +322ms
  telegraf:core Processing update 573133527 +10s
  telegraf:client HTTP call sendMessage { chat_id: 14103853, text: 'Sleeping' } +10s
  telegraf:client HTTP call sendMessage { chat_id: 14103853, text: 'Woke up' } +5s
  telegraf:client HTTP call getUpdates?offset=573133528&limit=100&timeout=30 {} +0ms
  telegraf:core Processing update 573133528 +6s
  telegraf:client HTTP call sendMessage { chat_id: 14103853, text: 'Sleeping' } +841ms
[...]

PR #1276, opened 2021-01-05
Improve update handling

KnorpelSenf:
# Description

This PR is yet another attempt to improve the update handling of `telegraf` (and it is hopefully the last one before 4.0). It takes all previous work on this issue into account, namely #1247 (which again is based on #1234), #1252, and #1262, but this PR is mostly a merge of #1247 and #1262. It has a minimal diff to both branches (on the touched files). It fixes all issues that #1247 fixes while staying close to #1262 whenever possible.

# Checklist:

### DIFF
--- a/src/core/network/polling.ts
+++ b/src/core/network/polling.ts
@@ -15,9 +15,11 @@ export class Polling {
   private readonly abortController = new AbortController()
   private skipOffsetSync = false
   private offset = 0
+  public limit: number | undefined
   constructor(
     private readonly telegram: ApiClient,
-    private readonly allowedUpdates: readonly tt.UpdateType[]
+    private readonly allowedUpdates: readonly tt.UpdateType[],
+    private readonly timeout: number
   ) {}
 
   private async *[Symbol.asyncIterator]() {
@@ -27,9 +29,10 @@ export class Polling {
         const updates = await this.telegram.callApi(
           'getUpdates',
           {
-            timeout: 50,
+            timeout: this.timeout,
             offset: this.offset,
             allowed_updates: this.allowedUpdates,
+            limit: this.limit,
           },
           this.abortController
         )
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -261,7 +261,7 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
    * Number of additional updates that may be processed until the `concurrency`
    * limit is reached
    */
-  capacity(): number {
+  capacity() {
     return this.concurrency - this.processing
   }
 
@@ -270,10 +270,11 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
    * becomes positive
    */
   bored(): Promise<number> {
-    const capacity = this.capacity()
-    // Resolve immediately if there is free space, otherwise postpone this
-    return capacity > 0
-      ? Promise.resolve(capacity)
-      : new Promise((resolve) => this.subscribers.push(resolve))
+    return new Promise((resolve) => {
+      const capacity = this.capacity()
+      // Resolve immediately if there is free space, otherwise postpone this
+      if (capacity > 0) resolve(capacity)
+      else this.subscribers.push(resolve)
+    })
   }
 }
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -10,22 +10,22 @@ import Context from './context'
 import d from 'debug'
 import generateCallback from './core/network/webhook'
 import { Polling } from './core/network/polling'
+import pTimeout from 'p-timeout'
 import Telegram from './telegram'
 import { TlsOptions } from 'tls'
 import { URL } from 'url'
 const debug = d('telegraf:main')
 
 const DEFAULT_OPTIONS: Telegraf.Options<Context> = {
-  telegram: {},
   handlerTimeout: 5000,
   contextType: Context,
+  concurrency: false,
 }
 
 function always<T>(x: T) {
   return () => x
 }
 const anoop = always(Promise.resolve())
-const wait = util.promisify(setTimeout)
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 namespace Telegraf {
@@ -34,12 +34,17 @@ namespace Telegraf {
       ...args: ConstructorParameters<typeof Context>
     ) => TContext
     handlerTimeout: number
+    concurrency: boolean | number
     telegram?: Partial<ApiClient.Options>
   }
 
   export interface LaunchOptions {
     /** List the types of updates you want your bot to receive */
     allowedUpdates?: tt.UpdateType[]
+    /** Configuration options for when the bot is run via long polling */
+    polling?: {
+      timeout?: number
+    }
     /** Configuration options for when the bot is run via webhooks */
     webhook?: {
       /** Public domain for webhook. If domain is not specified, hookPath should contain a domain name as well (not only path component). */
@@ -77,6 +82,10 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
     throw err
   }
 
+  private processing = 0
+  private readonly concurrency: number
+  private subscribers: Array<(capacity: number) => void> = []
+
   constructor(token: string, options?: Partial<Telegraf.Options<C>>) {
     super()
     // @ts-expect-error
@@ -85,6 +94,11 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
       ...compactOptions(options),
     }
     this.telegram = new Telegram(token, this.options.telegram)
+
+    const concurrency = this.options.concurrency
+    if (concurrency === false) this.concurrency = 1
+    else if (concurrency === true) this.concurrency = Infinity
+    else this.concurrency = Math.max(1, concurrency)
   }
 
   private get token() {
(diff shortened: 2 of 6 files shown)


src/core/network/polling.ts
```diff
@@ -15,9 +15,11 @@ export class Polling {
   private readonly abortController = new AbortController()
   private skipOffsetSync = false
   private offset = 0
+  public limit: number | undefined
   constructor(
     private readonly telegram: ApiClient,
-    private readonly allowedUpdates: readonly tt.UpdateType[]
+    private readonly allowedUpdates: readonly tt.UpdateType[],
+    private readonly timeout: number
```
wojpawlik, line 22: I said 

> It might make sense to make it configurable in the future. (https://github.com/telegraf/telegraf/discussions/1234#discussioncomment-215009)

Not "It makes sense right now"

wojpawlik (commented):
`concurrency` is a mistake.

When using webhook, you cannot tell the server to slow down sending updates, so only infinity makes sense.

When you wish to fetch updates while previous ones are still being processed (`concurrency: >100`), you're better off with webhook.

And when you wish to to confirm updates 1 by 1 (`concurrency: 1`), you're **much** better off with webhook.

Polling _may_ be advantageous only when you need to be _certain_ you're not processing updates out of order.

KnorpelSenf:
> When using webhook, you cannot tell the server to slow down sending updates, so only infinity makes sense.

Wrong. You are able to slow down the server per chat.

> When you wish to fetch updates while previous ones are still being processed (`concurrency: >100`)

This is not how `concurrency` works, I don't know what you mean.

> you're better off with webhook.

Being better or worse off with either choice is something I want to take for my bot myself. I don't want `telegraf` to make the decision how to host my bot.

> And when you wish to to confirm updates 1 by 1 (`concurrency: 1`)

That's again not how `concurrency` works, and not how anyone should do it … I'm really confused now :confused: 

> Polling _may_ be advantageous only when you need to be _certain_ you're not processing updates out of order.

Wrong, there are many other benefits of polling over webhooks. But again, `telegraf` should work well in both environments. I want to have the choice about my own hosting.

src/telegraf.ts
```diff
@@ -191,16 +212,12 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
...
       throw new TypeError(util.format('Updates must be an array, got', updates))
     }
-    // prettier-ignore
-    const processAll = Promise.all(updates.map((update) => this.handleUpdate(update)))
-    if (this.options.handlerTimeout === Infinity) {
-      return processAll
-    }
-    return Promise.race([processAll, wait(this.options.handlerTimeout)])
+    await Promise.all(updates.map((update) => this.handleUpdate(update)))
+    return await this.bored()
```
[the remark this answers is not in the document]
KnorpelSenf: Mostly style preference. You could also return the last element of the array, but I found that less readable.

src/core/network/polling.ts
KnorpelSenf: This is the result of a merge. #1247 could already do it a few weeks ago. I can of course remove it again and reintroduce it in a subsequent PR.

> Not "It makes sense right now"

Do you mean "it makes no sense right now" so we must remove it?

wojpawlik: Yes. Except I see no point in reintroducing it right now.

wojpawlik:
> Wrong. You are able to slow down the server per chat.

How? I'm aware you can wait with ending the reply, but

1. You reported that updates will be re-send if done for too long,
2. You cannot go lower than `max_connections`,
2. How is it per chat?

> This is not how `concurrency` works, I don't know what you mean.

True, let me elaborate.

As of `develop`, Telegraf can either process updates or fetch more, but not both at the same time. Trying to do both results in confirming updates prematurely, which is unsafe. This limitation doesn't exist on webhook.

> > And when you wish to to confirm updates 1 by 1 (`concurrency: 1`)
> 
> That's again not how `concurrency` works, and not how anyone should do it … I'm really confused now 😕

I'm confused as well. What does `concurrency: 1` mean if not fetching and processing and confirming 1 update at a time?

> Wrong, there are many other benefits of polling over webhooks.

Such as?

KnorpelSenf:
I don't get why I must not fix the library for my own use case when it is flawed. I understand that you don't have that use case and that you don't care about it, but there's literally zero difference in behaviour for everyone else including you (unless people want better performance, then they can opt-in).

Trust me, I have a very good reason to spend this much time on the issue. It took a long time to figure out that I cannot solve the issue without fixing the underlying processing, and it would be a real pity to have to migrate.

KnorpelSenf:
> I'm confused as well. What does `concurrency: 1` mean if not fetching and processing and confirming 1 update at a time?

It keeps on processing batches as it always has, each containing multiple updates. `concurrency: 1` resembles exactly the same behaviour as we've always had.

> Such as?

No need for a custom domain, no need for certificates, simpler to determine the load, the list goes on

KnorpelSenf:
> 3\. How is it per chat?

Updates are sent concurrently across chats, but no update is delivered before the previous one of the same chat is confirmed (at least that's what happens for me). If a message takes long to process, the subsequent messages in the same chat can be slowed down. If the request takes too long to complete, it is being resent.

> which is unsafe.

I know that? We talked about this before, and we have to compromise between reliability and performance. If you want to have radical reliability but bad responsiveness, that's fine, it's your choice. Why are you taking the choice away from other people who may have different requirements?

wojpawlik:
The problem you're trying to solve is #892, being able to poll for more updates before current batch is done processing, correct?

Right now I'm mostly worried that `concurrency` is simply confusing.
In plain English, what does the number passed to `concurrency` mean?
What should the default value be? It should work well for both polling and webhook.
Perhaps webhook should outright ignore the value, but I'm worried that'd lead to messy code.

> It keeps on processing batches as it always has, each containing multiple updates. `concurrency: 1` resembles exactly the same behaviour as we've always had.

Not exactly. It'll request less updates than `develop` (100):

https://github.com/telegraf/telegraf/blob/2889ad38a8b46056c1594fdd11fb81f0db00d40c/src/telegraf.ts#L143-L144

> No need for a custom domain, no need for certificates

Heroku takes care of these.

I agree that polling is simpler to set-up, and webhook isn't available everywhere (hello CGNAT my old friend...)

> simpler to determine the load

I disagree. No matter how you fetch updates, you can maintain a count of currently processed updates. If you maintain a set of currently processed `update_id`s, you can even trivially deduplicate incoming webhook updates.

KnorpelSenf:
We can rename it if you have a better suggestion. `concurrency` means "how long must I wait before I may look at the next updates", where "how long" can be expressed as "if the number of currently processed updates is above `concurrency`, then you have to wait longer". The default should be `true` for webhooks and `false` for polling, i.e. you could take `'webhook' in this.options` or something (not implemented yet).

wojpawlik:
I can't help but feel there's a simpler way... First thing that comes to my mind is moving `handleUpdates` to `LaunchOptions['polling']` so that consumers can override it with whatever logic they wish.

KnorpelSenf:
It is much, much easier to simply avoid using `telegraf.ts` in the first place. I know that I am going to do that from now on. You don't want my fixes and I don't want you to wire in half-baked dependencies, hence everyone would be better off with just leaving the file it as it is and describing the situation in the docs.

We should then also write a section about how to write a custom version of `telegraf.ts` that works with the middleware stack, and what to take care of when doing so. This file is already so messy before merging this PR, the timeouts won't have tests (except #1247), and it still has so many oddities and issues beyond that.

Maybe I'll publish a separate package that uses the middleware system of `telegraf` but makes it production ready for all occasions, that would have saved me so much time already. I'll let you know of it if I actually do that.

wojpawlik:
I'm sorry we couldn't reach an agreement. I'm not being malicious. It's just adding quite a lot of complexity to an already complex file.

Writing an alternative from scratch sounds great! If you do it, make sure to let me know before you take up a name on npm: if it's much better than `telegraf.ts`, I might want to bring it into the main package and deprecate `telegraf.ts`.

Perhaps you could allow tighter integration with `telegraf-throttler`. Currently middleware can time out due to retrying failed requests cc @KnightNiwrem 

To reimplement `telegraf.ts`, you need to construct `Context`, and to construct that you currently need `Telegram`, which is currently not exported. I could export `Core` namespace with internal stuff useful for writing libraries on top of Telegraf. Or I could export a function (I don't have a name yet) which'd construct these for you, without handling errors and timeouts:
```ts
(middleware: Middleware<Context>, botInfo: tt.UserFromGetMe) =>
  (update: tt.Update, webhookResponse?: http.ServerResponse) => Promise<void>
```

`Polling` isn't exported either, but I'm not sure if you wish to reuse it.

Since you will avoid `telegraf.ts`, are you fine with merging #1262 and calling it 4.0? I'll document all the gotchas after that. Or maybe even before, depending on Vitaly.

KnorpelSenf:
Sure, feel free, I won't interfere

wojpawlik:
Writing a `telegraf.ts` replacement is definitely **the** way to go. I wish I noticed sooner, it'd save us both time and stress. I'll be glad to help in any way I can. I could even try writing it myself, but I'd rather you do it to make sure it actually solves your problems.

I think the best data structure for in-progress task is something akin to Node stream. With `addMany` method, and `"drain"` event.

`"drain"` event would be emitted when number of currently processed updates dips below `highWaterMark` / `concurrency`, or when `addMany` call fails to bring it above it. `"drain"` event would invoke `getUpdates` when polling, and do nothing in webhook.

With premature update confirming, you may want to wait with rethrowing any errors until all updates finish processing.

wojpawlik:
Timeouts would also be easy to detect: if after `addMany`, `"drain"` is not emitted within `handlerTimeout`, _every_ currently processed update has timed out, and bot has to crash to recover.

wojpawlik:
You can improve performance _and_ correctness of polling by having a dedicated process to fetch updates and put them in a [work queue](https://12factor.net/disposability) in a separate process. Workers would obtain updates from it and acknowledge them once they're handled.

Screw fetching updates via WebSocket, Telegram should simply expose a work queue to obtain updates from!

PR #1280, opened 2021-01-07
Restore 3.38 names in session module

### LINKED ISSUE #1202: 4.0 Roadmap

KnorpelSenf:
# Description

Renames most things in the session module to reduce the number of breaking changes from 3.38.0. Also fixes storage types in the sense that a synchronous storage is allowed, as it was before.

If I'm not mistaken, this is the last pull request before the 4.0 major release, also confer #1202.

## Type of change

Names of types and properties and config in `src/session.ts`.

# How Has This Been Tested?

`npm run clean && npm run checks`

# Checklist:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,6 @@ export { Router } from './router'
 
 export * as Markup from './markup'
 
-export { session, MemorySessionStorage } from './session'
+export { session, MemorySessionStore } from './session'
 
 export * as Scenes from './scenes'
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,15 +1,16 @@
 import { Context } from './context'
+import { MaybePromise } from './composer'
 import { MiddlewareFn } from './middleware'
 
-export interface Storage<T> {
-  getItem: (name: string) => Promise<T | undefined>
-  setItem: (name: string, value: T) => Promise<void>
-  deleteItem: (name: string) => Promise<void>
+export interface SessionStore<T> {
+  get: (name: string) => MaybePromise<T | undefined>
+  set: (name: string, value: T) => MaybePromise<void>
+  delete: (name: string) => MaybePromise<void>
 }
 
 interface SessionOptions<S extends object> {
-  makeKey?: (ctx: Context) => Promise<string | undefined>
-  storage?: Storage<S>
+  getSessionKey?: (ctx: Context) => Promise<string | undefined>
+  store?: SessionStore<S>
 }
 
 export interface SessionContext<S extends object> extends Context {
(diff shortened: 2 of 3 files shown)


wojpawlik (commented):
> now `new Map()` is again a valid session store, as it was before.

Doubt, `Map::set` returns `Map`, not `void`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set#return_value

KnorpelSenf:
> > now `new Map()` is again a valid session store, as it was before.
> 
> Doubt, `Map::set` returns `Map`, not `void`: [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set#return_value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set#return_value)

Right, now I'm also getting compile errors, for some reason I did not before.

Anyway, being able to pass a `Map` in was a side-effect, and not the desire when changing the types. The only aim for me was to be compliant with the previous naming.

Hence, I'd leave it incompatible with `Map`. The alternative would be to allow returning `unknown` which would work well in practise but be inconsistent with the types of `MiddlewareFn` where we had 1:1 the same discussion.

wojpawlik (approved):

PR #1281, opened 2021-01-08
add copyMessage method

LetItCode:
Telegram have new native method like `sendCopy`: https://core.telegram.org/bots/api#copymessage

p.s. I think we need adaptor for current `sendCopy` too


KnorpelSenf:
Your PR targets `master`, however, the development happens on `develop` which should also be the default branch. It already contains `copyMessage`, can you confirm this?

KnorpelSenf:
https://github.com/telegraf/telegraf/blob/e78dae478bba42d3a184f47209775cff0bf43f64/src/telegram.ts#L981

LetItCode:
you have wrong concept in your ts branch
```const message = this.message ||
this.editedMessage ||
this.channelPost ||
this.editedChannelPost ||
(this.callbackQuery && this.callbackQuery.message)
```

`bot.hears('shit', ctx => ctx.copyMessage(ctx.from.id))` will not working from channel etc

LetItCode:
GIVE👏US👏OUR👏JS

KnorpelSenf:
`bot.hears` generally does not trigger for channel posts, but assuming that you meant `bot.on('channel_post')`, it does look like you're right and that `copyMessage` won't work then.

I reapplied your changes to `develop`. Could you check out #1282 and see if it does the job?

PR #1282, opened 2021-01-08
Fix copyMessage not working for channels

KnorpelSenf:
# Description

This PR fixes https://github.com/telegraf/telegraf/pull/1281#issuecomment-756644587.

TypeScript 4.1.2 seems to have bug:

```ts
type U = { a: string } | { b: number }
const c: U = { b: 42 }
function f(u: U) {
  if ('a' in u) return u.a
  else return undefined
}
console.log(f(c).a.length) // compiles but throws
```

This is fixed in 4.1.3. The getters on `Context` were affected by this, they did not return `undefined`, which was incorrect. This PR therefore also bumps the TS version (and takes the chance to update the other deps).

In addition, `ctx.senderChat` is added, which is useful since Bot API 5.0.

# How Has This Been Tested?

`npm run checks`

# Checklist:


src/context.ts
```diff
@@ -578,24 +566,53 @@ export class Context {
...
+}
+
+function getQueryFromAnySource(ctx: Context) {
+  return (
+    ctx.callbackQuery ??
+    ctx.inlineQuery ??
+    ctx.shippingQuery ??
+    ctx.preCheckoutQuery
+  )
+}
```
wojpawlik, line 610: This function is only called by `getUserFromAnySource`, I'd rather inline this into `getUserFromAnySource` that, or into `ctx.from`.

wojpawlik (changes requested):

src/context.ts
wojpawlik: Preferably into `ctx.from`

wojpawlik (approved):

PR #1294, opened 2021-01-14
Add URL import for older targets

noneedinmagic:
# Description

## Type of change

# How Has This Been Tested?

# Checklist:


src/core/network/client.ts
```diff
@@ -3,6 +3,7 @@ import * as crypto from 'crypto'
 import * as fs from 'fs'
 import * as http from 'http'
 import * as https from 'https'
+import { URL } from 'url'
```
wojpawlik, line 6: Please move this lower so that imports are sorted and eslint doesn't complain

wojpawlik (approved):

PR #1296, opened 2021-01-14
Remove `Telegraf::handleUpdates`

wojpawlik:
One closure and one private method less.

# Checklist:

### DIFF
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -3,7 +3,6 @@ import * as http from 'http'
 import * as https from 'https'
 import * as tg from './core/types/typegram'
 import * as tt from './telegram-types'
-import * as util from 'util'
 import { Composer, MaybePromise } from './composer'
 import ApiClient from './core/network/client'
 import { compactOptions } from './core/helpers/compact'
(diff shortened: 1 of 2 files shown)


src/telegraf.ts
```diff
@@ -92,11 +91,11 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
     return this.telegram.token
   }
 
-  /** @deprecated use `ctx.telegram.webhookReply` */
   set webhookReply(webhookReply: boolean) {
     this.telegram.webhookReply = webhookReply
   }
 
+  /** @deprecated use `ctx.telegram.webhookReply` */
```
EdJoPaTo, line 143: Changing these deprecation hints looks unrelated to the PR? (same with L95)
Also I'm not sure why one (set or get) is deprecated but not both?

wojpawlik: Thanks for reviews!

Yes, it's unrelated. I just fix small issues in the files I'm already editing.

Getter has to be deprecated, otherwise someone can think that the property is becoming read-only. Setter does not, write-only properties make very little sense, but might as well.

MKRhere (approved):

PR #1300, opened 2021-01-16
Remove `MessageSubTypes`

wojpawlik:
# Description

-1.3 kB install size
+1 maintainability
+1 readability: `"noErrorTruncation"` prevents VSCode from trimming type hints shown on hover

# How Has This Been Tested?

Verified that type remains the same, only adds previously missing `"proximity_alert_triggered"`.

# Checklist:

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -4,6 +4,7 @@ import * as tt from './telegram-types'
 import { Middleware, MiddlewareFn, MiddlewareObj } from './middleware'
 import Context from './context'
 import { SnakeToCamelCase } from './core/helpers/string'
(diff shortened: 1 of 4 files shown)


KnorpelSenf (approved):
Since when does install size matter?

I don't ŕeally understand how the points listed above are improvements, but on the other hand, I also don't think it seriously breaks something, so go ahead …

wojpawlik:
Since we advertise ourselves as "[Lightweight](https://packagephobia.com/result?p=telegraf,node-telegram-bot-api)" :)

Sub-types available in `.on` now depend only on `typegram`, improving maintainability.

PR #1311, opened 2021-01-19
Fix fastify-webhook-bot example

erfanium:
# Description

The old example won't show something if server has a problem (e.g. when port is busy)

# How Has This Been Tested?

Believe me 

# Checklist:

### DIFF
--- a/docs/examples/fastify-webhook-bot.js
+++ b/docs/examples/fastify-webhook-bot.js
@@ -22,6 +22,4 @@ bot.telegram.setWebhook(WEBHOOK_URL).then(() => {
   console.log('Webhook is set on', WEBHOOK_URL)
 })
 
-app.listen(PORT, () => {
-  console.log('Start listening on port', PORT)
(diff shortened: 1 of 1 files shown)


docs/examples/fastify-webhook-bot.js
```diff
@@ -22,6 +22,4 @@ bot.telegram.setWebhook(WEBHOOK_URL).then(() => {
   console.log('Webhook is set on', WEBHOOK_URL)
 })
 
-app.listen(PORT, () => {
-  console.log('Start listening on port', PORT)
-})
+app.listen(PORT).then(console.log)
```
wojpawlik, line 25: Printing bare address would be confusing, how about

```suggestion
app.listen(PORT).then(() => {
  console.log('Start listening on port', PORT)
})
```
or even just
```suggestion
app.listen(PORT)
```

wojpawlik (approved):

PR #1315, opened 2021-01-21
Set api root from environments

smaznet:
Sometimes maybe a user use Telegram instance multi times and for each of them user should set apiRoot 
if user forgot to set that field , bot uses official api and then maybe some updates not delivered to bot

## Type of change

# How Has This Been Tested?

CI typechecks, lints and runs the test suite.

# Checklist:

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -60,7 +60,7 @@ const DEFAULT_EXTENSIONS: Record<string, string | undefined> = {
 }
 
 const DEFAULT_OPTIONS: ApiClient.Options = {
-  apiRoot: 'https://api.telegram.org',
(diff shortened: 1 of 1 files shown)


smaznet:
i don't know libraries can use environments or no :thinking:

KnightNiwrem:
> i don't know libraries can use environments or no 🤔

They should be able to. The `process` object should be global.

wojpawlik:
The real question is, should they? And the answer is, usually no, but there are valid exceptions (https://www.npmjs.com/package/debug).

This could be useful for redirecting all bots on one machine to local bot API, but that's not the problem @smaznet is trying to solve.

PR #1328, opened 2021-01-26
Cache `node_modules/` in CI

wojpawlik:
# Description

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -19,6 +19,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        persist-credentials: false
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v2
       with:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -29,7 +29,7 @@ jobs:
       id: cache
       with:
         path: node_modules/
-        key: ${{ runner.os }}-node${{ matrix.node-version }}-${{ hashFiles('package-lock.json') }}
+        key: ${{ runner.os }}-node${{ matrix.node-version }}-${{ hashFiles('package*.json') }}
     - run: npm ci --ignore-scripts
       if: steps.cache.outputs.cache-hit != 'true'
     - run: npm test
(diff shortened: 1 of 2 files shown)


wojpawlik:
Caching `~/.npm` provides just <=1s speedup lol, suggesting that npm spends most time on disk IO, I don't think it's worth bothering. Perhaps caching `node_modules` could be faster, but [it's discouraged](https://github.com/actions/cache/blob/main/examples.md#node---npm).

I'll upgrade `actions/setup-node` to `v2` next time I touch the file, it shaves off 2 seconds with no extra code.

wojpawlik:
I'll try caching `node_modules/` instead, but only if `package-lock.json` haven't changed and Node versions match.

wojpawlik:
Bringing `node_modules/` from cache takes just 3 seconds. It's done only if Node version and hash of `package-lock.json` match. Otherwise a fresh `npm ci --ignore-scripts` is performed, without any caching, which takes 6 to 9 seconds.

wojpawlik:
If this works well, I'll encapsulate it in a GitHub Action for easier reuse. If it does not, I'll revert it.

KnorpelSenf:
Not quite sure which problem this PR solves. If it's just regular config maintenance (upgrading tooling), then sure, go ahead. It seems like you're trying to make CI faster, though. Does this matter? Caching files ignored by git `node_modules` really is not a good idea. There should be a very strong point why another setup is not possible before you buy in a new class of problems.

wojpawlik:
Faster CI improves contributor experience, and incentivizes parallelizing the checks.

Why is caching `node_modules/` a bad idea?

I couldn't find a better way, aside from switching package manager, which would be much more disruptive.

KnorpelSenf:
> Why is caching `node_modules/` a bad idea?

Because you by definition have no reproducibility of the builds anymore. If something breaks, it takes a lot of effort to guarantee that the problem is not related to caching. (You may be familiar with https://shouldiblamecaching.com/). Having CI build from scratch is super nice and makes you sleep well at night.

wojpawlik:
I find issues unlikely, I am defensive with cache key: https://github.com/telegraf/telegraf/blob/f4a30fdbd0efb3b973eada3c124662dca78450ea/.github/workflows/node.js.yml#L32

So I'm willing to take this risk. And to deal with any consequences that may arise, and learn from the experience.

PR #1334, opened 2021-01-29
feat(client): add `apiMode` option

Morb0:
# Description

Add feature to select `user` or `bot` type in api url to support `userbot`  from [TDLight](https://github.com/tdlight-team/tdlight-telegram-bot-api#user-mode)

## Type of change

# How Has This Been Tested?

Local bot start.
Run tests & lint.

**Test Configuration**:
* Node.js Version: v15.5.1

# Checklist:

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -344,7 +344,7 @@ class ApiClient {
           options.attachmentAgent
         )
       : await buildJSONConfig(payload)
-    const apiUrl = `${options.apiRoot}/${options.apiType}${token}/${method}`
+    const apiUrl = new URL(`${options.apiType}${token}/${method}`, options.apiRoot)
     config.agent = options.agent
     config.signal = signal
     config.timeout = 60_000 // 60s in ms
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -344,7 +344,10 @@ class ApiClient {
           options.attachmentAgent
         )
       : await buildJSONConfig(payload)
-    const apiUrl = new URL(`${options.apiType}${token}/${method}`, options.apiRoot)
+    const apiUrl = new URL(
+      `${options.apiType}${token}/${method}`,
+      options.apiRoot
+    )
     config.agent = options.agent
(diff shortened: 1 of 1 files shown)


src/core/network/client.ts
```diff
@@ -338,7 +344,7 @@ class ApiClient {
           options.attachmentAgent
         )
       : await buildJSONConfig(payload)
-    const apiUrl = `${options.apiRoot}/bot${token}/${method}`
+    const apiUrl = `${options.apiRoot}/${options.apiType}${token}/${method}`
```
wojpawlik: While the line is changed anyway, how about stripping handling trailing slash in `apiRoot` by
```suggestion
    const apiUrl = new URL(`${options.apiType}${token}/${method}`, options.apiRoot)
```

wojpawlik (changes requested):
Fixed all of the problems, with one exception: I don't like `apiType` name. How about `apiMode`?

Morb0:
@wojpawlik `apiMode` is also an appropriate name. I used `apiType` because in the [API documentation](https://tdlight-team.github.io/tdlight-telegram-bot-api/) this variable was called as `type`.

src/core/network/client.ts
Morb0: I don't see much point in this. It looks like it still needs to be converted to a `string` because I think `fetch` takes a `string`.

Morb0:
@wojpawlik So what about this?

src/core/network/client.ts
wojpawlik: Incorrect. `fetch` accepts `URL` too.

wojpawlik:
Sorry not-sorry, I needed a little extra rest.

I still think `apiMode` is a better name. The readme section you linked names the feature "User Mode", not "User API", and definitely not "User API Type", which just sounds awkward... what do you think @KnorpelSenf?

Morb0:
Agree, seems to be better, make a commit.

wojpawlik (approved):

PR #1339, opened 2021-02-01
Fix `startPayload` for commands with username

vrumger:
# Description

`ctx.startPayload` would include the bots username if you used `/start@bot payload`.

## Type of change

# Checklist:

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -211,7 +211,8 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
   start(...fns: ReadonlyArray<Middleware<C & { startPayload: string }>>) {
     const handler = Composer.compose(fns)
     return this.command('start', (ctx, next) => {
-      const startPayload = ctx.message.text.substring(7)
+      const entity = ctx.message.entities![0]!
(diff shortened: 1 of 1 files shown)


src/composer.ts
```diff
@@ -211,7 +211,11 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
   start(...fns: ReadonlyArray<Middleware<C & { startPayload: string }>>) {
     const handler = Composer.compose(fns)
     return this.command('start', (ctx, next) => {
-      const startPayload = ctx.message.text.substring(7)
+      const entity = ctx.message.entities?.find(
+        (entity) => entity.type === 'bot_command'
+      )
+      const startPayload =
+        entity !== undefined ? ctx.message.text.slice(entity.length + 1) : ''
```
wojpawlik: First entity is guaranteed to exist and be of type `bot_command`:

```suggestion
      const entity = ctx.message.entities![0]!
      const startPayload = ctx.message.text.slice(entity.length + 1)
```

wojpawlik (approved):

PR #1340, opened 2021-02-01
Explicitly define scene typings subpath

igorrybak:
# Description

If we start the project with scenes, stage and custom context we get an error:
> Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './typings/scenes' is not defined by "exports" in /project/node_modules/telegraf/package.json

The code example is:
```
import { Context, Scenes, Telegraf } from "telegraf";
import { BaseScene, Stage } from "telegraf/typings/scenes";

interface CustomContext extends Context {
    scene: Scenes.SceneContextScene<CustomContext>;
}

const bot = new Telegraf<CustomContext>(process.env.TELEGRAM_TOKEN!);

const start = new BaseScene<CustomContext>("start-scene");
start.enter(async (ctx) => {
    await ctx.reply("Choose action");
});

const stage = new Stage<CustomContext>([start]);
bot.use(stage.middleware());

bot.start(async (ctx) => {
    await ctx.scene.enter("start-scene");
});
bot.launch();
```
So maybe the solution is to explicitly define exporting subpath for scenes typings.

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* Node.js Version: 14.15.3
* Operating System: Ubuntu 20.04.1 LTS
* Telegraf: 4.0.2

With this changes project with the Telegraf library starts without errors and works as expected.

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,11 @@
   },
   "main": "lib/index.js",
   "exports": {
-    "import": "./lib/index.mjs",
-    "require": "./lib/index.js"
+    ".": {
+      "import": "./lib/index.mjs",
+      "require": "./lib/index.js"
+    },
+    "./typings": "./lib/scenes/index.js"
   },
   "files": [
     "bin/*",


wojpawlik:
```ts
import { Scenes } from "telegraf";
```

igorrybak:
> ```ts
> import { Scenes } from "telegraf";
> ```

It does not work as showed in the code above. It's my bad I didn't show the whole stack trace, but the error there points to the second line where we are importing from typings:
```ts
import { BaseScene, Stage } from "telegraf/typings/scenes";
```

wojpawlik:
```ts
const { BaseScene, Stage } = Scenes;
```

igorrybak:
Wow, it's so easy) So we have to get Stage and BaseScene from previously imported Scenes instead of importing from "telegraf/typings". Thank you very much @wojpawlik !

PR #1345, opened 2021-02-03
Fix koa example

alsotang:
1. it should not handle all `POST` requests
2. return 200. otherwise telegram server would think it's down and retry many times.

# Description

current implement doesn't work

## Type of change

bug fix

# How Has This Been Tested?

I am using the solution

# Checklist:

### DIFF
--- a/docs/examples/koa-webhook-bot.js
+++ b/docs/examples/koa-webhook-bot.js
@@ -23,8 +23,12 @@ bot.telegram.setWebhook('https://-----.localtunnel.me/secret-path')
 const app = new Koa()
 app.use(koaBody())
 // @ts-ignore
-app.use((ctx, next) => ctx.method === 'POST' || ctx.url === '/secret-path'
-  ? bot.handleUpdate(ctx.request.body)
-  : next()
-)
+app.use(async (ctx, next) => {
+  if (ctx.url === '/secret-path') {
+    await bot.handleUpdate(ctx.request.body)
+    ctx.status = 200
+    return
+  }
+  return next()
+})
 app.listen(3000)


alsotang:
seems the 2nd arg in `bot.handleUpdate(ctx.request.body, ctx.res)` doesn't  work correctly. it wont reply to the webhook req.

i would fix it this weekend.

alsotang:
`bot.handleUpdate` only response to several methods. for a method that is not in the allow list, nothing would be done with the response object.

docs/examples/koa-webhook-bot.js
```diff
@@ -9,7 +9,11 @@ if (token === undefined) {
   throw new Error('BOT_TOKEN must be provided!')
 }
 
-const bot = new Telegraf(token)
+const bot = new Telegraf(token, {
+  telegram: {
+    webhookReply: false
```
wojpawlik: No-op without `res`

docs/examples/koa-webhook-bot.js
```diff
@@ -23,8 +27,12 @@ bot.telegram.setWebhook('https://-----.localtunnel.me/secret-path')
 const app = new Koa()
 app.use(koaBody())
 // @ts-ignore
-app.use((ctx, next) => ctx.method === 'POST' || ctx.url === '/secret-path'
-  ? bot.handleUpdate(ctx.request.body)
-  : next()
-)
+app.use((ctx, next) => {
+  if (ctx.url === '/secret-path') {
```
wojpawlik, line 27:
```suggestion
  if (ctx.method === 'POST' && ctx.url === '/secret-path') {
```

docs/examples/koa-webhook-bot.js
```diff
@@ -23,8 +27,12 @@ bot.telegram.setWebhook('https://-----.localtunnel.me/secret-path')
...
+    bot.handleUpdate(ctx.request.body)
```
wojpawlik:
```suggestion
    await bot.handleUpdate(ctx.request.body)
```

wojpawlik (changes requested):

docs/examples/koa-webhook-bot.js
alsotang: dont understand

alsotang: since the `bot` only responds in specific conditions. so maybe disable webhook is a more generic solution.

docs/examples/koa-webhook-bot.js
alsotang: not matters. when matching the secret path, delegate all methods to the handler.

docs/examples/koa-webhook-bot.js
alsotang: it the final exprssion before return. so we can ignore `await`

wojpawlik: It affects backpressure and error handling.

alsotang: ok. i can fix it now

docs/examples/koa-webhook-bot.js
wojpawlik: This option does nothing when you're not passing 2nd argument to `bot.handleUpdate`, so revert this to not complicate the code.

alsotang: ok

wojpawlik (approved):

PR #1348, opened 2021-02-04
feat(client): add timeout option

Morb0:
# Description

With the current constant timeout impossible to upload big files.

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -47,6 +47,7 @@ namespace ApiClient {
      */
     apiMode: 'bot' | 'user'
     webhookReply: boolean
+    timeout: number
   }
 
   export interface CallApiOptions {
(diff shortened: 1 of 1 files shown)


wojpawlik:
I'd rather not, `node-fetch@3` drops `timeout` entirely: https://github.com/node-fetch/node-fetch/issues/523

wojpawlik:
[Transforming timeout into signal](https://github.com/node-fetch/timeout-signal) when needed is straightforward. However, [combining signals](https://github.com/jacobheun/any-signal) is not. Alternative (in `node-fetch@3`) is to drop the timeout when `signal` is passed, which [defeats the point](https://github.com/telegraf/telegraf/pull/1224#issuecomment-742820310).

How about we just drop the fetch timeout entirely for now, and rely only on existing [`handlerTimeout`](https://telegraf.js.org/interfaces/telegraf.options.html#handlertimeout), @KnorpelSenf?

KnorpelSenf:
Go ahead, I will keep myself out of the discussion around timeouts (and update handling). It's OK if you skip tagging me

wojpawlik:
I'll do that then. API calls could be aborted after `handlerTimeout`... but that's a separate discussion.

Issue #1297, opened 2021-01-15
WizardScene doesn't execute first step after entering


aedeveloper:
## Minimal Example Code Reproducing the Issue

```js
const newQuizWizard = new Scenes.WizardScene(
  "new-quiz-wizard",
  (ctx) => {
    ctx.reply("step 1");
    return ctx.wizard.next();
  },
  (ctx) => {
    ctx.reply("step 2");
    return ctx.scene.leave();
  }
);

const stage = new Scenes.Stage([newQuizWizard]);

bot.use(session());
bot.use(stage.middleware());

bot.command("newquiz", (ctx) => {
  return ctx.scene.enter("new-quiz-wizard");
});

bot.launch();

```

## Expected Behavior
the scene first step should be started after calling `ctx.scene.enter("new-quiz-wizard")`

## Current Behavior

I have to send one more message to the bot to start the first step
however when I set the default wizard option. this issue doesn't occur.

Telegraf version 4.0.0.

aedeveloper:
In version 3 it wasn't behave like that.

also I think [the examples](https://github.com/telegraf/telegraf/tree/develop/docs/examples/wizards) and documentation needs some updates.

I know `default` could lead to problems. but what I mean is the behavior changes. it was a little bit confusing.

foaudkajj:
I am facing the same problem. It was working properly before updating.
Have you solved it? @aedeveloper

aedeveloper:
for now as workaround move first step to `newQuizWizard.enter` as mentioned above by @Autapomorph.

foaudkajj:
I tried it. Now it doesn't continue to the second step.

linrium:
I have the same problem. I must enter one message.

LetItCode:
https://github.com/telegraf/telegraf/blob/develop/src/scenes/base.ts


```ts
enter(...fns: Array<Middleware<C>>) {
  this.enterHandler = compose([this.enterHandler, this.middleware, ...fns])
  return this
}

enterMiddleware() {
  return this.enterHandler
}
```

wojpawlik:
Duplicate of #1333

roshankshaw:
Can somebody provide what should be the correct code for the given code in the issue for telgraf 4?

SwapnilSoni1999:
> Can somebody provide what should be the correct code for the given code in the issue for telgraf 4?

same it broke again in v4

PR #1349, opened 2021-02-06
Call `WizardScene::middleware()` on enter

### LINKED ISSUE #1333: Invoking ctx.scene.enter() on a WizardScene doesn't actually start the scene until the next update
It does create the state within the session, and sets the current scene within the state, but doesn't set the cursor, and more importantly, never even enters the first step until the next update.

I tracked this down to the following code within SceneContextScene.enter():

```javascript
const handler =
      'enterMiddleware' in this.current &&
      typeof this.current.enterMiddleware === 'function'
        ? this.current.enterMiddleware()
        : this.current.middleware()
```

This predicate is <i>always</i> true: 
```javascript
'enterMiddleware' in this.current &&
      typeof this.current.enterMiddleware === 'function'
```

And therefore handler is always set to `this.current.enterMiddleware()`, which returns `this.current.enterHandler`, which is, in turn, a passthru composer that simply wastes the current request doing nothing.

I am currently working around this by setting:
```javascript
Scenes.WizardScene.prototype.enterMiddleware = Scenes.WizardScene.prototype.middleware;
```

Just before defining my `WizardScene`s.

Loskir:
# Description

Hopefully fixes #1333 

## Type of change

# How Has This Been Tested?

typecheck and ava

# Checklist:

### DIFF
--- a/src/scenes/wizard/index.ts
+++ b/src/scenes/wizard/index.ts
@@ -5,6 +5,8 @@ import Composer from '../../composer'
 import Context from '../../context'
 import SceneContextScene from '../context'
 
+const { compose } = Composer
+
 export class WizardScene<
     C extends Context & {
       scene: SceneContextScene<C, WizardSessionData>
@@ -54,5 +56,9 @@ export class WizardScene<
         return Composer.unwrap(ctx.wizard.step)(ctx, next)
       },
     ])
+  }
+
+  enterMiddleware() {
+    return compose([this.enterHandler, this.middleware()])
   }
 }
--- a/src/scenes/wizard/index.ts
+++ b/src/scenes/wizard/index.ts
@@ -5,8 +5,6 @@ import Composer from '../../composer'
 import Context from '../../context'
 import SceneContextScene from '../context'
 
-const { compose } = Composer
-
 export class WizardScene<
     C extends Context & {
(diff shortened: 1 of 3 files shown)


wojpawlik (approved):
Thank you for saving the day, this does fix #1297.

Please move the new tests to a separate file, then I'll merge.

src/scenes/wizard/index.ts
```diff
@@ -55,4 +57,8 @@ export class WizardScene<
       },
     ])
   }
+
+  enterMiddleware() {
+    return compose([this.enterHandler, this.middleware()])
```
wojpawlik:
```suggestion
    return Composer.compose([this.enterHandler, this.middleware()])
```

src/scenes/wizard/index.ts
```diff
@@ -5,6 +5,8 @@ import Composer from '../../composer'
 import Context from '../../context'
 import SceneContextScene from '../context'
 
+const { compose } = Composer
+
```
wojpawlik:
```suggestion
```

wojpawlik:
Would you mind moving the new tests to `test/scenes.js`?

Loskir:
Good idea

PR #1364, opened 2021-02-14
Add an example for Google Cloud Functions

### LINKED ISSUE #1369: Support vercel and nextjs Req and Res formats
Feature request, like there a lot of providers that uses custom AWS functions that wraps basic stuff.
`req` `res` etc.

Redefining `generateCallback` will make sure that get full response so people can use telegraf, make it rewritable or as an option on configuration when you working on AWS lambda env.
```js
export default function (
    hookPath,
    updateHandler
) {
    return async (
        req,
        res,
        next = () => {
            res.status(403);
        }
) => {
        if (req.method !== 'POST' || req.url !== hookPath) {
            return next()
        }
        let update = req.body;

        try {
            await updateHandler(update, res)
        } catch (err) {
            res.status(500);
            throw err
        }
    }
}
```

For example: 
```js
const bot = new Telegraf(token, {
    telegram: { webhookReply: true, webhookReplyGenerator: (req, res) => req.body  }
})
```

```ts
  webhookCallback(path = '/') {
   const fn = this.telegram.webhookReply && this.telegram.webhookReplyGenerator  ? this.telegram.webhookReplyGenerator : generateCallback;
    return fn(
      path,
      (update: tt.Update, res: http.ServerResponse) =>
        this.handleUpdate(update, res)
    )
  }
```

-
**Additional context**
-

jacobwaller:
# Description

## Type of change
Small change to include a new example

# How Has This Been Tested?
Ran linter, no functional changes

# Checklist:

### DIFF
--- a/docs/examples/google-cloud-function.ts
+++ b/docs/examples/google-cloud-function.ts
@@ -1,12 +1,15 @@
 import { Telegraf } from 'telegraf'
 
-const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env
+const BOT_TOKEN = process.env.BOT_TOKEN ?? '<BOT TOKEN HERE>'
+const PROJECT_ID = process.env.PROJECT_ID ?? '<Google Project ID>'
+const FUNCTION_NAME = process.env.FUNCTION_NAME ?? 'botFunction'
+const REGION = process.env.REGION ?? 'us-central1'
 
-const bot = new Telegraf(BOT_TOKEN!)
+const bot = new Telegraf(BOT_TOKEN)
 
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 bot.telegram.setWebhook(
-  `https://${REGION!}-${PROJECT_ID!}.cloudfunctions.net/${FUNCTION_NAME!}`
+  `https://${REGION}-${PROJECT_ID}.cloudfunctions.net/${FUNCTION_NAME}`
 )
 
 bot.command('hello', (ctx) => ctx.reply('Hello, friend!'))
--- a/docs/examples/google-cloud-function.ts
+++ b/docs/examples/google-cloud-function.ts
@@ -19,10 +19,9 @@ export const botFunction = async (req: any, res: any) => {
   try {
     await bot.handleUpdate(req.body)
     console.log('Success')
-    res.status(200).send('Success')
   } catch (err) {
-    console.log('Failure')
     console.log(err)
-    res.status(500).send('Something went wrong')
+  } finally {
+    res.status(200).send('Success')
   }
 }
--- a/docs/examples/google-cloud-function.ts
+++ b/docs/examples/google-cloud-function.ts
@@ -5,7 +5,7 @@ const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env
 // eslint-disable-next-line
 const bot = new Telegraf(`${BOT_TOKEN}`)
 
-// eslint-disable-next-line
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
 bot.telegram.setWebhook(
   // eslint-disable-next-line
   `https://${REGION}-${PROJECT_ID}.cloudfunctions.net/${FUNCTION_NAME}`
--- a/docs/examples/google-cloud-function.ts
+++ b/docs/examples/google-cloud-function.ts
@@ -1,27 +1,24 @@
 import { Telegraf } from 'telegraf'
 
-const BOT_TOKEN = process.env.BOT_TOKEN ?? '<BOT TOKEN HERE>'
-const PROJECT_ID = process.env.PROJECT_ID ?? '<Google Project ID>'
-const FUNCTION_NAME = process.env.FUNCTION_NAME ?? 'botFunction'
-const REGION = process.env.REGION ?? 'us-central1'
+const { BOT_TOKEN, FUNCTION_NAME, PROJECT_ID, REGION } = process.env
+
+if (BOT_TOKEN === undefined) {
+  throw new TypeError('BOT_TOKEN must be provided!')
+}
 
 const bot = new Telegraf(BOT_TOKEN)
 
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 bot.telegram.setWebhook(
-  `https://${REGION}-${PROJECT_ID}.cloudfunctions.net/${FUNCTION_NAME}`
+  `https://${REGION!}-${PROJECT_ID!}.cloudfunctions.net/${FUNCTION_NAME!}`
 )
 
 bot.command('hello', (ctx) => ctx.reply('Hello, friend!'))
 
 export const botFunction = async (req: any, res: any) => {
-  console.log(req.body)
   try {
     await bot.handleUpdate(req.body)
-    console.log('Success')
-  } catch (err) {
-    console.log(err)
   } finally {
-    res.status(200).send('Success')
+    res.status(200).end()
   }
 }
--- a/docs/examples/google-cloud-function.ts
+++ b/docs/examples/google-cloud-function.ts
@@ -1,3 +1,5 @@
+// @ts-expect-error `npm install --save-dev @types/express`
+import { Request, Response } from 'express'
 import { Telegraf } from 'telegraf'
 
 const { BOT_TOKEN, FUNCTION_NAME, PROJECT_ID, REGION } = process.env
@@ -15,7 +17,7 @@ bot.telegram.setWebhook(
 
 bot.command('hello', (ctx) => ctx.reply('Hello, friend!'))
 
-export const botFunction = async (req: any, res: any) => {
+export const botFunction = async (req: Request, res: Response) => {
   try {
     await bot.handleUpdate(req.body)
   } finally {
--- a/docs/examples/google-cloud-function.ts
+++ b/docs/examples/google-cloud-function.ts
@@ -1,0 +1,26 @@
+import { Telegraf } from 'telegraf';
+
+const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env;
+
+const bot = new Telegraf(BOT_TOKEN, {
+  telegram: { webhookReply: true },
+});
+
+bot.telegram.setWebhook(
+  `https://${REGION}-${PROJECT_ID}.cloudfunctions.net/${FUNCTION_NAME}`
+);
+
+bot.command('hello', async (ctx) => ctx.reply("Hello, friend!"));
+
+export const botFunction = async (req, res) => {
+  console.log(req.body);
+  try {
+    await bot.handleUpdate(req.body);
+    console.log('Success');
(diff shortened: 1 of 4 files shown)


docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,29 @@
+import { Telegraf } from 'telegraf'
+
+const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env
+
+// eslint-disable-next-line
+const bot = new Telegraf(`${BOT_TOKEN}`, {
+  telegram: { webhookReply: true },
```
wojpawlik: This line does nothing here, please remove.

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,29 @@
...
+})
+
+// eslint-disable-next-line
```
wojpawlik: Why is this needed?
```suggestion
```

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,29 @@
...
+export const botFunction = async (req: any, res: any) => {
+  console.log(req.body)
+  try {
+    await bot.handleUpdate(req.body)
+    console.log('Success')
+    res.status(200).send('Success')
+  } catch (err) {
+    console.log('Failure')
+    console.log(err)
+    res.status(500).send('Something went wrong')
```
wojpawlik: This means that Telegram will resend the update and probably cause the same error again, right? Why is that desirable?

jacobwaller:
@wojpawlik Not sure why you marked this PR as a draft, it is finished

wojpawlik (changes requested):

docs/examples/google-cloud-function.ts
jacobwaller: setWebhook returns a promise and the eslint rules are set to not allow unhandled promises

docs/examples/google-cloud-function.ts
jacobwaller: No, the catch is to handle calls to the cloud function that are not from telegram

docs/examples/google-cloud-function.ts
EdJoPaTo: you can explictly mark which rule this is disabling:

```suggestion
// eslint-disable-next-line @typescript-eslint/no-floating-promises
```

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,27 @@
+import { Telegraf } from 'telegraf'
+
+const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env
+
+// eslint-disable-next-line
+const bot = new Telegraf(`${BOT_TOKEN}`)
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+bot.telegram.setWebhook(
+  // eslint-disable-next-line
```
EdJoPaTo: Please specify exactly which rule is disabled

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,27 @@
+import { Telegraf } from 'telegraf'
+
+const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env
+
+// eslint-disable-next-line
```
EdJoPaTo: Please specify exactly which rule is disabled

EdJoPaTo (changes requested):
eslint can be a lot smarter when it knows which rules are disabled and will warn you when the rule is not disabled (anymore) because the rule got better for example.

Just disabling something is guesswork afterwards. Also I would prefer examples not needing bodges as that shouldn't be exemplary.

docs/examples/google-cloud-function.ts
```diff
@@ -2,12 +2,12 @@ import { Telegraf } from 'telegraf'
 
 const { BOT_TOKEN, PROJECT_ID, FUNCTION_NAME, REGION } = process.env
```
EdJoPaTo: You should check if these variables are set rather than disabling eslint warnings.
```ts
if (!BOT_TOKEN) {
  throw new Error('BOT_TOKEN required');
}
```

Having this check the first warning will be gone. Checking the other 3 and the second warning will be gone.

Typescript is there to help us, not to disable its warnings :)

jacobwaller: The solution in that case would end up being overly specific. If you don't provide a bot token, telegram already throws an error. And `PROJECT_ID`, `FUNCTION_NAME`, and `REGION` are provided as env variables by default in Google Cloud Functions

jacobwaller: Yes, I could check every env variable for undefined, but at that point, this example starts to become a little too complicated to be used as an example no?

EdJoPaTo: If you are sure they are there maybe just use the `!` then.

```ts
const bot = new Telegraf(BOT_TOKEN!)
```

Seems more descriptive to me than ignoring eslint warnings.

docs/examples/google-cloud-function.ts
wojpawlik: How do you expect this to work? I'd expect you to check request path instead.

Please don't mark a conversation as resolved that I unresolved.
Also, I'll probably be inactive here for a while, focusing on 4.1.0 release. This will ship in 4.1.1.

jacobwaller: I can just remove the error checking if the way I'm handling it isn't okay. Most of the other examples here don't have any

wojpawlik: I'm unfamiliar with the platorm, but I'd do something akin to

```ts
try {
  await bot.handleUpdate(req.body)
finally {
  res.status(200)
}
```

Then if an update causes the bot to error, it won't be redelivered and won't cause another error.

jacobwaller: Updated

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,27 @@
+import { Telegraf } from 'telegraf'
+
+const BOT_TOKEN = process.env.BOT_TOKEN ?? '<BOT TOKEN HERE>'
+const PROJECT_ID = process.env.PROJECT_ID ?? '<Google Project ID>'
+const FUNCTION_NAME = process.env.FUNCTION_NAME ?? 'botFunction'
+const REGION = process.env.REGION ?? 'us-central1'
```
wojpawlik: If these are always defined by Google, please just use non-null asserion

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,27 @@
...
+
+bot.command('hello', (ctx) => ctx.reply('Hello, friend!'))
+
+export const botFunction = async (req: any, res: any) => {
+  console.log(req.body)
+  try {
+    await bot.handleUpdate(req.body)
+    console.log('Success')
+  } catch (err) {
+    console.log(err)
```
wojpawlik: `telegraf` alsready has (optional) logging
```suggestion
```

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,27 @@
+import { Telegraf } from 'telegraf'
+
+const BOT_TOKEN = process.env.BOT_TOKEN ?? '<BOT TOKEN HERE>'
```
wojpawlik: Please use empty string as default value or throw if it's `undefined`.

wojpawlik (changes requested):
Please add this file to the `<ul>` in https://github.com/telegraf/telegraf/blob/develop/README.md#webhooks.

jacobwaller:
For a project with very little documentation and examples, it sure is difficult to submit an example...

wojpawlik:
Ok, I'll fix it myself

wojpawlik:
@jacobwaller please verify I haven't messed anything up :)

wojpawlik (approved):

jacobwaller:
> @jacobwaller please verify I haven't messed anything up :)

LGTM - Thanks!

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,24 @@
+import { Telegraf } from 'telegraf'
+
+const { BOT_TOKEN, FUNCTION_NAME, PROJECT_ID, REGION } = process.env
+
+if (BOT_TOKEN === undefined) {
```
EdJoPaTo, line 7: Maybe even check for falsy = empty strings?
```suggestion
if (!BOT_TOKEN) {
```

docs/examples/google-cloud-function.ts
```diff
@@ -0,0 +1,24 @@
...
+const bot = new Telegraf(BOT_TOKEN)
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+bot.telegram.setWebhook(
+  `https://${REGION!}-${PROJECT_ID!}.cloudfunctions.net/${FUNCTION_NAME!}`
+)
+
+bot.command('hello', (ctx) => ctx.reply('Hello, friend!'))
+
+export const botFunction = async (req: any, res: any) => {
```
EdJoPaTo: I'm not sure with the platform. I assume this exported function is then used by the google cloud thingy?

Two questions here:

- Can we provide typings for the arguments? If this example is copied over users might find these typings helpful. But they are probably not as simple and would require an additional import?

- I would expect functions to be declared like functions and not like variables? Or is there any reason for this I am missing here? Or just personal preference?
```suggestion
export async function botFunction(req: any, res: any) {
```

EdJoPaTo (approved):

docs/examples/google-cloud-function.ts
wojpawlik: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md

EdJoPaTo: Might be helpful as we dont have to know about this falsy stuff then.

wojpawlik: Telegraf does the "sloppy" check internally anyway:

https://github.com/telegraf/telegraf/blob/66a0020e8522c81d600f2c309800aad058cacbd5/src/core/network/client.ts#L336-L341

It's a leftover from JS days, possibly still useful to people writing JS.

docs/examples/google-cloud-function.ts
wojpawlik: Here function syntax is purely a preference, not a problem, so I left it as-is.

More accurate parameter types would require a `devDependency` on `@types/express`.

EdJoPaTo: Maybe just add a comment for this requirement then? Seems better than `any` to me. (I really don't like seeing `any`, especially in code meant to be copied.)
```ts
// You have to `npm install --save-dev @types/express` to use this example
import {Request, Response} from 'express'
```

But I assume the CI will be annoying about unknown types then…?

EdJoPaTo (approved):

Issue #1306, opened 2021-01-17
Not enough types for Context


NedloGlitch:
**Is your feature request related to a problem? Please describe.**
I like the new typings system, but I think it lacks types for dividing code - when I want to create a functions/methods as handlers, I can't just grab "GuardedContext" and use it.

Code Example
```ts
import { Telegraf } from 'telegraf';
import { Message } from 'telegraf/typings/telegram-types';
import { Context, TextContext, GuardedContext } from 'telegraf/typings/context';

const bot = new Telegraf(process.env.BOT_TOKEN as string);

let number = 0;

bot.on('text', async (ctx) => {
    if (number === 0) {
        await first_handler(ctx);
    } else if (number === 1) {
        await second_handler(ctx);
    } else if (number === 2) {
        await third_handler(ctx);
    }
});

async function first_handler(ctx: Context & { message: Message.TextMessage }) {
    const message = (ctx.message.text) // Message string   
    if (message === "hello"){
        ctx.reply("Hey!") 
    } else {
        ctx.reply("Please say hello to me!")
    }
    number += 1;
}

async function second_handler(ctx: GuardedContext<Context, { message: Message.TextMessage; }>) { // This is what could be used, but its not available for import
    const message = (ctx.message.text)  
    if (message === "hello") {
        ctx.reply("Hey again!");
    } else {
        ctx.reply("Please say hello to me!");
    }
    number += 1;
}

async function third_handler(ctx: TextContext) { // This is how it can look like if we add premade types
    const message = (ctx.message.text)
    if (message === "hello") {
        ctx.reply("I've already met you!");
    } else {
        ctx.reply("Please say hello to me!");
    }
    number = 0;
}
```

**Describe the solution you'd like**
Maybe it's better to create separate types (TextContext, AudioContext, VideoContext, etc.)

**Describe alternatives you've considered**
Or make GuardedContext available for import

**Additional context**
I'm new, so don't be mad at me.

wojpawlik:
I'm glad you asked! We probably should expose `GuardedContext`, just, how should it be described (JSDoc) to avoid confusing newcommers? And is `GuardedContext` even a good name?

Related: https://github.com/telegraf/telegraf/discussions/1284#discussioncomment-272819

> Maybe it's better to create separate types (TextContext, AudioContext, VideoContext, etc.)

No. Less powerful and needs way more code.

NedloGlitch:
[...]

PR #1368, opened 2021-02-15
Narrow more getters in `GuardedContext` to future-proof it

wojpawlik:
# Description

Before exposing `GuardedContext` (#1306), I decided to future-proof it. What better way to do that than to add a difficult feature and fix what breaks? So I narrowed `ctx.from`, `ctx.chat` and `ctx.senderChat`, and now I need someone to review this.

# Checklist:


KnorpelSenf (approved):
LGTM.

Maybe find a better name for the commit message

PR #1378, opened 2021-02-22
Fix type of `ctx.from`

wojpawlik:
@EdJoPaTo, what should I rename `GetMessageFromAnySource` to? Also, it'd be nice if you checked it actually works: `npm install --no-save telegraf/telegraf#8ad3d7d`

# Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4170,9 +4170,9 @@
       "dev": true
     },
     "typegram": {
-      "version": "3.1.4",
-      "resolved": "https://registry.npmjs.org/typegram/-/typegram-3.1.4.tgz",
-      "integrity": "sha512-AsYxLLynwBgeUJNU+3oSYNPSB3bwb/uyzrEPvrPwC0HXCYhW+otNV5q4T8Sb43SQSAP71hWC8EzBgDtlGGBx0A=="
+      "version": "3.1.10",
+      "resolved": "https://registry.npmjs.org/typegram/-/typegram-3.1.10.tgz",
+      "integrity": "sha512-VfsuVP4uqk9yllRpaGp9yFpzvLKXghFCOSc3TVi+fy1IL2Bi4z2gJ9rMgVSSug4pWQP8Ui7Isvl14yYV8VEWaQ=="
     },
     "typescript": {
       "version": "4.1.3",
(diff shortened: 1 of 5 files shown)


EdJoPaTo:
your fix does [seem to work](https://github.com/EdJoPaTo/wikidata-telegram-bot/runs/1952517232?check_suite_focus=true). `xo` fails for whatever reason but as that's the second command the first ran successfully… which means… `tsc` compiled without errors. Locally it also works.

About the name… Maybe just `MessageFromAnySource` without the get? Other than that the type exactly describes what it does.

wojpawlik:
The problem is, it can now return non-`Message`:
obraz

EdJoPaTo:
Maybe `UpdateContent`?

wojpawlik:
I need help fixing tests @KnorpelSenf

wojpawlik:
I had to rename `UpdateTypes` array to avoid name conflicts, so I moved it to `telegram-types.ts` while at it to avoid circular import. Because `telegram-types.ts` now contains runtime code, TypeScript must emit the import. These lines typecheck, but fail at runtime:

https://github.com/telegraf/telegraf/blob/2fcffb2cee939dafcd11ff65772a17ff562a932f/src/telegram-types.ts#L7-L13

There are 3 possible solutions:

1. Don't reexport the types here, import those directly from `typegram`,
2. Use named re-exports (`export { Foo }`),
3. Change `typegram` so that star re-exports work:
   ```ts
   "exports": {
     ".": "./index.js",
     "./callback": "./index.js",
     "./inline": "./index.js",
     "./manage": "./index.js",
     "./message": "./index.js",
     "./passport": "./index.js",
     "./payment": "./index.js",
     "./update": "./index.js"
   },
   ```

Related to #1380

EdJoPaTo:
Personally I try to avoid reexports as this introduces additional complexity. (I only use index.ts with reexports sometimes)

wojpawlik:
Option 1 is too much work for this PR, might reconsider for #1380.

So if @KnorpelSenf wants `typegram` to support `export * from "typegram/*"`, I'll just upgrade, otherwise option 2.

KnorpelSenf:
Does the remapping of all files to `index.js` actually have the desired effect? I'm not too familiar with how Node handles this, but I'd expect that doing `import * from 'typegram/message'` would then be the same as `import * from 'typegram'`, which does not solve the problem.

Otherwise, I'd accept a PR about this

PR #1380, opened 2021-02-22
Expose `src/telegram-types.ts` as `Types`

### LINKED ISSUE #1379: Expose telegram-types module

In the following example, I would like to create a method named `quote`, which will force the bot to quote the message, even in an one-to-one private chat.

```typescript
// Extends the Context
interface CustomContext extends Context {
	quote: (
		text: string,
		// **While the ExtraReplyMessage is now not exported**
		extra?: Omit<ExtraReplyMessage, 'reply_to_message_id'>
	) => ReturnType<Telegram['sendMessage']>,
}

// Then use middleware to implement it, it would only type-pass in Typescript
// with the extended CustomContext
ctx.quote = (text, extra) => ctx.reply(
	text,
	{ reply_to_message_id: ctx.message?.message_id, ...extra }
)

// So now we could use `quote` in the callback function
bot.on('document', ctx => ctx.quote('Gotcha'))
```

I would like to suggest that types defined in [telegram-types.ts](https://github.com/telegraf/telegraf/blob/develop/src/telegram-types.ts) (except types from `typegram`) should be publically exported. It should help a lot for stuff like defining aliases for existing methods in Typescript.

Just copy and paste the declarations, awkward.

**Additional context**

Maybe there are some workarounds while exposing it is the most intuitive way I could now think of.

equt:
# Description

- #1379 
- ~~Just one-line export~~ Now it's not

## Type of change

# How Has This Been Tested?

# Checklist:


wojpawlik:
Makes sense. `git merge develop` should fix CI.

I have only one concern: `src/telegram-types.ts` currently exports a big chunk of `typegram`, possibly preventing upgrading to Typegram 4, if it ever happens cc @KnorpelSenf https://github.com/telegraf/telegraf/blob/0894986e8f45d51a8e90c829be8a8184c29e6bf5/src/telegram-types.ts#L6-L13

On the other hand, perhaps docs can be improved by exporting `typegram` types and removing https://github.com/telegraf/telegraf/blob/0894986e8f45d51a8e90c829be8a8184c29e6bf5/tsconfig.json#L17-L18

In the meantime, use (internal) `telegraf/typings/telegram-types`.

equt:
Updated.

> `src/telegram-types.ts` currently exports a big chunk of `typegram`

Yes, and I believe all types coming from `typegram` should not be exposed. AFAIN TypeScript doesn't support hiding some exports. Maybe we should remove these wildcard exports and import them from `typegram` directly when needed. Any better ways to approach this is welcomed.

equt:
A bonus is that after exposing the types, methods signatures are now much more readable.

image

would become

image

Maybe left `tsconfig.json` as is should be just fine.

wojpawlik (approved):
`tt` is nicer to type than `tg`...

KnorpelSenf (changes requested):
The exports [here](https://github.com/telegraf/telegraf/pull/1380/files#diff-4ca0d022219464363b9b86506fcf0ad63fb51a68339ec7adb100a97188d8efc7R26-R36) should be seen as part of the typegram exports, confer [this section](https://github.com/KnorpelSenf/typegram#customizing-inputfile) of the typegram readme.

I'm not sure how to refactor this—it may make sense to create a new file that wraps typegram re-exports everything and does the proxy type stuff? This would also reduce the places where telegraf refers to typegram directly. But this is the naïve solution, I'm open for other approaches if you have a better idea :)

wojpawlik:
Quickfix: https://www.typescriptlang.org/tsconfig#stripInternal. I don't know how to meaningfully split the file into two.

KnorpelSenf:
> I don't know how to meaningfully split the file into two.

Well one way is what I suggested: remove the typegram-specific code from `src/telegram-types.ts`, and keep only the telegraf-specific code in there.

But it's also just my opinion that telegraf should either
1. export all of typegram, or
2. export none of it

The current state of the PR re-exports typegram partially. I think that should be changed, but it's not exactly _wrong_ to do it (just hard to understand for others), so you may as well keep it the way it is :)

wojpawlik:
https://github.com/telegraf/telegraf/blob/b45cfe45417d945b132c72e56dc3b1fcc00539af/src/telegram-types.ts#L40 could be removed / inlined.

I wouldn't mind moving [these](https://github.com/telegraf/telegraf/blob/b45cfe45417d945b132c72e56dc3b1fcc00539af/src/telegram-types.ts#L6-L36) to `src/internal/types/typegram.ts` and perhaps re-adding the `typegram` re-exports there.

wojpawlik:
What do you think @KnorpelSenf?

KnorpelSenf:
I think that sounds good (based on your description).

Just to be sure that we understand each other: there would be two main files:
- telegram types that telegraf introduces, all of them are exported
- typegram types that telegraf uses internally (re-exports typegram, used internally by telegraf, but does not expose the types to library consumers)

Are we on the same page?

wojpawlik:
Is there anything else that should be re-exported from `src/telegram-types.ts` @KnorpelSenf @equt?

wojpawlik:
This is blocking one PR I wish to make.

src/button.ts
```diff
@@ -1,4 +1,4 @@
-import { InlineKeyboardButton, KeyboardButton } from './telegram-types'
+import { InlineKeyboardButton, KeyboardButton } from 'typegram'
```
KnorpelSenf: I think this should not be a direct dependency to typegram, but use the new module instead

package.json
```diff
@@ -76,8 +76,8 @@
     "eslint-plugin-standard": "^4.1.0",
     "mjs-entry": "gist:de6257751f54b3c66319bae8d2a8aea0",
     "prettier": "2.2.1",
-    "typedoc": "^0.20.13",
-    "typescript": "^4.1.3"
+    "typedoc": "^0.20.29",
+    "typescript": "~4.2.3"
```
KnorpelSenf, line 80: Any rationale behind `~` not `^`?

KnorpelSenf (approved):
LGTM, but I have one question:

package.json
wojpawlik: To avoid breaking the build for users of alternative package managers (minor TypeScript releases contain breaking changes)

PR #1386, opened 2021-02-27
Make default `hookPath` deterministic

wojpawlik:
Should prevent 403s in cluster or when woken up on Heroku:

obraz

cc @ioscars

https://github.com/telegraf/telegraf/blob/develop/README.md#webhooks needs rewriting after this.

# Checklist:


KnorpelSenf (approved):
Being deterministic here makes sense. I can give a more thorough review in a few days, if you don't want to merge yet. I'll mainly investigate whether
- there's a more convenient way to use it
- this change has security implications or not

electromama (approved):

KnorpelSenf (approved):
LGTM

PR #1401, opened 2021-03-10
Allow narrowing via `Message::hasOwnProperty`

wojpawlik:
# Description

> `ctx.message?.hasOwnProperty("text")` vs `ctx.message !== undefined && "text" in ctx.message`

# How Has This Been Tested?

Manually.

# Checklist:

### DIFF
--- a/src/telegram-types.ts
+++ b/src/telegram-types.ts
@@ -172,3 +172,14 @@ export type MessageSubType =
       UnionKeys<Message>,
       keyof Message.CaptionableMessage | 'entities' | 'media_group_id'
     >
+
+declare module 'typegram/message' {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Message {
+    interface ServiceMessage {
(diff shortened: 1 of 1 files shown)


KnorpelSenf:
This is really hacky and I still don't see why anyone would want to do this, let alone recommend that approach. On the other hand, it should not hurt anyone either. Does it have any downsides to mess with the types of deps?

wojpawlik:
Well, it's not emitted to `.d.ts`, so it has no effect on Telegraf consumers. Would you be open to a PR to upstream it to Typegram?

It's only meant to improve on `in`, which cannot narrow deunionized or possibly `undefined` values.

KnorpelSenf:
If I understand it correctly, it only replaces `'text' in msg` by `msg.hasOwnProperty('text')`. Where is the benefit of this?

wojpawlik:
Future-proofing, and optional chaining:

> `ctx.message?.hasOwnProperty("text")` vs `ctx.message !== undefined && "text" in ctx.message`

PR #1436, opened 2021-04-17
Rewrite `session`

### LINKED ISSUE #1350: Stage should fail if there is no session exists instead of passing through
When there is no session attached to the bot (for example, you forgot to add `bot.use(session())`), stage middleware passes through 

https://github.com/telegraf/telegraf/blob/278046726faaf006f544b4c8e07fcd118d47dee6/src/scenes/stage.ts#L51

Maybe it should throw an error like `Session is required to use Stage`

Like, is there a use case when this is intended?

Preferably, it should throw when starting the bot, not when an update comes. But I don't think that is possible

### LINKED ISSUE #1372: `session` is prone to race conditions
* Telegraf.js Version: 4.0.3

```js
'use strict'

const { Telegraf, session } = require('telegraf')
const createDebug = require('debug')

const debug = createDebug('test')

bot.use(session())
bot.on('message', async ctx => {
  if (ctx.session === undefined) {
    const { update_id } = ctx.update
    ctx.session = { update_id }
    debug('Defaulting session to', ctx.session)
  } else {
    debug('Session already set to', ctx.session)
  }
})

createDebug.enable('telegraf:client test')
bot.launch()
```

Send a few messages, then start the bot. After it processes the messages, send one more.

## Expected Behavior

Each `session` change is atomic:

```
  telegraf:client HTTP call deleteWebhook { drop_pending_updates: undefined } +220ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 0, allowed_updates: [] } +164ms
  test Defaulting session to { update_id: 224422703 } +0ms
  test Session already set to { update_id: 224422703 } +1ms
  test Session already set to { update_id: 224422703 } +1ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 224422706, allowed_updates: [] } +368ms
  test Session already set to { update_id: 224422703 } +3s
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 224422707, allowed_updates: [] } +3s
```

## Current Behavior

`ctx.session` is read before processing each update, and written only after processing has finished. Concurrent tasks might not notice each-others' modifications, last task to finish overwrites data:

```
  telegraf:client HTTP call deleteWebhook { drop_pending_updates: undefined } +220ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 0, allowed_updates: [] } +164ms
  test Defaulting session to { update_id: 224422703 } +0ms
  test Defaulting session to { update_id: 224422704 } +1ms
  test Defaulting session to { update_id: 224422705 } +1ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 224422706, allowed_updates: [] } +368ms
  test Session already set to { update_id: 224422705 } +3s
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 224422707, allowed_updates: [] } +3s
```

I believe every https://www.npmjs.com/search?q=telegraf-session- and https://github.com/nitreojs/puregram/tree/master/packages/session are affected as well.
cc @EdJoPaTo @KnightNiwrem @KnorpelSenf @mdsina @nitreojs @TemaSM

## Solution?

Drop async `SessionStore` support (related: #1342), use `Object.defineProperty` so that each `ctx.session` access performs `.get`, `.set`, or `.delete`.

In the meantime, I'll advise `session` users to always use https://github.com/KnightNiwrem/telegraf-throttler, with default settings.

wojpawlik:
# Description

Fixes #1372, fixes #1350.

# Checklist:


src/session.ts
```diff
@@ -1,15 +1,14 @@
...
-import { MaybePromise } from './composer'
 import { MiddlewareFn } from './middleware'
 
 export interface SessionStore<T> {
-  get: (name: string) => MaybePromise<T | undefined>
-  set: (name: string, value: T) => MaybePromise<void>
-  delete: (name: string) => MaybePromise<void>
+  readonly get: (name: string) => T | undefined
+  readonly set: (name: string, value: T) => void
+  readonly delete: (name: string) => void
```
KnorpelSenf, line 7: What kind of session do you expect not to have async operations? Doesn't that completely defeat the point of persistence?

src/session.ts
```diff
@@ -30,66 +29,37 @@ export interface SessionContext<S extends object> extends Context {
...
     if (key == null) {
       return await next()
     }
-    ctx.session = await store.get(key)
+    Object.defineProperty(ctx, 'session', {
+      get() {
+        return store.get(key)
+      },
+      set(value) {
+        if (value === undefined) {
```
KnorpelSenf: Maybe check for nullish?

src/session.ts
wojpawlik: Mutable object is a bad database interface. So, storage must store all data in memory, but it can still persist it.

src/session.ts
KnorpelSenf: I.e., handle the case where users assign `null`. We may want to handle this the same way as for `undefined`.

src/session.ts
KnorpelSenf: But how to persist something synchronously? I assume you aren't suggesting to just dump some floating promises?

src/session.ts
KnorpelSenf:
```suggestion
        if (value == null) {
```

src/session.ts
wojpawlik: That, or sync IO, or `setInterval`...

src/session.ts
wojpawlik: I understand, but I don't see any benefit of (effectively) replacing `null` with `undefined`.

KnorpelSenf: The point is that telegraf does this in all places, it checks for nullish instead of strictly undefined. So why suddenly do it differently here?

src/session.ts
KnorpelSenf: Do you believe that this is a good API?

wojpawlik: No. Good API needs transactions. That's the best I can do without them.

src/session.ts
wojpawlik: We don't iterate over keys, so from JS PoV, the `.delete(key)` is just memory-efficient `.set(key, undefined)`, not `.set(key, null)`.

src/session.ts
KnorpelSenf: I vote for either introducing that transaction support, or dropping session altogether, rather than adding a module which we agree is bad.

wojpawlik: API would be completely different. If I do it, it'll be a separate package, independent from Telegraf.

src/session.ts
wojpawlik: If someone ignores the types and sets `ctx.session` to `null`, why should `ctx.session` be `undefined` and not `null`?

src/session.ts
KnorpelSenf: I.e. drop sessions?

src/session.ts
KnorpelSenf: Just because that's always what people were supposed to do. If was even in the example code for sessions earlier: if you want to delete the session, set it to `null`. But given this is a rewrite, you may as well introduce a breaking change here, too …

wojpawlik: Could you link the previous example?

src/session.ts
wojpawlik: That's more drastic than restricting it to be safe, but I could still do that.

KnorpelSenf: That is true, but it would also be the more consequent step. I leave it up to you

src/session.ts
KnorpelSenf: https://github.com/telegraf/telegraf/tree/b1c0e3c238a33ce6a367e8efd5ce2349c5eeba5f#destroying-a-session

KnorpelSenf: But I think I was thinking about this: https://github.com/telegraf/telegraf-session-firebase#destroying-a-session

src/session.ts
wojpawlik: Nothing stops users from using async 3rd party session.

KnorpelSenf: Correct, then feel free to go ahead with removing it from core

MKRhere: We won't remove it from core - I need more time to think about this. After working on the other stuff pending.

Loskir: > Nothing stops users from using async 3rd party session.

But since it's an interface and plugins will probably use it, won't these 3rd party sessions be incompatible with plugins?

PR #1463, opened 2021-06-01
Add a white background to make the header visible in dark mode UIs

alexslade:
## Problem
Users who use dark-mode (like myself) can't see the existing header image, it's dark text on a dark background. 

Screenshot 2021-06-01 at 16 32 30

## This PR
Changes the header background to #ffffff, matching the github light theme background. It is visually unchanged in light mode, and visible in dark mode. 

I haven't made any changes to the spacing, which you might want to adjust if to be tidier across backgrounds. 

Screenshot 2021-06-01 at 16 39 35


wojpawlik:
🔥👀

I'd rather remake the logo as SVG.

alexslade:
Yea, that would be good. This was a quick drive-by fix you can use in the meantime if you want. Happy for you to close this too.

wojpawlik:
What do you think?

telegraf

PR #1464, opened 2021-06-02
Delete useless check

sjinks:
# Description

This PR removes a useless check in the `attachFormValue()` function.

```typescript
async function attachFormValue(
// ...
) {
  if (value == null) {
    return
  }

// ...

  if (
    value &&
    typeof value === 'object' &&
    hasProp(value, 'media') &&
    hasProp(value, 'type') &&
    typeof value.media !== 'undefined' &&
    typeof value.type !== 'undefined'
  ) {
```

The `if (value)` check was meant to ensure that value is not `null` (because `typeof null === 'object'`) but that condition is already checked.

## Type of change

# How Has This Been Tested?

Tested with LGTM static code analysis (it reported this issue).

**Test Configuration**:
* Node.js Version: 14.17.0
* Operating System: Ubuntu 21.04

# Checklist:

- [ ] I have made corresponding changes to the documentation - N/A
- [ ] I have added tests that prove my fix is effective or that my feature works - N/A

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -188,7 +188,6 @@ async function attachFormValue(
     })
   }
   if (
-    value &&
     typeof value === 'object' &&
     hasProp(value, 'media') &&
     hasProp(value, 'type') &&


sjinks:
Looks like TypeScript 4.2 is unable to handle this :-( TS 4.3 does, but it finds many other issues.

Issue #1388, opened 2021-03-05
Missing type definitions on ctx.message


markkarrica:
The `message` property of `ctx` has a lot of missing type definitions which don't follow the [official ones](https://core.telegram.org/bots/api#message). 
![Missing properties](https://user-images.githubusercontent.com/60916242/110121707-03bee600-7dbf-11eb-985e-6880f9033254.png)
![Missing text property](https://user-images.githubusercontent.com/60916242/110121872-38cb3880-7dbf-11eb-83aa-351cae4e0746.png)

The line of code causing the error was this one:
`bot.on("message", (ctx) => ctx.reply(ctx.message.text));`
The bot should reply with the text you send to him.

wojpawlik:
I'm glad you asked! I explained it a few times already, but I think about this a lot, so I have a better understanding each time.

TypeScript is trying to tell you that not every message is a text message.

There are several ways you can proceed in Telegraf 4:

1. Use `bot.on('text', (ctx) => ...)`. It's that simple.
2. Narrow manually: `if (!('text' in ctx.message)) return next()` (messy when `ctx.message` might be `undefined`).
3. Manually use a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) (not included): `if (!has.text(ctx.message)) return next()`.
4. Pass a type guard to [`Composer.guard`](https://telegraf.js.org/classes/composer.html#guard-1).
5. `deunionize(ctx.message).text`.

What could Telegraf 5 do?

6. Simply make each `Message` property always optional. Telegraf 3 did this, and it needs a ton of (unsafe) non-null assertions:
	```ts
	bot.on('text', ctx => ctx.reply(ctx.message!.text!))

	// typechecks, fails at runtime
	bot.on('photo', ctx => ctx.reply(ctx.message!.text!))
	```
7. Add `_type` discriminator property. One needs to know to check it. Needs extra code that breaks with API updates.
8. `deunionize` by default. After narrowing manually, suggestions are bloated with multiple `?: undefined` properties.

equt:
You could also use ascription if you're totally sure about the message type, and this brings you zero cost at the runtime.

```typescript
import { Message } from 'typegram'

// ...

bot.on('text', (ctx) => ctx.reply((ctx.message as Message.TextMessage).text))

// ...

```
---

I believe this is where TypeScript takes over JavaScript. An explicit type guard is necessary to force you to consider those corner cases.

[...]

PR #1523, opened 2021-09-11
Export typings from composer

0wx:
# Description

I often create separated file for handling command, text, etc.
`index.ts`:
```ts
import { start } from './handling/start.ts'
bot.start(start)
```

`start.ts`:
```ts
import { StartContext } from "../typings/bot"

export const start = async (ctx: StartContext): Promise<void> => {
  ctx.reply(`Start message`)
}

```

But things getting really messy on my typings
`bot.d.ts`:
```ts
import { Context, NarrowedContext } from 'telegraf'
import { Update, Message } from 'typegram'
import * as tt from 'telegraf/src/telegram-types'

export type StartContext = Context<{
  message: Update.New & Update.NonChannel & Message.TextMessage
  update_id: number
}> &
  Omit<Context<Update>, keyof Context<Update>> & {
    startPayload: string
  }

export type MatchedContext<
  C extends Context,
  T extends tt.UpdateType | tt.MessageSubType
> = NarrowedContext<C, tt.MountMap[T]>


export type TextContext = MatchedContext<Context<Update>, 'text'>

```

It will be a lot better if typings in the `composer.ts` is exported. Thanks.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -5,16 +5,16 @@ import * as tt from './telegram-types'
 import { Middleware, MiddlewareFn, MiddlewareObj } from './middleware'
 import Context from './context'
 
-type MaybeArray<T> = T | T[]
+export type MaybeArray<T> = T | T[]
 export type MaybePromise<T> = T | Promise<T>
-type NonemptyReadonlyArray<T> = readonly [T, ...T[]]
-type Triggers<C> = MaybeArray<
+export type NonemptyReadonlyArray<T> = readonly [T, ...T[]]
+export type Triggers<C> = MaybeArray<
   string | RegExp | ((value: string, ctx: C) => RegExpExecArray | null)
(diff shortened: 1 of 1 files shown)


src/composer.ts
```diff
@@ -25,7 +25,7 @@ type MatchedMiddleware<
     Produces: a context that has some properties required, and some undefined.
     The required ones are those that are always present when the given update (or message) arrives.
     The undefined ones are those that are always absent when the given update (or message) arrives. */
-type MatchedContext<
+export type MatchedContext<
```
wojpawlik:
```suggestion
/** @deprecated */
type MatchedContext<
```

See https://github.com/telegraf/telegraf/issues/1388#issuecomment-808784654

wojpawlik (approved):

PR #1554, opened 2021-11-27
send  video with thumbnail please

mbahmunz:
please add feature to send video with thumbnail. thanks


KnorpelSenf:
This is already possible.

Also, consider opening an issue instead of a PR between two arbitrary branches next time you have a question :)

PR #1555, opened 2021-12-07
Bot API 5.4 and 5.5

MKRhere:
# Description

https://core.telegram.org/bots/api#november-5-2021
https://core.telegram.org/bots/api#december-7-2021

## Type of change

Update to Bot API 5.4 and 5.5.

# Checklist:


KnorpelSenf (commented):
All introduced changes look correct, but IIRC this isn't comprehensive. E.g. what about context shortcuts and message subtypes?

MKRhere:
@KnorpelSenf By context shortcuts do you not mean https://github.com/telegraf/telegraf/blob/8a5a335ad3bd14709af0c6974b0a85790a510fd8/src/context.ts#L860-L911

Could you explain message subtypes?

KnorpelSenf:
I mean the getters, not the methods. Message subtypes as in, not update types. Those two categories of strings need to be maintained when updates and messages are changed in the Bot API

src/telegram-types.ts
```diff
@@ -99,6 +99,7 @@ export type ExtraVenue = MakeExtra<
 export type ExtraVideo = MakeExtra<'sendVideo', 'video'>
 export type ExtraVideoNote = MakeExtra<'sendVideoNote', 'video_note'>
 export type ExtraVoice = MakeExtra<'sendVoice', 'voice'>
+export type ExtraBanChatSenderChat = MakeExtra<'banChatSenderChat'>
```
KnorpelSenf: I'm not sure why there's no `ExtraBanChatMember`. I also don't know if `sender_chat_id` should be passed here.

MKRhere: ### Bot API 5.3

* Renamed the method `kickChatMember` to `banChatMember`. The old method name can still be used.

That's why. It's `ExtraKickChatMember`.

KnorpelSenf:
There's a lot of implicit and non-obvious behaviour in Telegraf and you basically have to know a lot of things by heart with this library in order to judge whether or not this PR is complete. I knew those things before I moved on, but I am not confident that this is indeed a complete update. It looks like it will work at runtime, I'm mainly not sure about the types.

@wojpawlik may want to spend more time on this.

Either way, I'd just simply risk to introduce bugs and incomplete code. In case someone uses the new functionality before the lib migrates back to JS for v5 (people in the chat have been boldly announcing this recently IIRC), then they will notice the problems and report them, so they can be fixed in patch releases.

src/telegram-types.ts
MKRhere: I've deprecated `kickChatMember` and introduced `banChatMember`. Although this change is from Bot API 5.3, and I'd originally assumed it to be previously complete.

src/telegram-types.ts
```diff
@@ -81,6 +81,10 @@ export type ExtraKickChatMember = MakeExtra<
   'kickChatMember',
   'user_id' | 'until_date'
 >
+export type ExtraBanChatMember = MakeExtra<
+  'banChatMember',
+  'user_id' | 'until_date'
+>
```
wojpawlik:
```suggestion
export type ExtraBanChatMember = ExtraKickChatMember
```

wojpawlik (approved):

src/telegram-types.ts
MKRhere: I've gone the other way and replaced `kickChatMember`-related items with refs to `banChatMember`.

PR #1528, opened 2021-09-19
v3.40.0

murka:
https://github.com/telegraf/telegraf/discussions/1526#discussioncomment-1344548

PR #1610, opened 2022-04-25
add new_chat_member update type

alvintran:
# Description

Add new update types: my_chat_member (https://core.telegram.org/bots/api#update).
Add some functions to context.js file that make my_chat_member is work.

## Type of change

# How Has This Been Tested?

None

# Checklist:

### DIFF
--- a/context.js
+++ b/context.js
@@ -9,7 +9,8 @@ const UpdateTypes = [
   'pre_checkout_query',
   'message',
   'poll',
-  'poll_answer'
+  'poll_answer',
+  'my_chat_member'
 ]
 
 const MessageSubTypes = [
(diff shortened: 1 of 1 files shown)


context.js
```diff
@@ -380,11 +395,6 @@ class TelegrafContext {
     return this.telegram.unpinChatMessage(this.chat.id, ...args)
   }
 
-  unpinAllChatMessages () {
```
MKRhere, line 383: Why are `unpinAllChatMessages` and `copyMessage` removed?

MKRhere:
Actually, the `new_chat_member` addition seems to be made in #1528, pending an update from @murka. If you could work with them/pick up where they left off, might be possible to update the v3 branch. If you'd rather merge this and then we deal with #1528 later, that's probably fine as well.

PR #1614, opened 2022-04-28
feat: API 6.0

MKRhere:
* Upgrade to typegram@3.9.0; Bot API 6.0 support
* Update minimist to latest, resolves a [vulnerability](https://security.snyk.io/vuln/SNYK-JS-MINIMIST-2429795)
* Added `Markup.button.webApp` helper
* Experimental!: ⚠️ typegram & telegraf types exported as `telegraf/types` to be imported directly without relying on a separate dependency on `typegram`. The export interface is not stable. It may change at a later date

Try it out before it lands using `npm install telegraf@canary`

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -8,6 +8,12 @@ import Context from './context'
 export type MaybeArray<T> = T | T[]
 export type MaybePromise<T> = T | Promise<T>
 export type NonemptyReadonlyArray<T> = readonly [T, ...T[]]
+export type DefinitelyDefined<T> = T extends infer S | undefined ? S : T
+export type Expand<T> = T extends object
+  ? T extends infer O
+    ? { [K in keyof O]: O[K] }
+    : never
+  : T
 export type Triggers<C> = MaybeArray<
   string | RegExp | ((value: string, ctx: C) => RegExpExecArray | null)
 >
@@ -44,7 +50,11 @@ export type NarrowedContext<
 > = Context<U> & Omit<C, keyof Context>
 
 export interface GameQueryUpdate extends tg.Update.CallbackQueryUpdate {
-  callback_query: tg.CallbackQuery.GameShortGameCallbackQuery
+  callback_query: Expand<
+    Omit<tg.CallbackQuery, 'data'> & {
+      game_short_name: DefinitelyDefined<tg.CallbackQuery['game_short_name']>
+    }
+  >
 }
 
 function always<T>(x: T) {
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -8,7 +8,6 @@ import Context from './context'
 export type MaybeArray<T> = T | T[]
 export type MaybePromise<T> = T | Promise<T>
 export type NonemptyReadonlyArray<T> = readonly [T, ...T[]]
-export type DefinitelyDefined<T> = T extends infer S | undefined ? S : T
 export type Expand<T> = T extends object
   ? T extends infer O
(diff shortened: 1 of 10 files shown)


src/composer.ts
```diff
@@ -44,7 +50,11 @@ export type NarrowedContext<
 > = Context<U> & Omit<C, keyof Context>
 
 export interface GameQueryUpdate extends tg.Update.CallbackQueryUpdate {
-  callback_query: tg.CallbackQuery.GameShortGameCallbackQuery
+  callback_query: Expand<
+    Omit<tg.CallbackQuery, 'data'> & {
+      game_short_name: DefinitelyDefined<tg.CallbackQuery['game_short_name']>
+    }
```
KnorpelSenf, line 55: Is there a difference to `NonNullable`?

src/telegram.ts
```diff
@@ -1109,6 +1116,63 @@ export class Telegram extends ApiClient {
     })
   }
 
+  /**
+   * Use this method to change the bot's menu button in a private chat, or the default menu button. Returns true on success.
+   * @param chatId Unique identifier for the target private chat. If not specified, default bot's menu button will be changed.
+   * @param menuButton An object for the new bot's menu button.
```
KnorpelSenf:
```suggestion
   * @param menuButton An object for the bot's new menu button.
```
Or was this copied from the API? In that case, we should request the Telegram docs to be changed.

types.d.ts
```diff
@@ -0,0 +1,2 @@
+/** Experimental exports, shape of type exports may change in the future */
+export * from './typings/core/types/typegram'
```
KnorpelSenf, line 2: Is there any documentation about how this can be used, or how it should preferably be used? Not sure if docs are required, I'm just wondering how this will look

src/composer.ts
```diff
@@ -8,6 +8,12 @@ import Context from './context'
 export type MaybeArray<T> = T | T[]
 export type MaybePromise<T> = T | Promise<T>
 export type NonemptyReadonlyArray<T> = readonly [T, ...T[]]
+export type DefinitelyDefined<T> = T extends infer S | undefined ? S : T
+export type Expand<T> = T extends object
```
KnorpelSenf, line 11: Could you share what this does?

KnorpelSenf (approved):
LGTM, just a few nits here and there (optional to fix), plus a few questions about things I'm wondering about. Nothing that would be blocking.

src/composer.ts
MKRhere: Expands the type from being

```TS
Omit<tg.CallbackQuery, 'data'> & {
      game_short_name: DefinitelyDefined<tg.CallbackQuery['game_short_name']>
}
```

To just the object of keys and values. Just nicer UX.

KnorpelSenf: I see, thanks

src/telegram.ts
MKRhere: Yep, Telegram docs needs to be fixed
https://core.telegram.org/bots/api#setchatmenubutton

types.d.ts
MKRhere: No docs for now. As a first step, it allows consumers to use the same types Telegraf does, and I'd like to shape it better depending on how it's being used.

src/composer.ts
MKRhere: Just had a laugh at this, I forgot that NonNullable existed.

KnorpelSenf (approved):
LGTM

PR #1623, opened 2022-05-18
Add ctx.send* methods and optional useNewReplies from telegraf/future

MKRhere:
It has been a long-confusing feature of Telegraf that `ctx.reply` doesn't actually reply, but sends a message in the same chat. We now plan to fix it. This PR introduces `ctx.sendMessage`, `ctx.sendPhoto`, etc., that reflect the upstream API, but implicitly send a message in the same chat, more consistently with the other Context methods. This is the current behaviour of the `ctx.reply` methods.

This PR also introduces the new reply mode, which is optionally enabled using the `useNewReplies` middleware. Demonstration:

```TS
import { Telegraf } from "telegraf";
import { useNewReplies } from "telegraf/future";

const bot = new Telegraf(token);

// current behaviour:

bot.on("message", ctx => {
  // sends message in the same group, but doesn't reply
  ctx.reply("Hello");

  // sends message in the same group, but doesn't reply
  ctx.sendMessage("Hello");
});

bot.use(useNewReplies());

// new behaviour:

bot.on("message", ctx => {
  // actually replies to the message that was received
  ctx.reply("Hello");

  // sends message in the same group, but doesn't reply
  ctx.sendMessage("Hello");
});
```

This causes `ctx.reply`, `ctx.replyWithPhoto`, and the like to actually reply to the message in current context. If a message is not available, it will simply send a regular message.

A future minor release of telegraf will deprecate the current `ctx.reply*` methods in favour of the new ones, causing their usage to log warnings, prompting users to enable `useNewReplies`. v5 will completely replace the behaviour of `ctx.reply` with the new reply methods, ~~and an optional middleware, `useOldReplies`, to replace them with the old methods may be added.~~

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -2,6 +2,7 @@ import * as tg from './core/types/typegram'
 import * as tt from './telegram-types'
 import { Deunionize, PropOr, UnionKeys } from './deunionize'
 import ApiClient from './core/network/client'
+import { deprecate } from './util'
 import Telegram from './telegram'
 
 type Tail<T> = T extends [unknown, ...infer U] ? U : never
@@ -173,7 +174,10 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
     this.telegram.webhookReply = enable
   }
 
-  private assert<T extends string | object>(
+  /**
+   * @internal
+   */
+  assert<T extends string | object>(
     value: T | undefined,
     method: string
   ): asserts value is T {
@@ -351,9 +355,22 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
   /**
    * @see https://core.telegram.org/bots/api#sendmessage
    */
-  reply(this: Context, ...args: Shorthand<'sendMessage'>) {
-    this.assert(this.chat, 'reply')
+  sendMessage(this: Context, ...args: Shorthand<'sendMessage'>) {
+    this.assert(this.chat, 'sendMessage')
     return this.telegram.sendMessage(this.chat.id, ...args)
+  }
+
+  /**
+   * @see https://core.telegram.org/bots/api#sendmessage
+   */
+  reply(this: Context, ...args: Shorthand<'sendMessage'>) {
+    deprecate(
+      'ctx.reply',
+      'reply',
+      'sendMessage',
+      'https://telegraf.js.org/experimental#new-reply'
+    )
+    return this.sendMessage(...args)
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/future.ts
```diff
@@ -0,0 +1,209 @@
...
+
+type ReplyContext = { [key in keyof Context & `reply${string}`]: Context[key] }
+
+function makeReply<
+  C extends Context,
+  E extends { reply_to_message_id?: number }
+>(ctx: C, extra?: E) {
+  const ret = extra || ({} as E)
+  ret.reply_to_message_id = ctx.message?.message_id
+  return ret
```
wojpawlik: Avoid mutating users' arguments:
```suggestion
  const reply_to_message_id = ctx.message?.message_id
  return { reply_to_message_id, ... extra }
```

src/context.ts
```diff
@@ -573,117 +587,299 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
   }
 
   /**
-   * @see https://core.telegram.org/bots/api#replywithphoto
+   * @see https://core.telegram.org/bots/api#sendphoto
    */
-  replyWithPhoto(this: Context, ...args: Shorthand<'sendPhoto'>) {
+  sendPhoto(this: Context, ...args: Shorthand<'sendPhoto'>) {
     this.assert(this.chat, 'replyWithPhoto')
```
wojpawlik:
```suggestion
    this.assert(this.chat, 'sendPhoto')
```

src/context.ts
```diff
@@ -573,117 +587,299 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+      'https://telegraf.js.org/experimental#new-reply'
+    )
+    return this.sendPhoto(...args)
+  }
+
+  /**
+   * @see https://core.telegram.org/bots/api#sendmediagroup
+   */
+  sendMediaGroup(this: Context, ...args: Shorthand<'sendMediaGroup'>) {
     this.assert(this.chat, 'replyWithMediaGroup')
```
wojpawlik:
```suggestion
    this.assert(this.chat, 'sendMediaGroup')
```

src/context.ts
```diff
@@ -573,117 +587,299 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+      'https://telegraf.js.org/experimental#new-reply'
+    )
+    return this.sendDocument(...args)
+  }
+
+  /**
+   * @see https://core.telegram.org/bots/api#sendsticker
+   */
+  sendSticker(this: Context, ...args: Shorthand<'sendSticker'>) {
     this.assert(this.chat, 'replyWithSticker')
```
wojpawlik:
```suggestion
    this.assert(this.chat, 'sendSticker')
```

src/context.ts
```diff
@@ -573,117 +587,299 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+      'https://telegraf.js.org/experimental#new-reply'
+    )
+    return this.sendVideo(...args)
+  }
+
+  /**
+   * @see https://core.telegram.org/bots/api#sendanimation
+   */
+  sendAnimation(this: Context, ...args: Shorthand<'sendAnimation'>) {
     this.assert(this.chat, 'replyWithAnimation')
```
wojpawlik:
```suggestion
    this.assert(this.chat, 'sendAnimation')
```

src/util.ts
```diff
@@ -0,0 +1,40 @@
...
+  ctx: C,
+  value: T,
+  method: string
+): asserts ctx is C & { [k in T]: NonNullable<C[T]> } {
+  if (ctx[value] === undefined) {
+    throw new TypeError(
+      `Telegraf: "${method}" isn't available for "${ctx.updateType}"`
+    )
+  }
+}
```
wojpawlik: Why not just make `Context::assert` `@internal public` and consistently use it everywhere?

src/util.ts
```diff
@@ -0,0 +1,40 @@
...
+  const stack: { stack: string } = { stack: '' }
+  Error.captureStackTrace(stack)
+  const line = (stack.stack.split('\n')[3] || '').trim()
+
+  const useOther = use ? `; use ${use} instead` : ''
+  const pad = ' '.repeat('[WARN]'.length)
+
+  console.warn(`[WARN] ${method} is deprecated${useOther}`)
+  if (line) console.warn(pad, line)
+  if (see) console.warn(pad, `SEE ${see}`)
```
wojpawlik, line 25:
```suggestion
  process.emitWarning(`${method} is deprecated, use ${use} instead.`, 'DeprecationWarning')
```

wojpawlik (changes requested):
The history is a little messy already, I suggest merge-squashing or rebasing before merging.

> an optional middleware, `useOldReplies`, to replace them with the old methods may be added.

Just throw away the old behavior, that's the point of a major release.

src/util.ts
MKRhere: `process` is node-specific; we don't want telegraf to be more locked into node in the future

src/util.ts
MKRhere: What's `@internal`?

wojpawlik: https://www.typescriptlang.org/tsconfig#stripInternal

KnorpelSenf:
I will review this in a few days, my time is currently a bit limited. Please don't wait for me if prefer to merge before I get to this.

wojpawlik:
I see no reason to rush this

MKRhere:
@wojpawlik 

> I suggest merge-squashing or rebasing before merging

I have rebased, but prefer not to squash history. Rather than treat git as a changelog, I prefer to treat it as a devlog.

> Just throw away the old behavior

Okay, I've updated the PR description to reflect this

@KnorpelSenf 

> my time is currently a bit limited

No pressure; I might merge after some more testing to make sure nothing is breaking. If you get time meanwhile, appreciate a review.

wojpawlik (approved):

MKRhere:
This PR is ready to merge into feat-api-6.2, and I'll bundle them into one release at 4.9.0. However, I'm not merging yet so as to not clog reviews on #1671. When #1671 is approved and ready to be merged, I'll merge this first.

wojpawlik:
Why not merge this into v4 right now?

PR #1632, opened 2022-05-30
Add `types` reference also to the `exports` part to fix #1629

### LINKED ISSUE #1629: Typings resolution error with TS 4.7 in "Node16" mode
## Context

* Telegraf.js Version: 4.8.2
* Node.js Version: 16
* Operating System: Win

```ts
import * as telegraf from "telegraf";
```

With the (as of TS 4.7) released "native Node ESM module" support (activated via`"module": "node16"`) all "telegraf" imports fail with _"Could not find a declaration file for module 'telegraf'"_.

This is because in this mode TSC prefers the `"exports"` declaration in telegraf's package.json (and ignores the top-level `"types": "./typings/index.d.ts"`). See [this answer](https://github.com/microsoft/TypeScript/issues/49160#issuecomment-1137482639) from the TS team.

When locally changing `"exports": { ".":"./lib/index.js" }` to `"exports": {   ".": { "require": "./lib/index.js", "types": "./typings/index.d.ts" } }` the error disappears. This is also documented here: https://www.typescriptlang.org/docs/handbook/esm-node.html

Should I propose a PR with this fix?

ulrichb:
# Description / Type of change

Bugfix for #1629

See https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing for the TypeScript docs of the used `exports` change.

Note that I didn't change the `"./types"` exports entry because this seems to be only a workaround for some ESLint error, see 8ddeba23dba69e56f4da6aabdd14417ce7afeb1a.

# How Has This Been Tested?

When running `npx tsc -p .` in the repro sample (see https://github.com/telegraf/telegraf/issues/1629#issuecomment-1141257894) before the fix:

```plain
$npx tsc -p .
test.ts:1:27 - error TS7016: Could not find a declaration file for module 'telegraf'. 'C:/_work/3rdP/telegraf/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/telegraf` if it exists or add a new declaration (.d.ts) file containing `declare module 'telegraf';`

1 import * as telegraf from "telegraf"
                            ~~~~~~~~~~


Found 1 error in test.ts:1
```
... and after the fix it compiles fine.

# Checklist:

- n/a I have made corresponding changes to the documentation
- n/a My changes generate no new warnings
  -  Nope only a manual test, see above

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "main": "lib/index.js",
   "exports": {
-    ".": { "require": "./lib/index.js", "types": "./typings/index.d.ts" },
+    ".": { "types": "./typings/index.d.ts", "require": "./lib/index.js" },
     "./types": "./types.js"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "main": "lib/index.js",
   "exports": {
-    ".": "./lib/index.js",
+    ".": { "require": "./lib/index.js", "types": "./typings/index.d.ts" },
     "./types": "./types.js"
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -14,8 +14,14 @@
   },
   "main": "lib/index.js",
   "exports": {
-    ".": { "types": "./typings/index.d.ts", "require": "./lib/index.js" },
-    "./types": "./types.js"
+    ".": {
+      "types": "./typings/index.d.ts",
+      "require": "./lib/index.js"
+    },
+    "./types": {
+      "types": "./types.d.ts",
+      "require": "./types.js"
+    }
   },
   "files": [
     "bin/*",


MKRhere:
Does this continue working with current target modules that are not Node16 / NodeNext? And does this increase required TS version to `4.7.2`? If so, is there a backwards-compatible way to make this change?

ulrichb:
@MKRhere Very good question! As I've understood the docs, in `"module": "commonjs"` mode the `exports` field isn't used at all (instead the "older" top-level `main` and `types` is used). And therefore it should be backwards compatible.

But to go 100% sure, I've tested this PR change also with `"module": "commonjs"` with TS 3.9, 4.2, and 4.7 and it works fine 🙌.

package.json
```diff
@@ -14,7 +14,7 @@
   },
   "main": "lib/index.js",
   "exports": {
-    ".": "./lib/index.js",
+    ".": { "require": "./lib/index.js", "types": "./typings/index.d.ts" },
```
wojpawlik: > Note that the `"types"` condition should always come first in `"exports"`. (https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/)

```suggestion
    ".": { "types": "./typings/index.d.ts", "require": "./lib/index.js" },
```

package.json
```diff
@@ -14,7 +14,7 @@
...
     "./types": "./types.js"
```
wojpawlik:
```suggestion
    "./types": { "types": "./typings/types.d.ts", "require": "./lib/types.js" },
```

wojpawlik (changes requested):
Consider adding `"./scenes"`.

package.json
ulrichb: > Note that I didn't change the "./types" exports entry because this seems to be only a workaround for some ESLint error, see https://github.com/telegraf/telegraf/commit/8ddeba23dba69e56f4da6aabdd14417ce7afeb1a.

MKRhere:
> I didn't change the `"./types"` exports

Hope you don't mind that I went ahead and added this so we can unblock the PR. Thanks for the effort, cheers!

MKRhere (approved):

ulrichb:
@MKRhere Many thanks!

PR #1665, opened 2022-07-28
Support for API v6.1

hypernova7:
# Description

# Support for v6.1

## Type of change

  * Added `createInvoiceLink` method
  * Added `secretToken` option for webhook config
  * Improve `cb` option for webhook config, for better integration with http frameworks, like: fastify, express, etc.
  * Added an auto-filter that only handles webhook requests that match the `X-Telegram-Bot-Api-Secret-Token` header when `secretToken` is set
 
   ####  Note: You can add a manual filter using `cb` option

Basic example:

 ```javascript
bot.launch({
  webhook: {
    port: +process.env.PORT,
    domain: 'yourdomain.com',
    secretToken: 'my_secret_token',
    cb: webhookHandler => {
      return (req, res) => {
        if (req.headers['x-telegram-bot-api-secret-token'] === 'my_secret_token') {
          webhookHandler(req, res)
          return res.end()
        }
      }
    }
  }
})
 ```

Example using Express:
 ```javascript
const app = require('express')()

// ...

bot.launch({
  webhook: {
    port: +process.env.PORT,
    domain: 'yourdomain.com',
    secretToken: 'my_secret_token',
    cb: webhookHandler => {
      return app.use((req, res, next) => {
        if (req.headers['x-telegram-bot-api-secret-token'] === 'my_secret_token') {
          webhookHandler(req, res, next)
          return res.end()
        }
      })
    }
  }
})
 ```

# How Has This Been Tested?

**Test Configuration**:
* Node.js Version: 16.16.0
* Operating System: Ubuntu 20.04

# Checklist:

### DIFF
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -38,6 +38,10 @@ export namespace Telegraf {
     telegram?: Partial<ApiClient.Options>
   }
 
+  export interface WebhookCallback {
+    (webhookHandler: http.RequestListener): http.RequestListener
+  }
+
   export interface LaunchOptions {
     dropPendingUpdates?: boolean
     /** List the types of updates you want your bot to receive */
@@ -65,7 +69,9 @@ export namespace Telegraf {
       /** TLS server options. Omit to use http. */
       tlsOptions?: TlsOptions
 
-      cb?: http.RequestListener
+      secretToken?: string
+
+      cb?: Telegraf.WebhookCallback
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/telegraf.ts
```diff
@@ -135,18 +141,22 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
...
   }
 
-  private startWebhook(
-    hookPath: string,
-    tlsOptions?: TlsOptions,
-    port?: number,
-    host?: string,
-    cb?: http.RequestListener
-  ) {
+  private startWebhook(config: {
```
MKRhere: Wary of breaking change in a minor release, can we just accept `secretToken` as the last param for now and improve on it in v5?

MKRhere (changes requested):

src/telegraf.ts
hypernova7: Do you think it is necessary to leave it as it was?
The truth is that for my way of using Telegraf it is easier for me to use this method if only an object is passed as the only argument in case the Telegraf class is extended


```javascript
class NewTelegrafClass extend Telegraf {
  constructor(...) {
    super(...)
  }
  startWebhook({ anotherOption }) {
  }
}
 ```

MKRhere: Yes, I completely agree the object way is better; but it will break existing user code, which we must avoid in a non-major version. That's why I said we should do this change in v5

MKRhere (approved):

MKRhere:
Looks good, will do some changes in api-6.1 branch before release. Thanks

PR #1690, opened 2022-09-15
Formatting helpers

MKRhere:
Ref: [related conversation](https://t.me/TelegrafJSChat/83737)

Formatting is often troublesome, especially if you have many characters that need escaping. Escaping is messy, and it is much cleaner to create entities instead, but there were no utils to help with this in Telegraf so far. This PR adds support for `telegraf/format`. To be used like this:

```TS
import { fmt, bold, italic, underline, mention, link } from "telegraf/format";

ctx.reply(fmt`
Ground control to ${mention("Major Tom", 10000000)}
${bold`Lock your Soyuz hatch`} and ${italic`put your helmet on`}

— ${link(underline`David Bowie`), "https://en.wikipedia.org/wiki/David_Bowie")}
`);
```

This also just works with captions!

```TS
ctx.replyWithPhoto(
  file.id,
  { caption: fmt`${bold`File name:`} ${file.name}` },
);
```

For the future: `emoji()` may be a nice addition to formatting; however emojibase is quite large and would not be a suitable dependency for telegraf. Creating a separate `@telegraf/format-emoji` package could work.

### DIFF
--- a/src/core/helpers/formatting.ts
+++ b/src/core/helpers/formatting.ts
@@ -51,12 +51,6 @@ export function _fmt(kind: Types.Text | 'very-plain') {
     return new FmtString(text, entities.length ? entities : undefined)
   }
 }
-
-const entityOffset =
-  (length: number) =>
-  (entity: MessageEntity): MessageEntity =>
-    Object.assign(entity, { offset: length + entity.offset })
-
 export const linkOrMention = (
   content: string | FmtString,
   data:
@@ -64,10 +58,7 @@ export const linkOrMention = (
     | { type: 'text_mention'; user: User }
 ) => {
   const text = content.toString()
-  const length = text.length
-  const elems = FmtString.normalise(content).entities || []
-  const entities = elems.map(entityOffset(length))
-  // insert to start of array
-  entities.unshift(Object.assign(data, { offset: 0, length }))
+  const entities = FmtString.normalise(content).entities || []
+  entities.unshift(Object.assign(data, { offset: 0, length: text.length }))
   return new FmtString(text, entities)
 }
--- a/src/core/helpers/formatting.ts
+++ b/src/core/helpers/formatting.ts
@@ -1,6 +1,8 @@
 import { MessageEntity, User } from 'typegram'
 
 export class FmtString {
+  // to force parse_mode to become undefined when using FmtString
+  public parse_mode: undefined
   constructor(public text: string, public entities?: MessageEntity[]) {}
   static normalise(content: string | FmtString) {
     if (typeof content === 'string') return new FmtString(content, [])
--- a/src/core/helpers/formatting.ts
+++ b/src/core/helpers/formatting.ts
@@ -7,9 +7,12 @@ export interface FmtString {
 }
 
 export class FmtString implements FmtString {
-  constructor(public text: string, public entities?: MessageEntity[]) {
-    // force parse_mode to undefined if entities are present
-    if (entities) this.parse_mode = undefined
+  constructor(public text: string, entities?: MessageEntity[]) {
+    if (entities) {
+      this.entities = entities
+      // force parse_mode to undefined if entities are present
+      this.parse_mode = undefined
+    }
   }
   static normalise(content: string | FmtString) {
     if (typeof content === 'string') return new FmtString(content)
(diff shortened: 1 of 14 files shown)


Kamrulhasan12345 (approved):
Looks ok.

KnorpelSenf:
So effectively you're slowly integrating the most useful plugins into the core lib? That sounds like a better strategy than hoping for permission to publish packages under `@telegraf/` :D

MKRhere:
@KnorpelSenf I have access to npm `@telegraf/` now :) I don't have access to the GitHub org to create repositories yet. But this is the strategy for telegraf now - everything that's very useful and is lightweight (~100 loc) goes in the core.

I started this PR thinking of format being a tiny function in `Markup.fmt` which sounded very appropriate, but it's grown into a full file :D This is mostly because we don't want to clutter the main import space, but still want to import `{ bold, italic, ... }` as top-level items with import suggestions.

The benefit of in-core here is that user doesn't have to mess with Context/API types or implementation - even external plugins (like a future `@telegraf/format-emoji`) can use the `FmtString` contract and telegraf will know what to do.

src/telegram.ts
```diff
@@ -129,10 +131,11 @@ export class Telegram extends ApiClient {
    */
   sendMessage(
     chatId: number | string,
-    text: string,
+    text: string | FmtString,
     extra?: tt.ExtraReplyMessage
   ) {
-    return this.callApi('sendMessage', { chat_id: chatId, text, ...extra })
+    const t = typeof text === 'string' ? new FmtString(text) : text
+    return this.callApi('sendMessage', { chat_id: chatId, ...extra, ...t })
```
wojpawlik: This makes it impossible to manually pass custom `parse_mode`.

```suggestion
    const t = FmtString.normalise(text)
    return this.callApi('sendMessage', { chat_id: chatId, ...t, ...extra })
```

`editMessage*` should be updated as well.

src/core/helpers/formatting.ts
```diff
@@ -0,0 +1,75 @@
+import { MessageEntity, User } from 'typegram'
+
+export interface FmtString {
+  text: string
+  entities?: MessageEntity[]
+  parse_mode?: undefined
```
wojpawlik, line 6:
```suggestion
  parse_mode = undefined
```

src/core/helpers/formatting.ts
```diff
@@ -0,0 +1,75 @@
...
+    // force parse_mode to undefined if entities are present
+    if (entities) this.parse_mode = undefined
+  }
+  static normalise(content: string | FmtString) {
+    if (typeof content === 'string') return new FmtString(content)
+    return content
+  }
+  toString() {
+    return this.text
+  }
```
wojpawlik: Unnecessary?
```suggestion
```

MKRhere: https://github.com/telegraf/telegraf/blob/18a88154efb3b3fdd84cc988c2382c7724d771b7/src/core/helpers/formatting.ts#L71

Used here to convert `string | FmtString` to `string` without type-checking.

wojpawlik: You already `normalise` in the next line :unamused:

MKRhere: Lol, you're right. Fixed in f70829c729a348b2130c1284c3fd35e66a7bb3e5

src/telegram.ts
MKRhere: > ...impossible to manually pass custom `parse_mode`

Fixed in 18a88154efb3b3fdd84cc988c2382c7724d771b7

> `editMessage*` ...

Fixed in a0ece2e9b029ba71f786cfa0c8de7d24ded0b48b

wojpawlik (approved):

Viiprogrammer:
Interesting idea, but what about localization?

Many bots use it. I think the use of such solutions complicates the localization and translating of the text, even if i18n was not used now, may be needed in the future.

About escaping - [telegram-escape](https://github.com/mainarthur/telegram-escape)

MKRhere:
@Viiprogrammer That's a very good point - at the moment I don't think there's any way to reconcile fmt and i18n.

> may be needed in the future.

Any solution in this regard will need to be rewritten when i18n is needed - so not many ways to be prematurely prepared for it. I think the two needs are orthographic.

PR #1694, opened 2022-09-20
InputFile helpers

MKRhere:
Input helpers to create the InputFile object.

```TS
import { Telegraf, Input } from "telegraf";
const bot = new Telegraf(token);

bot.telegram.sendVideo(chatId, Input.fromLocalFile("../assets/cats.mp4"));

bot.telegram.sendDocument(chatId, Input.fromBuffer(buf));

bot.command("cat", ctx => {
  return ctx.sendPhoto(Input.fromURL("https://funny-cats.example/cats.jpg"))
});
```

This helps clear the confusion many users have about InputFile, and provides a layer of abstraction that can be used in the future to declutter telegraf/client from having to handle all combinations of InputFile.

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -5,14 +5,16 @@ import { InputFile } from './core/types/typegram'
  *
  * 10 MB max size for photos, 50 MB for other files.
  */
-export const fromLocalFile = (path: string): InputFile => ({ source: path })
+// prettier-ignore
+export const fromLocalFile = (path: string, filename?: string): InputFile => ({ source: path, filename })
 
 /**
  * The buffer will be uploaded as file to Telegram using multipart/form-data.
  *
  * 10 MB max size for photos, 50 MB for other files.
  */
-export const fromBuffer = (buffer: Buffer): InputFile => ({ source: buffer })
+// prettier-ignore
+export const fromBuffer = (buffer: Buffer, filename?: string): InputFile => ({ source: buffer, filename })
 
 /**
  * Contents of the stream will be uploaded as file to Telegram using multipart/form-data.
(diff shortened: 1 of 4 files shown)


wojpawlik (approved):
`fromBuffer` and `fromReadableStream` should accept `filename`.

MKRhere:
@wojpawlik Isn't that why there's `fromLocalFile` though? `fromBuffer` takes a buffer and `fromReadableStream` takes a stream and returns InputFile. Why overload them and go back to having ambiguous functions?

wojpawlik:
How do you name a buffer or stream?

src/input.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * * The target file must have the correct MIME type (e.g., audio/mpeg for `sendAudio`, etc.).
+ * * `sendDocument` with URL will currently only work for GIF, PDF and ZIP files.
+ * * To use `sendVoice`, the file must have the type audio/ogg and be no more than 1MB in size.
+ * 1-20MB voice notes will be sent as files.
+ *
+ * 5 MB max size for photos and 20 MB max for other types of content.
+ */
+// prettier-ignore
+export const fromURL = (url: string, filename?: string): InputFile => ({ url, filename })
```
wojpawlik:
```suggestion
export const fromUrl = (url: string, filename?: string): InputFile => ({ url, filename })
```

mkrnavana: Why?

MKRhere: Why? URL is an acronym. Either `URL` or `url` is fine, but why `Url`?

MKRhere:
> How do you name a buffer or stream?

Now I get what you mean - I did it this way because of this:

https://github.com/telegraf/telegraf/blob/e39398acddbaa80900460bc5a54c9f89fe04af7c/src/core/types/typegram.ts#L15-L32

Fixed in 4ad64f9e5d1cd4d742e8408fb0b7b03639ec5997

src/input.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Contents of the stream will be uploaded as file to Telegram using multipart/form-data.
+ *
+ * 10 MB max size for photos, 50 MB for other files.
+ */
+// prettier-ignore
+export const fromReadableStream = (stream: NodeJS.ReadableStream, filename?: string): InputFile => ({ source: stream, filename })
+
+/**
+ * Provide Telegram with an HTTP URL for the file to be sent.
+ * Telegram will download and send the file.
```
wojpawlik, line 37: Wrong.

```suggestion
 * Stream a given URL to Telegram.
```

wojpawlik (approved):

PR #1698, opened 2022-09-22
feat: `telegraf/filters`

### LINKED ISSUE #1388: Missing type definitions on ctx.message

wojpawlik:
## Type of change

## Checklist:


src/util.ts
```diff
@@ -28,6 +32,14 @@ export function fmtCaption<
   }
 }
 
+/** Construct a generic type guard */
+export type Guard<X = unknown, Y extends X = X> = (x: X) => x is Y
```
wojpawlik, line 36: https://github.com/telegraf/telegraf/blob/2d15150160e644af1fbbe0e3c6d3dc5bb2024dba/src/filters.ts#L15

wojpawlik (commented):
`Composer.guard` should probably be deprecated in favor of `Composer.on`.

src/util.ts
wojpawlik: There's a naming problem in general. `Composer.filter` doesn't accept `telegraf/filters`.

PR #1699, opened 2022-09-24
Useful options while using webhooks

threej-in:
# Description

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,17 +214,26 @@ const bot = new Telegraf(token);
 bot.on("text", ctx => ctx.reply("Hello"));
 
 // Start webhook via launch method (preferred)
-bot.launch({ 
+bot.launch({
   webhook: {
-    domain: webhookDomain, // https://example.com
-    port: port, // Use server's port
-    hookPath: webhookPath, // If not passed, Telegraf will generate a secure path for you.
-    secretToken: randomAlphaNumericString // Use "crypto.randomBytes(64).toString('hex')" to generate cryptographically secure string; Optional but recommmended for more security. See https://core.telegram.org/bots/api#setwebhook
-  }
+    // Public domain for webhook; e.g.: example.com
+    domain: webhookDomain,
+
+    // Port to listen on; e.g.: 8080
+    port: port,
+
+    // Optional path to listen for.
+    // `bot.secretPathComponent()` will be used by default
+    hookPath: webhookPath,
+
+    // Optional secret to be sent back in a header for security.
+    // e.g.: `crypto.randomBytes(64).toString("hex")`
+    secretToken: randomAlphaNumericString,
+  },
 });
 ```
 
-Use `webhookCallback()` if you want to attach Telegraf to an existing http server.
+Use `createWebhook()` if you want to attach Telegraf to an existing http server.
 
 <!-- global bot, tlsOptions -->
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -214,7 +214,14 @@ const bot = new Telegraf(token);
 bot.on("text", ctx => ctx.reply("Hello"));
 
 // Start webhook via launch method (preferred)
-bot.launch({ webhook: { domain: webhookDomain, port: port } });
+bot.launch({ 
+  webhook: {
+    domain: webhookDomain, // https://example.com
+    port: port, // 443
```
MKRhere: Not necessarily 443; if the server is behind a reverse proxy (recommended), this would be the upstream server's local port for Telegraf to listen on.

README.md
```diff
@@ -214,7 +214,14 @@ const bot = new Telegraf(token);
...
+    hookPath: webhookPath, // Application path on server
```
MKRhere: "If not passed, Telegraf will generate a secure path for you."

README.md
```diff
@@ -214,7 +214,14 @@ const bot = new Telegraf(token);
...
+    secretToken: randomAlphaNumericString // Optional but recommmended for more security. Also have to provide while setting the webhook url. see https://core.telegram.org/bots/api#setwebhook
```
MKRhere: `setWebhook` is not needed when using `bot.launch()`.  It's called internally.

Maybe also recommend how to securely generate the token. On Node, this could be: `crypto.randomBytes(64).toString('hex')`

MKRhere (changes requested):

README.md
threej-in: Done

README.md
threej-in: Okay replaced 443 example with "Use server's port"

README.md
threej-in: Specified token generation in the comment, please review it.

README.md
```diff
@@ -214,7 +214,14 @@ const bot = new Telegraf(token);
 bot.on("text", ctx => ctx.reply("Hello"));
 
 // Start webhook via launch method (preferred)
-bot.launch({ webhook: { domain: webhookDomain, port: port } });
+bot.launch({ 
+  webhook: {
+    domain: webhookDomain, // https://example.com
+    port: port, // Use server's port
+    hookPath: webhookPath, // If not passed, Telegraf will generate a secure path for you.
+    secretToken: randomAlphaNumericString // Use "crypto.randomBytes(64).toString('hex')" to generate cryptographically secure string; Optional but recommmended for more security. Also have to provide while setting the webhook url. see https://core.telegram.org/bots/api#setwebhook
```
MKRhere: > Also have to provide while setting the webhook url

This is extraneous - while using `bot.launch()`, this is the only place you need to pass `secretToken`. You don't need to, and shouldn't call `setWebhook` as well.

threej-in: From that I mean when user going to specify webhook to telegram api using get method they have to specify secret_token as parameter in the url.

MKRhere: `bot.launch()` calls `setWebhook` internally. You must NOT call `setWebhook` yourself.

https://github.com/telegraf/telegraf/blob/7907a7a79c468dbe7f15b6582a73aa57c643a923/src/telegraf.ts#L280-L287

threej-in: Ohh okay, I wasn't aware of this. Let me edit the comment.

threej-in: Done

MKRhere (approved):

PR #1701, opened 2022-09-24
feat: Add string to getUserProfilePhotos.userId

### LINKED ISSUE #1700: getUserProfilePhotos should probably takes number & string
For now, in the typescript definition of `getUserProfilePhotos` we have :
```typescript
getUserProfilePhotos(userId: number, offset?: number, limit?: number) {
  return this.callApi('getUserProfilePhotos', {
    user_id: userId,
    offset,
    limit,
  })
}
```

Since telegram id could be pretty big (cf. big int), it would be more wise to also add `string` in the supported type for userId, as it is the case for other method.

raskyer:
# Description
Close https://github.com/telegraf/telegraf/issues/1700

## Type of change

- Update of type

# How Has This Been Tested?

No test required

# Checklist:

### DIFF
--- a/src/telegram.ts
+++ b/src/telegram.ts
@@ -164,7 +164,7 @@ export class Telegram extends ApiClient {
     return this.callApi('sendChatAction', { chat_id: chatId, action })
   }
 
-  getUserProfilePhotos(userId: number, offset?: number, limit?: number) {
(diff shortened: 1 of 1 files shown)


wojpawlik (approved):

wojpawlik:
Strings are for usernames.

PR #1702, opened 2022-09-24
v5 Init: switch to @telegraf/client and type fixes

wojpawlik:
## Type of change

# Checklist:


src/telegram.ts
```diff
@@ -38,32 +35,6 @@ export class Telegram {
     return this.callApi('getFile', { file_id: fileId })
   }
 
-  /**
-   * Get download link to a file.
-   */
-  async getFileLink(fileId: string | tg.File) {
```
MKRhere: Are there any blockers to implement `downloadFile()` in its stead? (#1703)

wojpawlik: I wanna defer dealing with Client until transformers are being implemented.

MKRhere: If `this.#client.token` is still available, and `downloadFile` is not implemented yet - why remove `getFileLink` already?

Also, how do you see it being implemented with transformers?

wojpawlik: Both are simply heavily tied to Client. I removed it cuz I want to get to 0 type errors first.

MKRhere:
Tests are broken from the breaking changes, will just merge this and fix other problems in a separate commit.

src/telegram.ts
```diff
@@ -1,10 +1,32 @@
...
 import * as tt from './telegram-types'
-import ApiClient from './core/network/client'
+import { TT } from './telegram-types'
+import { Client, ClientOptions, createApi, type Opts } from '@telegraf/client'
 import { isAbsolute } from 'path'
 import { URL } from 'url'
+import { TelegramError } from './index'
+
+export class Telegram {
+  client: Readonly<Client>
```
wojpawlik, line 10: It must be `#private`, otherwise every middleware can access `ctx.telegram.client.context`.

wojpawlik: `Telegraf` must store `Client`.

PR #1705, opened 2022-09-26
feat: include `@telegraf/client`

### LINKED ISSUE #1703: Replace `Telegram::getFileLink` with `Telegram::downloadFile`

wojpawlik:
# Checklist:

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -3,7 +3,7 @@
 export type { Update, UserFromGetMe } from 'typegram'
 import type { ApiResponse, File, Typegram } from 'typegram'
 import createDebug from 'debug'
-import { fetch, RequestInit, FormData } from '../../vendor/fetch'
+import { fetch, FormData, type RequestInit } from '../../vendor/fetch'
 
 const debug = createDebug('telegraf:client')
 
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -3,12 +3,18 @@
 export type { Update, UserFromGetMe } from 'typegram'
 import type { ApiResponse, File, Typegram } from 'typegram'
 import createDebug from 'debug'
+import {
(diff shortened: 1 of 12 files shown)


src/core/network/client.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const defaultOptions: ClientOptions = {
+  api: {
+    mode: 'bot',
+    root: new URL('https://api.telegram.org'),
+  },
+}
+
+// based on https://github.com/node-fetch/fetch-blob/blob/8ab587d34080de94140b54f07168451e7d0b655e/index.js#L229-L241 (MIT License)
+export function isFile(value: any): value is InputFile {
```
wojpawlik, line 51: Probably should be renamed to `isInputFile`

MKRhere:
@wojpawlik Need your review on my changes:

* Node 18.x is at 100% ES2021, but not quite there with ES2022 - so I've reduced target and lib to ES2022; curiously, mentioning no lib seems to default to bringing lib.dom.d.ts, causing all DOM types to be available globally. Fixing this revealed the following other problems.
* fetch, FormData are not defined in @types/node yet
	- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924
* File is not exposed by Node as a global yet
	- https://github.com/nodejs/node/issues/39015
* I've went ahead and vendored just the types we want from https://github.com/nodejs/undici
* Input helpers reimplemented from https://github.com/telegraf/client/blob/0a3e4757efe179fd060a7b2c42c64516add020b4/src/input.node.ts with these differences:
	- Input.fromBuffer uses `ReadableStream.from(buffer)` instead of `new File` (since File is unavailable)
	- Input.fromURLStream is reimplemeted

MKRhere (approved):

PR #1710, opened 2022-10-04
Replace `FormData` with grammY's Deno-friendly `payload.ts`

wojpawlik:


src/core/network/payload.ts
```diff
@@ -0,0 +1,209 @@
...
+
+/**
+ * Calls `JSON.stringify` but removes `null` values from objects before
+ * serialization.
+ *
+ * @throws {TypeError} when encountering `InputFile`
+ */
+export function str(value: unknown) {
+  return JSON.stringify(value, (_, v) => {
+    if (v instanceof StreamFile) throw new TypeError() // caught by `createPayload`
```
MKRhere: This is a funny way to do it; but if we're doing this, I'd prefer it throw with a custom error so it doesn't swallow other possible `JSON.stringify` errors with invalid payloads.

MKRhere (changes requested):

MKRhere (approved):

PR #1713, opened 2022-10-07
Resist session race conditions using reference counting

### LINKED ISSUE #1372: `session` is prone to race conditions

### LINKED ISSUE #1465: scenes and wizards after dropping sessions
Hi. 
this library was the one of the coolest things that I've worked with.
but when i came back here after years. most of things not working e.g : scene and wizard.
i wanted to know is there any fix for that?

I wrote a bot with Scenes but it forget where he is every time and its not working well.

and another question is, is there any special reason that you do not have any documentation. i mean this library was great and have a good document make it easy for everyone to use it.

thanks a lot. hope it was a fix for the issue :)

### LINKED ISSUE #1695: Un-deprecate session until v5

Session has been deprecated for quite some time (#1372). While it may be appropriate, no satisfactory solution has been provided yet, and the earliest solution, if any, may not be expected before v5. Session is required by Scenes and Wizards, and hence all users are using it anyway. All third-party plugins are also currently affected.

I propose we un-deprecate session in v4, and wait for a possible solution in v5. We could mention that session is prone to racing and recommend [telegraf-throttler](https://www.npmjs.com/package/telegraf-throttler) in the TSDoc.

### LINKED ISSUE #1799: scene.enter occurs twice if messages come to fast
### Discussed in https://github.com/telegraf/telegraf/discussions/1798

<div type='discussions-op-text'>

<sup>Originally posted by **mvarakin** February 13, 2023</sup>
Hi,

I have a listener

bot.on(message('text'), async ctx => await ctx.scene.enter('my_scene')) 

in my code.

Inside the scene among other scene stuff (commands etc.) I catch all text messages and simply ignore them.

It works well if I chat with the bot.

But I noticed that if I forward two, three... messages to a bot at once from another chat, bot enters my_scene on every message!

But the bot is supposed to skip second, third... message as the bot should be already inside the scene after the first message coming.

Looks like it's a race condition. Entering the scene is still in progress but a new message comes and global listener (the above code) runs leading to call scene.enter again.</div>

MKRhere:
Fixes #1372, fixes #1695, fixes #1799
Supersedes #1436

This is a very involved, but well-commented solution. Short explanation follows.

Edit: This PR now also introduces `defaultSession` as an optional parameter to `session({})`. Unit tests have been updated.

### Current state:

1. Telegraf receives two messages from the same Telegram chat+user
2. Telegraf reads session from store in both updates
3. Both updates run through the middleware system
4. When they're done, they both update store, BUT one doesn't know that the other message also performed an update, so it overwrites session, losing data

### Solution:

1. Receive two messages concurrently
2. First update to fetch the session caches it in memory with a reference and counter
3. Everyone else checks the cache, and see that it's already in memory, so they just increment the counter and reuse the reference
4. When every update is done processing, counter is decremented
5. If counter reaches zero, the reference object is cleared from memory cache

### Problems:

1. ⚠️VERY IMPORTANT!
	~If an update is terminated halfway (because it threw an error), any changes it already made are not rolled back in the memory cache. If other concurrent requests are using the reference, they will see the updated object, and if they complete processing successfully, the updated session will persist! We could make this behaviour more predictable by also persisting session in the error catch branch.~
	This behaviour has been changed so session is always written back, regardless of whether middleware chain completes or errors. This means assigning to session is always almost as good as having written to store.
3. `getSessionKey` is async in v4, so it still allows some racing, but provided solution automatically corrects for this by ignoring duplicate store fetches if a reference is already cached when they return. In v5, `getSessionKey` should probably be made synchronous—counter-arguments welcome.
4. This solution does not protect the user from introducing race-conditions of their own. We cannot prevent them from doing that, but we can document this as a possible footgun.
5. Obviously any solution relying on synchronous in-memory cache does not scale horizontally. This solution will protect single-process session users only.
	* In the future (v5?), this may be expanded to use Worker threads / Web Workers where they synchronise using `SharedArrayBuffer` and `Atomics`. Will explore this if there is demand for such a thing.
	* Multi-process support (especially useful for FaaS and edge runtimes and probably more important than workers) could be implemented using db mutexes. But this also enforces sequentialisation across all processes for a given session key. Concurrency is easy; safe concurrency is hard. Maybe there are other solutions; we're not exploring them for now, but maybe in the future.
	* Meanwhile, we can recommend simply using a custom `getSessionKey :: ctx -> ctx.chat.id` for webhooks. Telegram ensures that webhook updates are serial for a given chatId already!
6. ~`session` has to be an object for references to be shared. It cannot be a primitive value. This is already true for Telegraf, because Scenes and Wizards cannot work otherwise.~ This is not true; since this PR, session can be a primitive, but we're leaving the type-level constraint in place until someone asks for it.

If there are concerns about this solution subtly changing the behaviour of existing deprecated `session`, particularly in multi-process situations, I'm open to discussion about splitting this into an external module at `@telegraf/session` which users can opt-in to, with the naive implementation as the default.

However, I don't like the idea of permanently moving session out of telegraf core as previously discussed in #1436, because other core (Scenes and Wizards) and thirdparty modules inherently rely on sessions to exist.

### Testing

The original reproduction code posted in #1372 works as expected with the proposed version of session. Try it yourself by installing this branch:

```shell
npm i "telegraf/telegraf#feat-better-session"
```

```TS
"use strict"

const { Telegraf, session } = require("telegraf")
const createDebug = require("debug")

const debug = createDebug('test')

bot.use(session())
bot.on('message', async ctx => {
  if (ctx.session === undefined) {
    const { update_id } = ctx.update
    ctx.session = { update_id }
    debug('Defaulting session to', ctx.session)
  } else {
    debug('Session already set to', ctx.session)
  }
})

createDebug.enable('telegraf:client test')
bot.launch()
```

Further, unit tests have been added with synchronous and asynchronous session stores. They appear rock-solid, but if anyone wants to produce a breaking test, I'd be happy to fix the implementation to conform.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -97,19 +97,15 @@ export function session<S extends object>(
       },
     })
 
-    const decrement = () => {
+    try {
+      await next()
+    } finally {
       if (cached && --cached.counter === 0) {
+        // decrement to avoid memory leak
         debug(`(${updId}) refcounter reached 0, removing cached`)
         cache.delete(key)
       }
-    }
-
-    try {
-      await next()
-
-      debug(`(${updId}) middlewares ran successfully, checking session`)
-
-      decrement()
+      debug(`(${updId}) middlewares completed, checking session`)
       if (ctx.session == null) {
         debug(`(${updId}) ctx.session missing, removing from store`)
         await store.delete(key)
@@ -117,10 +113,6 @@ export function session<S extends object>(
         debug(`(${updId}) ctx.session found, updating store`)
         await store.set(key, ctx.session)
       }
-    } catch (e) {
-      // decrement to avoid memory leak, and rethrow error
-      decrement()
-      throw e
     }
   }
 }
--- a/src/session.ts
+++ b/src/session.ts
@@ -13,6 +13,7 @@ export interface SessionStore<T> {
 interface SessionOptions<S extends object> {
   getSessionKey?: (ctx: Context) => Promise<string | undefined>
   store?: SessionStore<S>
+  defaultSession?: () => S
 }
 
 export interface SessionContext<S extends object> extends Context {
@@ -45,6 +46,8 @@ export function session<S extends object>(
   const concurrents = new Map<string, MaybePromise<S | undefined>>()
 
   // this function must be handled with care
+  // read full description on the original PR: https://github.com/telegraf/telegraf/pull/1713
+  // make sure to update the tests in test/session.js if you make any changes or fix bugs here
   return async (ctx, next) => {
     const updId = ctx.update.update_id
 
@@ -85,7 +88,7 @@ export function session<S extends object>(
         cached = c
       } else {
         // we're the first, so we must cache the reference
-        cached = { ref: upstream, counter: 0 }
+        cached = { ref: upstream ?? options?.defaultSession?.(), counter: 0 }
         cache.set(key, cached)
       }
     }
@@ -94,11 +97,15 @@ export function session<S extends object>(
     // It will, however, guard `c` here.
     const c = cached
 
+    let touched = false
+
     Object.defineProperty(ctx, 'session', {
       get() {
+        touched = true
         return c.ref
       },
       set(value: S) {
+        touched = true
         c.ref = value
       },
     })
@@ -112,13 +119,16 @@ export function session<S extends object>(
         cache.delete(key)
       }
       debug(`(${updId}) middlewares completed, checking session`)
-      if (ctx.session == null) {
-        debug(`(${updId}) ctx.session missing, removing from store`)
-        await store.delete(key)
-      } else {
-        debug(`(${updId}) ctx.session found, updating store`)
-        await store.set(key, ctx.session)
-      }
+
+      // only update store if ctx.session was touched
+      if (touched)
+        if (ctx.session == null) {
+          debug(`(${updId}) ctx.session missing, removing from store`)
+          await store.delete(key)
+        } else {
+          debug(`(${updId}) ctx.session found, updating store`)
+          await store.set(key, ctx.session)
+        }
     }
   }
 }
--- a/src/session.ts
+++ b/src/session.ts
@@ -11,9 +11,8 @@ export interface SessionStore<T> {
 }
 
 interface SessionOptions<S extends object> {
-  initial?: () => S
-  store?: SessionStore<S>
   getSessionKey?: (ctx: Context) => Promise<string | undefined>
+  store?: SessionStore<S>
 }
 
 export interface SessionContext<S extends object> extends Context {
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,6 +1,7 @@
 import { Context } from './context'
 import { MaybePromise } from './util'
 import { MiddlewareFn } from './middleware'
+import d from 'debug'
 
 export interface SessionStore<T> {
   get: (name: string) => MaybePromise<T | undefined>
@@ -9,8 +10,9 @@ export interface SessionStore<T> {
 }
 
 interface SessionOptions<S extends object> {
-  getSessionKey?: (ctx: Context) => Promise<string | undefined>
+  initial?: () => S
   store?: SessionStore<S>
+  getSessionKey?: (ctx: Context) => Promise<string | undefined>
 }
 
 export interface SessionContext<S extends object> extends Context {
@@ -20,7 +22,7 @@ export interface SessionContext<S extends object> extends Context {
 /**
  * Returns middleware that adds `ctx.session` for storing arbitrary state per session key.
  *
- * The default `getSessionKey` is <code>\`${ctx.from.id}:${ctx.chat.id}\`</code>.
+ * The default `getSessionKey` is `${ctx.from.id}:${ctx.chat.id}`.
  * If either `ctx.from` or `ctx.chat` is `undefined`, default session key and thus `ctx.session` are also `undefined`.
  *
  * Session data is kept only in memory by default,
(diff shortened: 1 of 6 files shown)


trgwii (approved):
LGTM

KnorpelSenf (changes requested):
This seems to break sessions that are primitive values., say `S` is `string`.

KnorpelSenf:
Ah telegraf cannot even do this so nvm

KnorpelSenf (approved):

src/session.ts
```diff
@@ -29,25 +31,96 @@ export interface SessionContext<S extends object> extends Context {
...
+  // when counter reaches 0, the cached ref will be freed from memory
+  const cache = new Map<string, { ref?: S; counter: number }>()
+  // temporarily stores concurrent requests
+  const concurrents = new Map<string, MaybePromise<S | undefined>>()
+
+  // this function must be handled with care
   return async (ctx, next) => {
+    const updId = ctx.update.update_id
+
+    const debug = d('telegraf:session')
```
wojpawlik: Nitpick: this should be at the top level

src/session.ts
```diff
@@ -29,25 +31,96 @@ export interface SessionContext<S extends object> extends Context {
...
+      })
+      cache.set(key, cached)
+    }
+
+    Object.defineProperty(ctx, 'session', {
+      get() {
+        return cached?.ref
+      },
+      set(value: S) {
+        if (cached) cached.ref = value
```
wojpawlik: What if `!cached`?

wojpawlik (commented):
This is quite complex, I'll need to re-read it before approving

wojpawlik (approved):

src/session.ts
```diff
@@ -29,25 +31,96 @@ export interface SessionContext<S extends object> extends Context {
...
  *
- * @example https://github.com/telegraf/telegraf/blob/develop/docs/examples/session-bot.ts
- * @deprecated https://github.com/telegraf/telegraf/issues/1372#issuecomment-782668499
+ * @example https://github.com/feathers-studio/telegraf-docs/blob/master/examples/session-bot.ts
  */
 export function session<S extends object>(
   options?: SessionOptions<S>
 ): MiddlewareFn<SessionContext<S>> {
   const getSessionKey = options?.getSessionKey ?? defaultGetSessionKey
   const store = options?.store ?? new MemorySessionStore()
```
wojpawlik, line 41:
```suggestion
  const store = options?.store ?? new Map()
```

MKRhere: Let's make this change in v5. This PR targets v4.

src/session.ts
MKRhere: `cached` is always set by this line, TS is not able to tell.

wojpawlik: Non-null assertion or a comment would clarify.

src/session.ts
```diff
@@ -9,8 +11,9 @@ export interface SessionStore<T> {
 }
 
 interface SessionOptions<S extends object> {
-  getSessionKey?: (ctx: Context) => Promise<string | undefined>
+  initial?: () => S
```
wojpawlik: Unused.

MKRhere: I've removed `initial` for now. It can be added later if required. This PR is large enough.

Viiprogrammer:
> This seems to break sessions that are primitive values., say `S` is `string`.

This is not a breaking change because it was never said anywhere that sessions could be strings 😁

MKRhere:
> This seems to break sessions that are primitive values., say `S` is `string`.

Just to clarify, this is only a type-level constraint that has already existed on Telegraf, and is useful because scenes need session to be an object. However, since this PR, it is not a technical requirement, and sessions can be primitive. After discussing with @wojpawlik, we decided to leave the constraint be until someone asks for it.

MKRhere:
FINALLY!

PR #1714, opened 2022-10-09
Fix default scene session pollution

uqee:
The bug can be reproduced by trying to do `ctx.scene.leave()` inside `scene.enter(() => {...})`.


MKRhere (approved):
lgtm, thanks!

PR #1763, opened 2022-12-17
Extended Composer.acl parameter with predicate types

Jokero:
To allow the use of `ctx => boolean` functions with `Composer.acl` in TypeScript

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -722,13 +722,13 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
    * Generates middleware responding only to specified users.
    */
   static acl<C extends Context>(
-    userId: MaybeArray<number>,
+    predicateOrUserId: Predicate<C> | AsyncPredicate<C> | MaybeArray<number>,
     ...fns: NonemptyReadonlyArray<Middleware<C>>
   ): MiddlewareFn<C> {
-    if (typeof userId === 'function') {
-      return Composer.optional(userId, ...fns)
+    if (typeof predicateOrUserId === 'function') {
+      return Composer.optional(predicateOrUserId, ...fns)
     }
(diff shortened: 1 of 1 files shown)


MKRhere:
`Composer.optional` already accepts a `ctx -> boolean` predicate. `acl` is access control list, as the JSDoc says, _generates middleware responding only to specified users._

If you're in a position where you have a function or userId, you should wrap the call in userland.

```TS
const middleware = typeof x === "function" ? Composer.optional(x) : Composer.acl(x);
```

Jokero:
@MKRhere `Composer.acl` already supports predicate (works with JavaScript code), I just extended the parameter type for TypeScript to make the cases with function compilable.

Take a look at `typeof userId === 'function'` (https://github.com/telegraf/telegraf/blob/v4.11.2/src/composer.ts#L728-L730):

```ts
static acl<C extends Context>(
  userId: MaybeArray<number>,
  ...fns: NonemptyReadonlyArray<Middleware<C>>
): MiddlewareFn<C> {
  if (typeof userId === 'function') {
    return Composer.optional(userId, ...fns)
  }
  const allowed = Array.isArray(userId) ? userId : [userId]
  // prettier-ignore
  return Composer.optional((ctx) => !ctx.from || allowed.includes(ctx.from.id), ...fns)
}
```
Current code looks inconsistent from interface perspective. It either should be updated by removing type check (`typeof userId === 'function'`) or by adding the correct type for the first parameter.

MKRhere:
I'm aware some methods do more than the types suggest, however it is a bad idea to have multiple methods do the same thing. These behavioural inconsistencies should be removed in v5.

Jokero:
Generally agree, thanks

PR #1787, opened 2023-01-26
merge: #1787 from mikemikhaylov

mikemikhaylov:
Hi! Just reordering the chats by the number of members since the ru-community grew a lot since https://github.com/telegraf/telegraf/pull/1663/ Thanks again @MKRhere for giving it a chance!


MKRhere (approved):
Awesome, thanks for keeping it going!

PR #1804, opened 2023-02-22
feat: `Context::persistentChatAction`

### LINKED ISSUE #1801: Feature Request: Continuously Sending Chat Action Requests
# Description
I'm frustrated when the sendChatAction method stops after 5-6 seconds. This can be problematic when I want to run a function that takes a longer time to process, as the user may assume that the bot has stopped responding. Therefore, I'd like to request a new feature that allows for the continuous sending of chat action requests for a specified amount of time.
I'd like to request this feature, and it would be an honor to contribute and help implement it. I have experience working with Typescript and would be happy to collaborate with the development team to ensure that this feature is implemented in a way that meets their standards.

# Solution
My proposed solution is to send the sendChatAction request repeatedly, given a specified amount of time. This would allow the bot to give the user a sense of its responsiveness, and prevent the user from sending multiple requests while the bot is still processing a previous request.

# Example
For example, the "typing" chat action could be sent continuously for 20 seconds, with a delay of 1-2 seconds between requests. This would ensure that the user is aware that the bot is still processing their request, and give them a better experience overall.

# Considerations
I understand that this feature may impact the performance of the bot or the Telegram servers, and I'd be happy to discuss any potential trade-offs or drawbacks with the development team.

https://user-images.githubusercontent.com/97595296/219793214-d137b6ab-e9d8-4479-abea-4b03d23744b0.mp4

orimiles5:
Fixes #1801

# Description
This pull request introduces a new feature to allow for the continuous sending of chat action requests while running long processes, in order to improve the user experience with bots.

The `persistentChatAction` utility sends the `sendChatAction` request repeatedly, with a delay between requests, as long as the provided callback function is being processed. This gives the user a sense of the bot's responsiveness while they wait for a long process to complete. To use this utility, simply call `persistentChatAction` with the `ctx` object (or as a utility), the chat action type (e.g. 'typing'), and the long process as a callback function.

To use the feature, there are two ways:
As @MKRhere suggested:
```ts
bot.start(async (ctx) => {
  persistentChatAction(
    ctx,
    async () => {
      // some prolonged actions...
      await ctx.reply('done')
    },
    'typing'
  )
})
```

As @wojpawlik suggested:
```ts
bot.start(async (ctx) => {
  await ctx.persistentChatAction('typing', async () => {
    // some prolonged actions...
    await ctx.reply('done')
  })
})
```

Alternatively, you can use the `timedChatAction` utility to send a chat action continuously for a specified amount of time. 
To use this utility, simply call `timedChatAction` with the `ctx` object, the chat action type, and the duration in milliseconds. For example:

```ts
bot.start(async ctx => {
    await timedChatAction(ctx, "typing", 10000) // "typing" for 10 seconds 
    // some code...
})
```

I have tested the implementation and it works as expected.

I would appreciate feedback from you, including any potential issues or ideas for improvement. I am open to collaboration and willing to work with you to ensure that this feature meets your standards and benefits the community.

Thank you for considering this feature.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -896,8 +896,11 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
       extra?.intervalDuration ?? 4000
     )
 
-    await callback()
-    clearInterval(timer)
+    try {
+      await callback()
+    } finally {
+      clearInterval(timer)
+    }
   }
 
   /**
--- a/src/context.ts
+++ b/src/context.ts
@@ -870,12 +870,12 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
 
   /**
    * @see https://core.telegram.org/bots/api#sendchataction
-   * 
-   * Sends the sendChatAction request repeatedly, with a delay between requests, 
+   *
+   * Sends the sendChatAction request repeatedly, with a delay between requests,
    * as long as the provided callback function is being processed.
-   * 
+   *
    * The sendChatAction errors should be ignored, because the goal is the actual long process completing and performing an action.
-   * 
+   *
    * @param action - chat action type.
    * @param callback - a function to run along with the chat action.
    * @param extra - extra parameters for sendChatAction.
--- a/src/context.ts
+++ b/src/context.ts
@@ -865,18 +865,18 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
     })
   }
 
-  persistentChatAction(action: Shorthand<'sendChatAction'>[0], callback: Function, extra?: tt.ExtraSendChatAction) {
+  persistentChatAction(action: Shorthand<'sendChatAction'>[0], callback: () => Promise<void>, extra?: tt.ExtraSendChatAction) {
     return new Promise<void>(async (resolve) => {
       await this.sendChatAction(action, {...extra})
+
       const timer = setInterval(
         async () => await this.sendChatAction(action, {...extra}),
-        8000
+        extra?.intervalDuration ?? 8000
       )
-  
-      await Promise.resolve(callback()).then(async () => {
-        clearInterval(timer)
-        resolve()
-      })
+        
+      await callback();
+      clearInterval(timer);
+      resolve();
     })
   }
 
--- a/src/context.ts
+++ b/src/context.ts
@@ -5,6 +5,9 @@ import ApiClient from './core/network/client'
 import { Guard, Guarded, MaybeArray } from './util'
 import Telegram from './telegram'
 import { FmtString } from './format'
+import d from 'debug'
+
+const debug = d('telegraf:context')
 
 type Tail<T> = T extends [unknown, ...infer U] ? U : never
 
--- a/src/context.ts
+++ b/src/context.ts
@@ -859,7 +859,7 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
    */
   sendChatAction(
     action: Shorthand<'sendChatAction'>[0],
-    extra?: Omit<tt.ExtraSendChatAction, 'intervalDuration'>
+    extra?: tt.ExtraSendChatAction
   ) {
     this.assert(this.chat, 'sendChatAction')
     return this.telegram.sendChatAction(this.chat.id, action, {
--- a/src/context.ts
+++ b/src/context.ts
@@ -856,7 +856,7 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
    */
   sendChatAction(
     action: Shorthand<'sendChatAction'>[0],
-    extra?: tt.ExtraSendChatAction
+    extra?: Omit<tt.ExtraSendChatAction, 'intervalDuration'>
   ) {
     this.assert(this.chat, 'sendChatAction')
     return this.telegram.sendChatAction(this.chat.id, action, {
--- a/src/context.ts
+++ b/src/context.ts
@@ -888,15 +888,16 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
   ) {
     await this.sendChatAction(action, { ...extra })
 
-      const timer = setInterval(
-        async () => await this.sendChatAction(action, {...extra}),
-        extra?.intervalDuration ?? 8000
-      )
-        
-      await callback();
-      clearInterval(timer);
-      resolve();
-    })
+    const timer = setInterval(
+      async () =>
+        await this.sendChatAction(action, { ...extra }).catch((err) => {
+          debug('Ignored error while persisting sendChatAction:', err)
+        }),
+      extra?.intervalDuration ?? 8000
+    )
+
+    await callback()
+    clearInterval(timer)
   }
 
   /**
--- a/src/context.ts
+++ b/src/context.ts
@@ -892,11 +892,11 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
     await this.sendChatAction(action, { ...extra })
 
     const timer = setInterval(
-      async () =>
-        await this.sendChatAction(action, { ...extra }).catch((err) => {
+      () =>
+        this.sendChatAction(action, { ...extra }).catch((err) => {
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -61,3 +64,39 @@ export function* zip<X, Y>(xs: Iterable<X>, ys: Iterable<Y>): Iterable<X | Y> {
     y1 = y.next()
   }
 }
+
+export async function persistentChatAction(
```
MKRhere: Keep only one version of this implementation. The Context method is clearly the superior; I only suggested it as a utility to avoid adding methods to Context. But since this is useful enough, let's keep the Context version.

src/util.ts
```diff
@@ -61,3 +64,39 @@ export function* zip<X, Y>(xs: Iterable<X>, ys: Iterable<Y>): Iterable<X | Y> {
...
+      every
+    )
+
+    await callback()
+    clearInterval(timer)
+    resolve()
+  })
+}
+
+export async function timedChatAction(ctx: any, action: ChatAction, duration: number, every?: number) {
```
MKRhere: `timedChatAction` seems wasteful; not clear where it can be useful, but where it is, it can be implemented rather trivially:

```TS
// sendChatAction will be called for 10 seconds
await ctx.persistentChatAction("typing", () => sleep(10000));
```

src/context.ts
```diff
@@ -865,6 +865,21 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
     })
   }
 
+  persistentChatAction(action: Shorthand<'sendChatAction'>[0], callback: () => Promise<void>, extra?: tt.ExtraSendChatAction) {
+    return new Promise<void>(async (resolve) => {
+      await this.sendChatAction(action, {...extra})
+
+      const timer = setInterval(
```
MKRhere: Using a `setInterval` loop makes any errors in `sendChatAction` unhandle-able. This is bad because the process will crash if it errors.

But also we may want such errors to be ignored, because what we really care about is the actual long process completing and performing an action. I do not think we'd ever want a failure in `sendChatAction` to abort the handler early. So the best course of action is to do:

```TS
setInterval(
	async () => this
		.sendChatAction(action, { ...extra })
		.catch(err => { debug("Ignored error while persisting sendChatAction:", err) })
	...
```

You'll need to instantiate debug as `d("telegraf:context")` in this file, similar to

https://github.com/telegraf/telegraf/blob/5137c75b249898cfdf35bb1560d649f6a7658e62/src/telegraf.ts#L19

This error-ignoring behaviour must be documented in the method's JSDoc, along with a description of what the function itself.

src/context.ts
```diff
@@ -865,6 +865,21 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
     })
   }
 
+  persistentChatAction(action: Shorthand<'sendChatAction'>[0], callback: () => Promise<void>, extra?: tt.ExtraSendChatAction) {
```
MKRhere: Perhaps you tried adding `intervalDuration` to `ExtraSendChatAction` and reverted it, so it no longer exists in `persistentChatAction` extra as well. Make this

```TS
extra?: tt.ExtraSendChatAction & { intervalDuration?: number }
```

src/context.ts
```diff
@@ -856,7 +856,7 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
    */
   sendChatAction(
     action: Shorthand<'sendChatAction'>[0],
-    extra?: tt.ExtraSendChatAction
+    extra?: Omit<tt.ExtraSendChatAction, 'intervalDuration'>
```
MKRhere: Revert this to the original form since you seem to have reverted changes to `ExtraSendChatAction`.

src/context.ts
```diff
@@ -865,6 +865,21 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
     })
   }
 
+  persistentChatAction(action: Shorthand<'sendChatAction'>[0], callback: () => Promise<void>, extra?: tt.ExtraSendChatAction) {
+    return new Promise<void>(async (resolve) => {
```
MKRhere: The `new Promise` construction is unnecessary; make the function `async persistentChatAction (action: ...`

MKRhere (changes requested):
Awesome, thanks for the PR! I have a few thoughts I've added as review comments.

Also remember to run `npm run lint`, and `npm run build` so you don't commit with errors.

orimiles5:
Made changes following your comments.
Thank you for the feedback! 
Tell me if there is something else I can change.

MKRhere (approved):
Lgtm, will wait for @wojpawlik to have a look as well.

src/context.ts
```diff
@@ -865,6 +868,38 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+    const timer = setInterval(
+      async () =>
+        await this.sendChatAction(action, { ...extra }).catch((err) => {
+          debug('Ignored error while persisting sendChatAction:', err)
+        }),
+      extra?.intervalDuration ?? 8000
+    )
+
+    await callback()
+    clearInterval(timer)
```
wojpawlik: In case of exceptions:

```suggestion
    try {
      await callback()
    } finally {
      clearInterval(timer)
    }
```

src/context.ts
```diff
@@ -865,6 +868,38 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+    extra?: tt.ExtraSendChatAction & { intervalDuration?: number }
+  ) {
+    await this.sendChatAction(action, { ...extra })
+
+    const timer = setInterval(
+      async () =>
+        await this.sendChatAction(action, { ...extra }).catch((err) => {
+          debug('Ignored error while persisting sendChatAction:', err)
+        }),
+      extra?.intervalDuration ?? 8000
```
wojpawlik:
```suggestion
      extra?.intervalDuration ?? 4000
```

https://core.telegram.org/bots/api#sendchataction

wojpawlik (changes requested):

orimiles5:
Fixed. 
Thank you! @wojpawlik

src/context.ts
```diff
@@ -865,6 +868,41 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+  async persistentChatAction(
+    action: Shorthand<'sendChatAction'>[0],
+    callback: () => Promise<void>,
+    extra?: tt.ExtraSendChatAction & { intervalDuration?: number }
+  ) {
+    await this.sendChatAction(action, { ...extra })
+
+    const timer = setInterval(
+      async () =>
+        await this.sendChatAction(action, { ...extra }).catch((err) => {
```
wojpawlik: Nitpick: unnecessary `async` / `await`
```suggestion
      () =>
        this.sendChatAction(action, { ...extra }).catch((err) => {
```

src/context.ts
```diff
@@ -865,6 +868,41 @@ export class Context<U extends Deunionize<tg.Update> = tg.Update> {
...
+   *
+   * @param action - chat action type.
+   * @param callback - a function to run along with the chat action.
+   * @param extra - extra parameters for sendChatAction.
+   * @param {number} [extra.intervalDuration=8000] - The duration (in milliseconds) between subsequent sendChatAction requests.
+   */
+  async persistentChatAction(
+    action: Shorthand<'sendChatAction'>[0],
+    callback: () => Promise<void>,
+    extra?: tt.ExtraSendChatAction & { intervalDuration?: number }
```
wojpawlik: Nitpick: this sends `intervalDuration` to bot API.
```suggestion
    {
      intervalDuration,
      ...extra
    }: tt.ExtraSendChatAction & { intervalDuration?: number } = {}
```

wojpawlik (approved):

orimiles5:
Fixed.

MKRhere:
Thanks @orimiles5!

orimiles5:
Thank you for the opportunity to contribute!! 
@MKRhere @wojpawlik

PR #1840, opened 2023-05-09
fix: allow thumbs in media objects

lmcsu:
Fix to make Telegraf handle thumbnails in cases like the following

```ts
bot.telegram.editMessageMedia(111, 222, undefined, {
    type: 'audio',
    media: {
        source: audioBuffer,
    },
    thumb: {
        source: thumbBuffer,
    },
});
```

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -198,13 +198,31 @@ async function attachFormValue(
   ) {
     const attachmentId = crypto.randomBytes(16).toString('hex')
     await attachFormMedia(form, value.media as FormMedia, attachmentId, agent)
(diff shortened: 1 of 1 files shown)


MKRhere:
I looked at this several times over the months, and it never made sense to me. Anyway, uploading thumbnails should already work well. If it doesn't, post a repro and I'll reopen.

lmcsu:
@MKRhere it SHOULD, but it DOES NOT.
Exactly when editing media.
This PR was fixing it.
Anyway, this repo is a piece of buggy trash, I moved to Grammy, thank you.

PR #1842, opened 2023-05-18
fix: Missing type for param in editMessageText

fjodor-rybakov:
The `text` param in the `editMessageText` function can be either `string` or `FmtString`


MKRhere (approved):

MKRhere:
Thanks!

PR #1862, opened 2023-07-20
Add support for Telegram API 6.6, 6.7, and 6.8

hypernova7:
@MKRhere please review this changes

### DIFF
--- a/src/telegram-types.ts
+++ b/src/telegram-types.ts
@@ -22,12 +22,7 @@ export type ChatAction = Opts<'sendChatAction'>['action']
 
 // Modify type so caption, if exists, can be FmtString
 export type WrapCaption<T> = T extends { caption?: string }
-  ? Expand<
-      Omit<T, 'caption'> & {
-        caption?: string | FmtString
-        message_thread_id?: number | undefined
-      }
-    >
+  ? Expand<Omit<T, 'caption'> & { caption?: string | FmtString }>
   : T
 
 // extra types
--- a/src/telegram-types.ts
+++ b/src/telegram-types.ts
@@ -22,7 +22,12 @@ export type ChatAction = Opts<'sendChatAction'>['action']
 
 // Modify type so caption, if exists, can be FmtString
 export type WrapCaption<T> = T extends { caption?: string }
-  ? Expand<Omit<T, 'caption'> & { caption?: string | FmtString }>
+  ? Expand<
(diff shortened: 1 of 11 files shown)


chakzefir:
@hypernova how about to bump `answerInlineQuery`(https://core.telegram.org/bots/api#answerinlinequery) and add `button` param with TWA support?

MKRhere:
Seems like TypeScript dropped support for Node 12. Since it's been EOL since over a year ago, should we drop it from CI testing at least, to make things easier for ourselves?

MKRhere:
@chakzefir 

Did you mean this?
```TS
await ctx.answerInlineQuery([], { button: { text: "title", web_app: { url: "url" } } });
```

This currently works as per this PR.

I considered adding a button helper, but it does more harm than good:

```TS
await ctx.answerInlineQuery([], { button: Markup.button.inlineQueryResultsWebApp("title", "url") });
```

I think we can leave this at that.

MKRhere:
So... Logical operators like `??=` only have support from Node 15. We don't use it, but the tsc ecosystem seems to have moved on, so we're still having CI issues with the linter.

Node 14 is also EOL since 4 months ago, so I'm gonna cut it out of CI as well. We'll still fix bugs if someone reports issues with Node 14 until Telegraf v4.

src/telegram-types.ts
```diff
@@ -22,7 +22,12 @@ export type ChatAction = Opts<'sendChatAction'>['action']
 
 // Modify type so caption, if exists, can be FmtString
 export type WrapCaption<T> = T extends { caption?: string }
-  ? Expand<Omit<T, 'caption'> & { caption?: string | FmtString }>
+  ? Expand<
+      Omit<T, 'caption'> & {
+        caption?: string | FmtString
+        message_thread_id?: number | undefined
```
MKRhere: Why was this necessary here? This type exists only to convert caption from `string` to `FmtString | string`.

MKRhere (changes requested):

src/telegram-types.ts
MKRhere: This type is only used in `editMessageMedia`, where `media` does not take `message_thread_id`, so this change is wrong; I'll remove it.

MKRhere (approved):

MKRhere:
I ended up adding API 6.8 support as well in this PR; feel free to review changes if you can.

PR #1870, opened 2023-08-27
feat: custom property in sessions

### LINKED ISSUE #1824: grammY offers the possibility to have multiple sessions, can Telegraf have that too?
I want to store all the messages from all the chats that my bot is a part of. However, since currently my bot stores sessions per user, that means that in a group chat, it will be storing the messages of that group chat in a fragmented way, only grouped per user.

I'd like to have multiple sessions, based on multiple session keys.
One session called `chat` which is based on `ctx.chat.id` as the session key, and it would contain a property called `messages`, storing all the messages in a chat.
And then I'd like to have a session called `user` which has `ctx.from.id` as the session key, and it would contain properties like `settings` related to the settings that the user has set.

I didn't think of any alternatives tbh.

MKRhere:
Long ago, Telegraf had the ability to specify custom property name while creating session. [This was lost `51dfca`](https://github.com/telegraf/telegraf/commit/51dfca4e3df3cdcc0ab785ad5ce0f3e1210ccbf4) because it was deemed too hard to write types for in v4. However, with this PR, we regain this ability in full.

Although the usage below seems like a trivial addition, this took several days and at least one sleepless night to type correctly. Thanks for the patience!

### Usage

```TS
// defaults to ctx.session
bot.use(session({ defaultSession: () => ({}) }));

bot.use(session({
	property: "chatSession",
	getSessionKey: ctx => String(ctx.chat.id),
	defaultSession: () => ({ count: 0 }),
}));

bot.on("message", ctx => {
	ctx.session.seen = true;
});

bot.on("message", ctx => {
	ctx.chatSession.count++;
});
```

I'm still investigating Stage/Scene support for custom properties.

Fixes #1824

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -26,6 +26,7 @@ interface SessionOptions<S, C extends Context, P extends string> {
   defaultSession?: (ctx: C) => S
 }
 
+/** @deprecated session can use custom properties now. Construct this type directly. */
 export interface SessionContext<S extends object> extends Context {
   session?: S
 }
@@ -185,6 +186,7 @@ export class MemorySessionStore<T> implements SyncSessionStore<T> {
   }
 }
 
+/** @deprecated session can use custom properties now. Directly use `'session' in ctx` instead */
 export function isSessionContext<S extends object>(
   ctx: Context
 ): ctx is SessionContext<S> {
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,5 +1,5 @@
 import { Context } from './context'
-import { MaybePromise } from './util'
+import { ExclusiveKeys, MaybePromise } from './util'
 import { MiddlewareFn } from './middleware'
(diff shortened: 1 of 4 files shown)


hypernova7 (approved):

MKRhere:
Update: Stage/Scene support for custom prop is not planned for this release. If it happens, it'll be a future PR. This PR is ready to merge.

Evertt:
@MKRhere, I'm a bit confused. Will this merge be delayed to be integrated into the next release? And if so, what's the timeline you're expecting?

MKRhere:
> Will this merge be delayed to be integrated into the next release?

No. Support for custom session prop in <ins>scenes</ins> is delayed. Custom prop in session (this PR) is already merged into v4 and will release today, along with API updates.

PR #1880, opened 2023-09-11
WIP refactor: use WebCrypto API for hashing instead of node:crpyto

### LINKED ISSUE #1879: SHA3 unsupported algorithm in some other runtimes
Telegraf is using SHA3 to hash secret, which broke my codebase after porting it to Bun instead of Nodejs.

Then, I dig deeper into SHA3 usage in the ecosystem in general, which seems like a drop-replacement for SHA2 if it is broken (future-proof), and I was surprised that many runtimes do not support SHA3. For instance,  Google/BoringSSL/Chrome might stand in the way because they might [skip SHA-3](https://www.imperialviolet.org/2017/05/31/skipsha3.html). See also https://github.com/w3c/webappsec-subresource-integrity/issues/11 and https://github.com/electron/electron/issues/30530.

I propose that Telegraf replace any use for SHA3 with SHA2 to make this framework compatible with other runtimes. We can return to SHA3 if there is a real need for it.

If the proposal is accepted, I would happily open PR for this issue.

Khalid-Nowaf:
Resolves #1879

### DIFF
--- a/src/platform/crypto.ts
+++ b/src/platform/crypto.ts
@@ -1,9 +1,0 @@
-import { createHash } from 'https://deno.land/std@0.165.0/node/internal/crypto/hash.ts'
-import { process } from './process.ts'
-
-export function hash(message: string) {
-  return createHash('SHA3-256')
-    .update(message)
-    .update(process.version) // salt
-    .digest('hex')
-}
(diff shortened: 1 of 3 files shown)


MKRhere:
I just noticed you marked it as WIP — sorry, did you intend to make improvements before merge?

Khalid-Nowaf:
Yes, I do have a typo, but I will make another PR for it. 

Thanks

MKRhere:
I thought you should know, version was used as hash so that while token can't be sniffed in transit, it's predictably generated for your environment. That way if you call it multiple times you get the same hash. Fixed in 5b46fadd1808032f6b47def6aa9f363f5d223224

PR #1893, opened 2023-10-16
Fix: add parse_mode type to FmtString

deadletterq:


MKRhere:
`parse_mode` is explicitly set to undefined to force `FmtString` to override any explicit `parse_mode` (incorrectly) set by the user. `fmt` is mutually exclusive with `parse_mode`.

deadletterq:
> `parse_mode` is explicitly set to undefined to force `FmtString` to override any explicit `parse_mode` (incorrectly) set by the user. `fmt` is mutually exclusive with `parse_mode`.

@MKRhere, telegram API allows parse_mode to be set. This works:

```typescript
        const res: FmtString = {
         text: txt,
          // @ts-ignore
          parse_mode: 'HTML'
        }
        ctx.editMessageText(res);
```

MKRhere:
This is not the usecase for `FmtString`. It's to be returned from `fmt`. [See docs](https://github.com/feathers-studio/telegraf-docs/blob/595d38649347af3f9dd08252113e3cdb5fa5a79d/guide/formatting.md#fmt).

```TS
ctx.editMessageText(fmt`${This is bold("Bold")}`);

// fmt returns FmtString, forcing parse_mode to be undefined, and ignoring the explicitly set value
ctx.editMessageText(fmt`${This is bold("Bold")}`, { parse_mode: "HTML" });
```

For what you're doing, you just need:

```TS
// not fmt, just parse_mode
ctx.editMessageText(text, { parse_mode: "HTML" });
```

Issue #1749, opened 2022-11-24
Awaiting for telegraf.launch() stops the code.


GordiNoki:
## Context

Trying to await for the bot to launch stops the code. I think this is done so that the code does not close after the bot is launched, but my code requires waiting for the bot to connect for the rest code to work. Is there any better way of doing so?
It worked before with older version of telegraf (4.10.0).

* Telegraf.js Version: 4.11.2
* Node.js Version: v18.0.0
* Operating System: Windows 11

## Minimal Example Code Reproducing the Issue

```ts
const config = require("./config.json");
const { Telegraf } = require("telegraf");

(async () => {
    const telegraf = new Telegraf(config.telegram.token);
    await telegraf.launch();
    console.log("Telegram started.");
})();
```

## Expected Behavior
"Telegram started." in console.

## Current Behavior
Nothing

## Error Message and Logs (`export DEBUG='telegraf:*'`)

```
  telegraf:main Created a `Telegraf` instance +0ms
  telegraf:main Connecting to Telegram +2ms
  telegraf:client HTTP call getMe {} +0ms
  telegraf:main Launching @<botname>_bot +443ms
  telegraf:client HTTP call deleteWebhook { drop_pending_updates: undefined } +442ms
  telegraf:main Bot started with long polling +85ms // Holds there cause of awaiting at telegraf.ts:267 line
  telegraf:polling Starting long polling +0ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 0, allowed_updates: [] } +86ms
```

GordiNoki:
```
> npm install telegraf@4.10.0

added 1 package, changed 1 package, and audited 109 packages in 2s

11 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
> node .\a.js         
  telegraf:main Created a `Telegraf` instance +0ms
  telegraf:main Connecting to Telegram +2ms
  telegraf:client HTTP call getMe {} +0ms
  telegraf:main Launching @<botname>_bot +460ms
  telegraf:client HTTP call deleteWebhook { drop_pending_updates: undefined } +460ms
  telegraf:polling Starting long polling +0ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 0, allowed_updates: [] } +89ms
  telegraf:main Bot started with long polling +91ms
Telegram started.
> 
```

MKRhere:
Thanks for opening this issue.

[...]

Issue #1657, opened 2022-07-17
Add ability to recover from errors


nrzull:
**Is your feature request related to a problem? Please describe.**
For now, if your bot receives, for example, TelegramError, it will crash your entire app and if you have multiple bots then you will not even know which bot crashed (with `process.on("unhandledRejection", (e) => {})`). Related code snippet:

```javascript
        // src/core/network/polling.ts
        if (
          err instanceof TelegramError &&
          // Unauthorized      Conflict
          (err.code === 401 || err.code === 409)
        ) {
          this.skipOffsetSync = true
          throw err
        } else {
          throw err
        }
```

**Describe the solution you'd like**
Would be cool to add ability to handle such errors. It also may have restart method

```javascript
const bot = new Telegraf(token);

bot.on("unhandledRejectionOrAnyOtherReasonThatKilledOurBot", (error) => {
 // here bot is dead, so we could restart him
 bot.restart()
} )
```

**Describe alternatives you've considered**
-

**Additional context**

hypernova7:
You can just do this:

```javascript
const launchOptions = {
  webhook: {
    // ...
  }
}
bot.launch(launchOptions)

process.on('unhandledRejectionOrAnyOtherReasonThatKilledOurBot', (error) => {
  bot.launch(launchOptions)
})
```
You can also make use of `bot.catch()` to prevent your bot from stopping due to an error, just try to use `async/await`

```javascript
bot.catch(error => console.error(error))
bot.help(async ctx => {
  await ctx.reply('Hi')
})
```

nrzull:
@hypernova7 Well, `catch` method will not catch `TelegramError` that crashes your app. At least it didn't help me  :shrug:

Your first snippet is a "fine" workaround, but if I have pool of bots then I will not know which one was stopped. I could register one `unhandledRejection` per one bot creation, but for me such solution looks "smelly".

Anyway, thanks for reply!

MKRhere:
`bot.catch()` catches `TelegramError`, but you must always `await` all your promises for it to work. For example, `await ctx.reply(...)` or `await next()`. If you don't add await/return for all async calls, Telegraf is unable to handle the error, and only Node can. That's why it's an `unhandledRejection`. Telegraf cannot catch `unhandledRejection` per bot, since it is per-process.

AshenCoon:
`bot.catch()`  will not catch this error, it will always be unhandled. It is thrown here
`src/core/network/polling.ts#L77`
```node
[...]

PR #1915, opened 2023-12-21
fix: telegraf/telegraf issue#1749

erfanGharib:
fix issue #1749 by not awating for ```startPolling``` method on launch.
now launch method can resolve on start if successful.

### DIFF
--- a/src/telegraf.ts
+++ b/src/telegraf.ts
@@ -275,7 +275,7 @@ export class Telegraf<C extends Context = Context> extends Composer<C> {
         drop_pending_updates: config.dropPendingUpdates,
       })
       debug('Bot started with long polling')
-      await this.startPolling(config.allowedUpdates)
+      this.startPolling(config.allowedUpdates)
       return
     }
 


MKRhere:
This is how it used to be. The problem with this is that polling errors cannot be caught at runtime (related: #1657) and will always crash the process (except a global exception handler, which is a terrible idea). Being able to catch errors is, well, useful when you have other things running in your application and do not want one bot going down take the whole process with it.

The leading idea we have to counter both issues is to return a LaunchHandle on bot.launch. In that case, bot.launch will resolve immediately after `getMe`, but you have an additional handle you can use to catch polling errors.

```TS
bot.launch().then(handle => {
  console.log("Bot launched successfully");
  handle.onError(err => console.error("A polling error occurred:", err));
});
```

This should be okay to implement, but it'd break the launch function again in v4. So I'm delaying it for v5.

erfanGharib:
That's a good point. By acknowledging this, we can address the issue more effectively.
actually I was not familiar enough with the workflow🙏

MKRhere:
>  I was not familiar enough

That is fine; I appreciate the PR. If you want to try implementing `LaunchHandle` as in my previous comment, you could target the https://github.com/telegraf/telegraf/tree/feat-deno branch in a new PR.

erfanGharib:
Ok! i try to work on it on my free time👍👍

MKRhere:
Thanks! I'm closing this in the meanwhile.

PR #1930, opened 2024-01-14
Add default chatId in defaultGetSessionKey

amorev:
I didn't understand why `chatId` must not be same as `fromId` if `chatId` is not provided. This fix is needed for `pre_checkout_query`, cause there is no `chatId` provided.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -164,7 +164,7 @@ export function session<
 
 function defaultGetSessionKey(ctx: Context): string | undefined {
   const fromId = ctx.from?.id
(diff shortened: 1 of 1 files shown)


MKRhere:
By coincidence, chat_id and from_id happen to be the same value, but nothing in the spec guarantees they will remain the same. This was confirmed by Bot API dev. We should not assume this to be the case. If you want this behaviour, I suggest passing a custom `getSessionKey` parameter to the session function.

MKRhere:
Closing in view of above comment

PR #1935, opened 2024-01-24
Fix getFileLink in test environment

alishermuhamed:
When working in test environment, file links must have `/test` before file path


MKRhere (approved):
lgtm

PR #1949, opened 2024-02-19
API 7.0 & 7.1

### LINKED ISSUE #1749: Awaiting for telegraf.launch() stops the code.

### LINKED ISSUE #1923: Bot API 7.0 Support
**Additional context**

Telegram has released [Bot API 7.0](https://core.telegram.org/bots/api#december-29-2023) which has some cool features like _Reactions_. Would be nice if the _telegraf_ would add a new API features.

### LINKED ISSUE #1938: Property 'match' does not exist for commands
* Telegraf.js Version: 4.15.3
* Node.js Version: 20.10.0

```ts
bot.command(/cmd_(\\d+)/), (async (ctx) => {
    const data = ctx.match[1] // <= Property 'match' does not exist
    ....
})
```

## Expected Behavior

Property 'match' exist.

It is works for hears and action. Support for regexp for command declared, but not fully implemented.

MKRhere:
This is a massive API update and comes with loads of goodies for Telegraf!

See full [release notes here](https://github.com/telegraf/telegraf/blob/feat/api-7.0/release-notes/4.16.0.md).

Resolves #1749 (bot.launch callback)
Resolves #1923 (API 7.0)
Resolves #1938 (ctx.match for Composer.command)
Resolves #1947 (thumbnails in media groups)


hypernova7 (approved):

hypernova7 (approved):

MKRhere:
Need an opinion. For the following update types, a `user` parameter exists instead of `from`. However, it does refer to the user who triggered the update. Perhaps `ctx.from` should also pull from these properties?

- `message_reaction.user`
- `poll_answer.user`
- `chat_boost.boost.source.user`

Edit: This was added in https://github.com/telegraf/telegraf/pull/1949/commits/5bce9a778b82cb1841c1737a368288f88bbfc7e6

hypernova7 (approved):

hypernova7 (approved):

hypernova7 (approved):

hypernova7 (commented):
:+1:

src/context.ts
```diff
@@ -1411,17 +1526,135 @@ type Getter<U extends Deunionize<tg.Update>, P extends string> = PropOr<
...
+        // these updates have `user`
+        | tg.Update.MessageReactionUpdate
+        | tg.Update.PollAnswerUpdate
+        | tg.Update.ChatBoostUpdate
+    ? tg.User
+    : undefined
+
+function getUserFromAnySource<U extends tg.Update>(ctx: Context<U>) {
+  const msg = ctx.msg
+  if (msg?.has('from')) return msg.from
```
ishurmin, line 1597: @MKRhere hi, there might be a problem here with some inconsistent behaviour in case of `callbackQuery` update:
it has `message` field but also has `from` field: https://core.telegram.org/bots/api#callbackquery
So before this change both `ctx.from` and `ctx.callbackQuery.from` were returning the same result, but currently they are different in case of `callbackQuery` update:
`ctx.from` returns `message` and `ctx.callbackQuery.from` returns `from`.

MKRhere: Thanks! This issue was reported in chat, will roll out fix soon!

MKRhere: Fix published as 4.16.1!

PR #1958, opened 2024-02-26
fix: `Context::text` is undefined when message has only text but no entities

AnotiaWang:
The tool function `getTextAndEntitiesFromAnySource()` wrongly checks the `entities` property instead of `text`, so when a message is plain text, `ctx.text` returns `undefined` which is not expected. (Same when there is no caption entities in a photo message)

Example: With the following code:

```ts
console.log('text' in ctx.msg, 'entities' in ctx.msg, ctx.msg.text, ctx.msg.entities, ctx.text)
```

1. Send '123' to the bot

```bash
true false 123 undefined undefined
                         ^^^^^^^^^ (Should be 123)
```

2. Send '**123**' (bold text) to the bot

```bash
true true 123 [
  {
    offset: 0,
    length: 3,
    type: "bold",
  }
] 123
```

Similar issue for `caption` and `caption_entities`.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -112,8 +112,8 @@ export class Composer<C extends Context> implements MiddlewareObj<C> {
         Filter extends tt.MessageSubType
           ? MatchedContext<C, Filter>
           : Filter extends tt.UpdateType | Guard<C['update']>
-            ? FilteredContext<C, Filter>
-            : never
+          ? FilteredContext<C, Filter>
+          : never
       >
     >
   ): this {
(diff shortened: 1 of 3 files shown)


MKRhere:
Thanks! `Context::entities` was added first, and `Context::text` was added after so it was an oversight.

Could you run the linter with `npm run lint -- --fix`?

AnotiaWang:
I ran `npm run lint -- --fix` and pushed, but the CI checks still failed

MKRhere:
Not sure why it formats differently for you; I took the liberty of running over it once. Thanks, though!

MKRhere (approved):

PR #1962, opened 2024-03-02
Fix prettier ci

shahradelahi:
Hello @MKRhere,
On [this commit](https://github.com/telegraf/telegraf/commit/c591338a5b7396c57435d4b29383feea9be71dd5), it seems like the conditional types are supposed to look like this. It's more readable.


shahradelahi:
@MKRhere a review please!

MKRhere (commented):
I find this an unnecessarily overreaching change. We're already running eslint fix with the prettier plugin which keeps things consistent. While this is marginally better, it affects git blame on far too many files for not that much benefit.

PR #1964, opened 2024-03-03
(dx): Better lint

shahradelahi:
@MKRhere


.github/workflows/node.js.yml
```diff
@@ -5,9 +5,11 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ v4 ]
```
MKRhere, line 8: Why remove `branches`? Maybe just add v5 instead?

shahradelahi: By removing `branches`, it's going to run on any push event on any branch. Why just `v5`? Are you going to drop support for `v4`?

MKRhere: No, I mean add v5 to the existing array and make it `[v4, v5]`. It also runs on pull requests, so that's fine.

MKRhere (approved):
lgtm

the1Riddle (commented):
👍

Issue #2032, opened 2024-09-12
Add optional global event listeners in scenes


ialexanderlvov:
**Is your feature request related to a problem? Please describe.**
In most projects made on telegraf it is more convenient when callback_query and some other updateTypes work off-stage, so that action and other listeners are available when in any scene

**Describe the solution you'd like**
I suggest that this behavior be selected by the developer himself, so that there is flexibility in configuring this behavior in BaseScene and Stage themselves
With 
globalListeners?: Partial<Record<MessageTypes, boolean>>

BaseScene

```ts
export class BaseScene<C extends Context = Context> extends Composer<C> {
  id: string
  ttl?: number
  enterHandler: MiddlewareFn<C>
  leaveHandler: MiddlewareFn<C>
  globalListeners?: Partial<Record<MessageTypes, boolean>>

  constructor(id: string, options?: SceneOptions<C>) {
    const opts: SceneOptions<C> = {
      handlers: [],
      enterHandlers: [],
      leaveHandlers: [],
      ...options,
      globalListeners: {
        callback_query: true,
        ...options?.globalListeners,
      },
    }
    super(...opts.handlers)
    this.globalListeners = opts.globalListeners
    this.id = id
    this.ttl = opts.ttl
    this.enterHandler = compose(opts.enterHandlers)
    this.leaveHandler = compose(opts.leaveHandlers)
  }

  enter(...fns: Array<Middleware<C>>) {
    this.enterHandler = compose([this.enterHandler, ...fns])
    return this
  }

  leave(...fns: Array<Middleware<C>>) {
    this.leaveHandler = compose([this.leaveHandler, ...fns])
    return this
  }

  enterMiddleware() {
    return this.enterHandler
  }

  leaveMiddleware() {
    return this.leaveHandler
  }
}
```

Stage

```ts
export class Stage<
  C extends SessionContext<SceneSession<D>> & {
    scene: SceneContextScene<C, D>
  },
  D extends SceneSessionData = SceneSessionData,
> extends Composer<C> {
  options: Partial<SceneContextSceneOptions<D>>
  scenes: Map<string, BaseScene<C>>
  private readonly globalListeners: Partial<Record<MessageTypes, boolean>>

  constructor(
    scenes: ReadonlyArray<BaseScene<C>> = [],
    options?: Partial<SceneContextSceneOptions<D>>,
    globalListeners?: Partial<Record<MessageTypes, boolean>>
  ) {
    super()
[...]

PR #2034, opened 2024-09-13
Feat/global scene listeners

ialexanderlvov:
issue: https://github.com/telegraf/telegraf/issues/2032

Added ability to add global event handlers to Stage and BaseScene

### DIFF
--- a/src/scenes/base.ts
+++ b/src/scenes/base.ts
@@ -1,6 +1,7 @@
 import { Middleware, MiddlewareFn } from '../middleware'
 import Composer from '../composer'
 import Context from '../context'
+import { MessageTypes } from '../telegram-types'
 
 const { compose } = Composer
 
@@ -16,14 +17,27 @@ export class BaseScene<C extends Context = Context> extends Composer<C> {
   ttl?: number
   enterHandler: MiddlewareFn<C>
   leaveHandler: MiddlewareFn<C>
-  constructor(id: string, options?: SceneOptions<C>) {
-    const opts: SceneOptions<C> = {
+  globalListeners?: Partial<Record<MessageTypes, boolean>>
+
+  constructor(
+    id: string,
+    options?: SceneOptions<C>,
+    globalListeners?: Partial<Record<MessageTypes, boolean>>
+  ) {
+    const opts: SceneOptions<C> & {
+      globalListeners?: Partial<Record<MessageTypes, boolean>>
+    } = {
       handlers: [],
       enterHandlers: [],
       leaveHandlers: [],
       ...options,
+      globalListeners: {
+        callback_query: true,
+        ...globalListeners,
+      },
     }
     super(...opts.handlers)
+    this.globalListeners = opts.globalListeners
     this.id = id
     this.ttl = opts.ttl
     this.enterHandler = compose(opts.enterHandlers)
--- a/src/scenes/stage.ts
+++ b/src/scenes/stage.ts
@@ -20,7 +20,7 @@ export class Stage<
   private readonly globalListeners: Partial<Record<MessageTypes, boolean>>
 
   constructor(
-    scenes: ReadonlyArray<BaseScene<C>>,
+    scenes: ReadonlyArray<BaseScene<C>> = [],
     options?: Partial<SceneContextSceneOptions<D>>,
     globalListeners?: Partial<Record<MessageTypes, boolean>>
   ) {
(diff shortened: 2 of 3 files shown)


MKRhere:
You could already do this with normal middleware flow:

```TS
bot.use(session());

const stage = new Stage([ ... ]);

// runs BEFORE scenes, you can call next() to pass through into scenes
stage.command("exit", ctx => ctx.scene.leave());

// match against scenes
bot.use(stage);

// runs if no scene handlers matched or if any matched scenes called next()
bot.command("unmatched", ...);
```

What specifically did you want to achieve by adding this feature?

ialexanderlvov:
inline keyboard is very often present in bots
if you use several scenes, then when you use a scene with inline keyboard and then switch to another scene, inline keyboard stops working until the user is in the scene's context again.

here is a simple example:
```ts
import { session, Telegraf } from 'telegraf';
import { BaseScene, Stage } from 'telegraf/scenes';

const scene1 = new BaseScene<any>('1');
const scene2 = new BaseScene<any>('2');

const bot = new Telegraf<any>('TOKEN');
const stage = new Stage([scene1, scene2]);

scene1.enter((ctx) => {
	ctx.reply('First Menu', {
		reply_markup: {
			inline_keyboard: [
				[
					{ callback_data: 'msg1', text: 'msg 1' },
					{ callback_data: 'msg2', text: 'msg 2' },
				],
				[{ callback_data: 'scene2', text: 'Next Scene' }],
			],
		},
	});
});

// scene 1
scene1.action('msg1', (ctx) => {
	ctx.reply('message 1');
});
scene1.action('msg2', (ctx) => {
	ctx.reply('message 2');
});

scene1.action('scene2', (ctx) => {
	ctx.scene.enter('2');
});

// scene 2
scene2.enter((ctx) => {
	ctx.reply('Second menu', {
		reply_markup: {
			inline_keyboard: [[{ callback_data: 'scene1', text: 'Back Scene' }]],
		},
	});
});

scene2.action('scene1', (ctx) => {
	ctx.scene.enter('1');
});

// middlewares
bot.use(session());
bot.use(stage.middleware());

bot.command('test', (ctx) => {
	return ctx.scene.enter('1');
});
bot.launch();
```

Yes, it is possible to create global listeners via bot.action, but when the bot has many “endpoints” it becomes very inconvenient, and it is more convenient to set these listeners in the context of the scene to which this listener belongs.

demonstration:

desired behavior:

https://github.com/user-attachments/assets/bbc2e717-f99e-4a9e-8057-0d19d5ff0564

scenes:

https://github.com/user-attachments/assets/e0a5dc2d-2123-4c00-9f16-fe368e2ca53f

MKRhere:
If the callback_actions must act globally, I don't see why you would use scenes in this case. Why not also bring the menu out of the scene? Maybe you're looking for spatial encapsulation (colocating logic) whereas scenes provide temporal encapsulation (while within the scene, the scene takes over). You should probably just use Composers.

```TS
const composer = new Composer();

composer.command("test", ctx => {
	return ctx.reply('First Menu', {
		reply_markup: {
			inline_keyboard: [
				[
					{ callback_data: 'msg1', text: 'msg 1' },
					{ callback_data: 'msg2', text: 'msg 2' },
				],
				[{ callback_data: 'scene2', text: 'Next Scene' }],
			],
		},
	});
});

composer.action("msg1", ctx => {
	// .. do things
});
```

Alright, as I wrote the example I think I've understood what you're really trying to achieve. You're using scenes as an addressable container. You want to reuse the menu by name.

Some time back, I started working on a `@telegraf/menu` API [^1]. It never received an implementation, but you're welcome to comment on it and we can refine that rather than adding arguably bad patterns into scenes. I think the menu API addresses our problems better.

[^1]: `@telegraf/menu` https://gist.github.com/MKRhere/bad4b243c6945df82732671613e31b97

ialexanderlvov:
Yeah, I think I'd like to try and contribute to the development of this library
Am I right to understand that @telegraf/menu is focused specifically on inline keyboard? In general, I'm happy with the functionality of the scenes in everything except working with inline keyboard, because of the above described behavior

MKRhere:
That's correct, it's specifically designed to work with inline keyboards.

MKRhere:
I'm closing this issue because the original want is already addressed; global listeners can be attached to `stage` as in my first comment (https://github.com/telegraf/telegraf/pull/2034#issuecomment-2581707901).

As of now, you could create a builder for a menu instead of using scenes, such as `ctx.reply("...", buildMenu(...))` to be able to send a menu from multiple places, but handle them from global bot handlers. I'll also revisit the `@telegraf/menu` proposal some time soon.

PR #2086, opened 2026-02-11
fix: return appropriate HTTP status codes in webhook responses

### LINKED ISSUE #2065: Invalid error status codes for method/path mismatch (always `403`)
Hi, I think is not good to always return `403` when the request method or path does not match the expected. Instead:
- when method does not match return `405`
- when path does not match return `404`
- when secret does not match return `403`

Always returning `403` confuses the client that there is something wrong with the credentials, auth or etc. When there is just an issue with method/path

https://github.com/telegraf/telegraf/blob/0638cf4cc7ba8467ccb9222726024c99c54d119f/src/telegraf.ts#L121-L131

https://github.com/telegraf/telegraf/blob/0638cf4cc7ba8467ccb9222726024c99c54d119f/src/core/network/webhook.ts#L21-L24

https://github.com/telegraf/telegraf/blob/0638cf4cc7ba8467ccb9222726024c99c54d119f/src/core/network/webhook.ts#L13-L17

kaigritun:
## Summary

This PR fixes #2065 by returning appropriate HTTP status codes when webhook requests are rejected:

- **405 Method Not Allowed** for non-POST requests
- **404 Not Found** for webhook path mismatches  
- **403 Forbidden** for secret token mismatches (unchanged behavior)

Previously, all rejection scenarios returned 403 Forbidden, which confused clients into thinking there was an authentication issue when the problem was actually a method or path mismatch.

## Changes

1. **`src/core/network/webhook.ts`**: Added optional `WebhookOptions` parameter with `path` property. The default `next` callback now determines the appropriate status code based on:
   - Whether the request method is POST (405 if not)
   - Whether the path matches (404 if it doesn't)
   - Otherwise assumes authentication failure (403)

2. **`src/telegraf.ts`**: Updated `webhookCallback()` to pass the expected path to `generateCallback`.

3. **`test/telegraf.js`**: Added three test cases to verify the new status code behavior.

## Backward Compatibility

This change is backward compatible:
- The `WebhookOptions` parameter is optional
- Custom `webhookFilter` implementations still work (they return boolean as before)
- The filter logic itself is unchanged - only the rejection status codes are now more specific

## Testing

All existing tests pass, plus 3 new tests for the status code behavior:
- `webhookCallback should return 405 for non-POST requests`
- `webhookCallback should return 404 for path mismatch`  
- `webhookCallback should return 403 for secret token mismatch`

### DIFF
--- a/src/core/network/webhook.ts
+++ b/src/core/network/webhook.ts
@@ -3,15 +3,35 @@ import d from 'debug'
 import { type Update } from '../types/typegram'
 const debug = d('telegraf:webhook')
 
+export interface WebhookOptions {
+  /**
+   * Expected webhook path for proper status code responses.
+   * When provided, enables returning 404 for path mismatches
+   * and 405 for method mismatches instead of generic 403.
+   */
+  path?: string
+}
+
 export default function generateWebhook(
   filter: (req: http.IncomingMessage) => boolean,
-  updateHandler: (update: Update, res: http.ServerResponse) => Promise<void>
+  updateHandler: (update: Update, res: http.ServerResponse) => Promise<void>,
+  options?: WebhookOptions
 ) {
   return async (
     req: http.IncomingMessage & { body?: Update },
     res: http.ServerResponse,
     next = (): void => {
-      res.statusCode = 403
+      // Use appropriate HTTP status codes based on rejection reason
+      if (req.method !== 'POST') {
+        // 405 Method Not Allowed for non-POST requests
+        res.statusCode = 405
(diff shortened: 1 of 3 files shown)


siakinnik:

Leask:
Included in #2092 via commit f703fc51 with webhook status-code regression tests.

siakinnik:
Impemented in [#12](https://github.com/telegraf-hardened/telegraf-hardened/pull/12) in @telegraf-hardened

PR #2087, opened 2026-04-10
fix: add type predicate return types to Msg methods

augchan42:
Fixes TypeScript error where isAccessible and has methods in Msg interface were missing type predicate return types, causing build failure.


siakinnik:

Leask:
Included in #2092 via commit e2f44793.

PR #2088, opened 2026-04-10
fix: create new Error instead of mutating in redactToken

augchan42:
Fix issue where redactToken was mutating the original error object and rethrowing it. Now creates a new Error to avoid potential side effects.

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -292,11 +292,13 @@ async function answerToWebhook(
 }
 
 function redactToken(error: Error): never {
-  error.message = error.message.replace(
+  const message = error.message.replace(
     /\/(bot|user)(\d+):[^/]+\//,
     '/$1$2:[REDACTED]/'
   )
-  throw error
+  const newError = new Error(message)
+  newError.stack = error.stack
+  throw newError
 }
 
 type Response = http.ServerResponse


MKRhere:
By creating a new error, you lose the original stack trace.

According to the spec [1], Error#message should be a writable property. I think an issue must be raised in runtimes that break from the spec. Meanwhile, it might be okay to have a way to opt-out of redaction with an option like dangerouslyLogToken or similar. Thoughts?

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message

augchan42:
Thanks for your comment.  I just noticed my telegram bot was crashing when encountering this so this was done to avoid that crash.  

I've updated the fix to preserve it:
```
function redactToken(error: Error): never {
  const message = error.message.replace(
    /\/(bot|user)(\d+):[^/]+\//,
    '/$1$2:[REDACTED]/'
  )
  const newError = new Error(message)
  newError.stack = error.stack
  throw newError
}
```

I think this would be easier for users.

MKRhere:
And you've tested that overwriting Error#stack instead of Error#message does not error for you?

augchan42:
Not yet.   Give me some time.  Sent from my iPhoneOn 11 Apr 2026, at 4:14 PM, Muthu Kumar ***@***.***> wrote:﻿MKRhere left a comment (telegraf/telegraf#2088)
And you've tested that overwriting Error#stack instead of Error#message does not error for you?

siakinnik:

Leask:
Included in #2092 via commit 6b7664fd. The branch also preserves fetch error names in commit 327d7d37 so abort-related errors remain identifiable.

PR #1992, opened 2024-05-19
fix: `attachFormValue` did not consider value as JSON object such as `reply_parameters`

clansty:
fix #1970

PR #2095, opened 2026-07-07
fix: serialize object-valued multipart fields such as reply_parameters

### LINKED ISSUE #1970: `reply_parameters` doesn't work in `Context.replyWithPhoto`
## Context

In a typescript + webpack project, using `webpack --watch` and `nodemon` to develop.

* Telegraf.js Version: 4.16.3
* Node.js Version: 20.11.1
* Operating System: Kubuntu 23.10 x86_64

```ts
export async function replyAPhoto(ctx: Context) {
    const command_message_id = ctx.message.message_id;
    const photoBuffer = await getPhotoBuffer();
    await ctx.replyWithPhoto({source: photoBuffer}, {
        reply_parameters: { message_id: command_message_id },
    });
}
```

I have registered that function in `/see` command. When I send the command, the photo is sent, but not reply.

## Expected Behavior

Send the photo and reply my command message

## Current Behavior

Send the photo but not reply my command message

## Error Message and Logs (`export DEBUG='telegraf:*'`)

```
  telegraf:main Created a `Telegraf` instance +0ms
Bot is running
  telegraf:main Connecting to Telegram +1ms
  telegraf:main Launching @mybot +1s
  telegraf:client HTTP call deleteWebhook { drop_pending_updates: undefined } +1s
  telegraf:main Bot started with long polling +298ms
  telegraf:polling Starting long polling +0ms
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 0, allowed_updates: [] } +299ms
  telegraf:main Processing update 306145224 +4s
  telegraf:client HTTP call sendPhoto {
  chat_id: 6152025353,
  photo: {
    source: <Buffer 89 50 4e 47 0d 0a 1a 0a 00 00 00 0d 49 48 44 52 00 00 02 80 00 00 04 70 08 06 00 00 00 7a 8f e5 00 00 03 30 8f 49 44 41 54 78 01 ec c1 07 80 5e f7 59 ... 209046 more bytes>
  },
  message_thread_id: undefined,
  reply_parameters: { message_id: 88 }
} +4s
  telegraf:main Finished processing update 306145224 +2s
  telegraf:client HTTP call getUpdates { timeout: 50, offset: 306145225, allowed_updates: [] } +2s

```

### LINKED ISSUE #2094: reply_parameters silently dropped from multipart requests (sendVideo/sendPhoto with uploaded media) in 4.16.3 — fix exists but was never released
## Context

This is essentially the same root cause as #1970, but I'm filing it with a concrete non-`Context` reproduction (`telegram.sendVideo` with an uploaded file) and, more importantly, an ask: the fix has existed for a long time (#1992, also included in #2092 via commit 877d9515) but **no released version on npm contains it** — `latest` is still `4.16.3`.

## Telegraf.js Version

4.16.3 (npm `latest`)

## User OS

Linux (Node.js 20)

```ts
await bot.telegram.sendVideo(
  chatId,
  { url: 'http://localhost:3000/uploads/some-video.mp4' }, // any InputFile that triggers an upload
  {
    caption: 'edited message notification',
    parse_mode: 'HTML',
    reply_parameters: { message_id: 2919 },
  }
);
```

## Expected Behavior

The sent video is a reply to message `2919`.

## Current Behavior

The video is sent successfully, but **not** as a reply. No error, no warning — `reply_parameters` is silently discarded.

The same options object works fine with `sendMessage` (JSON request body) or when the media is passed as a plain URL string (also JSON body). The field is only lost when Telegraf switches to `multipart/form-data` because the media needs to be uploaded (`{ url }` / `{ source }` input).

## Root Cause

In `src/core/network/client.ts`, `buildFormDataConfig` only JSON-serializes the fields listed in `FORM_DATA_JSON_FIELDS` (`results`, `reply_markup`, `mask_position`, `shipping_options`, `errors`). `reply_parameters` is not in that list, so it reaches `attachFormValue` as a plain object; since it has no `media`/`type`/`url`/`source` props, it falls through to `attachFormMedia`, which finds nothing to attach and drops the field entirely.

This affects any object-valued field that isn't in `FORM_DATA_JSON_FIELDS`, e.g. `link_preview_options` as well.

## Ask

Could #1992 / #2092 (commit 877d9515) be merged and published to npm? This bug silently breaks reply threading for every bot that uploads media, and the fix has been sitting unreleased for over a year.

## Workaround for anyone hitting this

Pre-serialize the field yourself — Telegraf passes string fields through to the form untouched and the Bot API accepts JSON-encoded object fields in multipart requests:

```ts
reply_parameters: JSON.stringify({ message_id: 2919 }) as never,
```

Related: #1970, #1992, #2092

mostafa8026:
## Summary

Fixes a long-standing bug where `reply_parameters` (and other plain object fields) are **silently dropped** when Telegraf sends media via multipart/form-data.

This is the same root cause reported in #1970 and #2094. In `attachFormValue`, any object that was not a media-group item or `InputMedia` wrapper fell through to `attachFormMedia`, which only handles `{ url }` / `{ source }` attachments. Plain objects like `{ message_id: 2919 }` were discarded with no error.

**Reproduction (4.16.3):**

```ts
await bot.telegram.sendVideo(
  chatId,
  { url: 'https://example.com/video.mp4' },
  { reply_parameters: { message_id: 2919 } }
);
// video is sent, but NOT as a reply
```

`sendMessage` with the same `reply_parameters` works because it uses a JSON body.

## Fix

After handling known media shapes, only treat values with `url`/`source` as `InputFile` attachments. All other objects are JSON-serialized into the multipart form, which the Bot API accepts.

This is the same approach as #1992 (approved LGTM in 2024) and the relevant part of #2092, extracted as a minimal, reviewable change against `v4`.

## Test plan

- [x] `npm run build`
- [x] `npm test` (162 tests passed)
- [ ] Manual: `sendVideo` / `sendPhoto` with `{ url }` input and `reply_parameters` now replies correctly

Fixes #1970
Closes #2094

Related: #1992, #2092

### DIFF
--- a/src/core/network/client.ts
+++ b/src/core/network/client.ts
@@ -220,7 +220,18 @@ async function attachFormValue(
       }),
     })
   }
-  return await attachFormMedia(form, value as InputFile, id, agent)
+  if (
+    value &&
+    typeof value === 'object' &&
+    ((hasProp(value, 'source') && typeof value.source !== 'undefined') ||
+      (hasProp(value, 'url') && typeof value.url !== 'undefined'))
+  ) {
+    return await attachFormMedia(form, value as InputFile, id, agent)
+  }
+  return form.addPart({
+    headers: { 'content-disposition': `form-data; name="${id}"` },
+    body: JSON.stringify(value),
+  })
 }
 
 async function attachFormMedia(


ramincsy (approved):
Looks good — I walked the diff in attachFormValue against the #2094 repro (sendVideo / multipart + reply_parameters).

The previous fall-through to attachFormMedia for any leftover object is exactly why plain objects like { message_id } were silently dropped. Guarding on source / url and JSON-serializing everything else into the multipart part matches Bot API expectations and aligns with the approach from #1992 / #2092, but as a minimal v4-targeted change. Nice extraction.

One concrete follow-up I would love to see (or can help land): a focused regression test that builds a multipart payload for media + reply_parameters (and ideally link_preview_options) and asserts the form includes a JSON part for those fields rather than treating them as media. That would lock the silent-drop bug from coming back.

@mostafa8026 — if you are open to a small Co-authored-by test commit from me for Pair Extraordinaire collaboration, I am happy to push one onto this branch (or open a tiny follow-up PR against yours). Approve from my side either way.

mostafa8026:
@ramincsy Thanks for reviewing this and offering to help! I agree that a focused regression test is the missing piece here.

A small PR against `mostafa8026/telegraf:fix/multipart-reply-parameters` would be welcome. For the main case, we could use `sendVideo` or `sendPhoto` with a buffer-backed upload and `reply_parameters: { message_id: 2919 }`, capture the outgoing request with a local HTTP server, and assert that:

- the request is multipart and still contains the uploaded media;
- it includes a `reply_parameters` form part whose JSON parses to the original object.

That should fail on the previous implementation and pass with this fix, without needing a bot token or a real Telegram request.

One nuance on `link_preview_options`: it belongs to methods such as [`sendMessage`](https://core.telegram.org/bots/api#sendmessage), so I would cover it separately by asserting that a preview option containing `url` stays in the JSON request and does not trigger an attachment download. `includesMedia` already special-cases that field; the multipart attachment guard still recognizes objects with `url` as files.

Happy to collaborate and preserve credit for your test contribution when we incorporate it. Thanks again!

MKRhere:
To unblock this PR I added some tests to `attachFormValue`. I'm working on major upgrades to Telegraf, but that's going to take a few days so let's land this for now in a minor version.

